PR #56, opened 2017-03-16
fix(image/lazyLoad): duplicate aspect ratios config for sass and js code

kikoruiz:
@SUI-Components/developers please review! 😿


carlosvillu (approved):
Piensa que aun tenemos los breakpoints !!

PR #70, opened 2017-04-24
feat(form/autocompleted): make focus input method public

kikoruiz:

### DIFF
--- a/components/form/autocompleted/src/index.js
+++ b/components/form/autocompleted/src/index.js
@@ -66,7 +66,7 @@ export default class FormAutocompleted extends Component {
     })
   }
 
-  _focusInput = () => {
+  focusInput = () => {
     this.input.focus()
   }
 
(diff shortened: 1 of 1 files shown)


components/form/autocompleted/src/index.js
```diff
@@ -66,7 +66,7 @@ export default class FormAutocompleted extends Component {
     })
   }
 
-  _focusInput = () => {
+  focusInput = () => {
```
carlosvillu, line 69: Why we dont use a focus prop ?!

JoanClaret (approved):

components/form/autocompleted/src/index.js
kikoruiz: We could, but in some cases (such in ours) it's preferable to use a public method as it has less performance implications and re-renderings issues, because we have to manage use cases like this one: https://github.schibsted.io/scmspain/frontend-mt--uilib-components/blob/master/components/search/box/src/index.jsx#L51

PR #72, opened 2017-04-24
feat(card/article): add featured card

JoanClaret:
add featured card

### DIFF
--- a/components/card/article/src/index.js
+++ b/components/card/article/src/index.js
@@ -33,15 +33,19 @@ export default function CardArticle (props) {
     description,
     tag,
     comments,
-    lazyLoad
+    lazyLoad,
+    featured = false
   } = props
   const Link = props.linkFactory
   const tagClassName = cx('sui-CardArticle-tag', {
     [`sui-CardArticle-tag--${tag.type}`]: typeof tag.type !== 'undefined'
   })
+  const cardClassName = cx('sui-CardArticle', {
+    'is-featured': featured
+  })
 
   return (
-    <div className='sui-CardArticle'>
+    <div className={cardClassName}>
       <Link href={link} className='sui-CardArticle-link'>
         {lazyLoad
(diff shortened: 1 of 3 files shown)


zecafa (approved):

components/card/article/src/index.js
```diff
@@ -33,26 +33,32 @@ export default function CardArticle (props) {
     description,
     tag,
     comments,
-    lazyLoad
+    lazyLoad,
+    featured = false
```
kikoruiz: Better use `defaultProps`

components/card/article/src/index.js
```diff
@@ -33,26 +33,32 @@ export default function CardArticle (props) {
...
   } = props
   const Link = props.linkFactory
```
kikoruiz: We can move this assignment into the previous object and then move the whole destructured object to the function definition, so that we get rid off the first 2 consts.

components/card/article/src/index.scss
```diff
@@ -13,11 +13,17 @@ $c-article-tag-types: () !default;
   }
 
   &-info {
-    align-items: center;
-    display: flex;
-    justify-content: space-between;
-    margin: 0 $m-h;
-    padding: $p-v 0;
+    .is-featured & {
```
kikoruiz: Not sure if this is a good solution, because we're defining a global selector, what will happen if there's something like this?:
```html
<div class="app is-featured">
  <!-- Bla, bla, bla... -->
  <div class="sui-CardArticle-infoInner">Text</div>
</div>
```

We should just apply this selector to the specific element.

JoanClaret: agree, thanks for point it out!

components/card/article/src/index.js
JoanClaret: 👍  Fixed!

kikoruiz (approved):
👍

PR #75, opened 2017-04-26
Feature/card article tag

JoanClaret:
This PR includes 2 main features:

- Modify tag-chip so it can receive a `className`
- Include `sui-tag-chip` in card-article instead of using a link.

### DIFF
--- a/components/card/article/src/index.scss
+++ b/components/card/article/src/index.scss
@@ -1,8 +1,8 @@
 /* sass-lint:disable no-important, class-name-format*/
 @import '~@schibstedspain/theme-basic/lib/index';
 @import '~@schibstedspain/sui-image-lazy-load/lib/index';
+@import '~@schibstedspain/sui-tag-chip/lib/index';
 
-$c-article-tag: $c-gray-light !default;
 $c-article-tag-types: () !default;
 
 .sui-CardArticle {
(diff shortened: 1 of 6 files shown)


components/card/article/src/index.scss
```diff
@@ -28,17 +28,12 @@ $c-article-tag-types: () !default;
 
   &-tag {
     @include reset-link;
-    background-color: $c-article-tag;
-    border-radius: $bdrs-small;
-    color: $c-white;
-    font-size: $fz-s;
-    font-weight: $fw-bold;
-    padding: $p-v-xsmall $p-h-xsmall;
-    text-transform: uppercase;
```
kikoruiz: Missing uppercase.

components/tag/chip/src/index.js
```diff
@@ -6,8 +6,9 @@ const Tag = ({Link, children, url, ...rest} = {}) => url
   ? <Link href={url} {...rest}>{children}</Link>
   : <span {...rest}>{children}</span>
 
-const tagChipClassName = ({isClickable}) => cx('sui-TagChip', {
-  'sui-TagChip-link': isClickable
+const tagChipClassName = ({isClickable, className}) => cx('sui-TagChip', {
+  'sui-TagChip-link': isClickable,
+  [className]: className
```
kikoruiz: I think it's not needed. In any case, you can define a default prop as empty string.

JoanClaret: Fixed!

components/card/article/src/index.scss
JoanClaret: yeah! done.

kikoruiz (approved):

PR #76, opened 2017-05-02
feat(card/article): remove tag types

JoanClaret:
- remove tag types in this sui component in order to manage them in the mt-component. 
- include a tag classname prop

### DIFF
--- a/components/card/article/src/index.js
+++ b/components/card/article/src/index.js
@@ -152,7 +152,6 @@ CardArticle.propTypes = {
 
 CardArticle.defaultProps = {
   featured: false,
-  tagClassName: null,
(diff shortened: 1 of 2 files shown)


components/card/article/src/index.js
```diff
@@ -149,6 +152,7 @@ CardArticle.propTypes = {
 
 CardArticle.defaultProps = {
   featured: false,
+  tagClassName: null,
```
kikoruiz: I'd expect some string as default value regarding the prop types, if there's none we can just remove it.

JoanClaret: ok, fixed! @kikoruiz

kikoruiz (approved):

PR #79, opened 2017-05-04
feat(card/article): get tag chip as a prop

JoanClaret:
include tag-chip as a prop, so other components will be able to pass specific tag-chip-component.


kikoruiz (approved):

naxhh (commented):
Shouldn't you add the prop to the `prorpTypes` also?

JoanClaret:
absolutely @naxhh . Fixed!

naxhh (approved):

PR #87, opened 2017-05-23
Bugfix/mouse over before interactive dom

naxhh:
This PR Fixes an issue present in the current drop downs.

If you keep your mouse on the menu item that should be displayed the `expanded` state gets inverted.
This is because we where using `_toggleMenu` who does not care about the previous state.

The change ensures that `onMouseOver` expandend is always `true` and in `onMouseOut` expanded is always `false`


kikoruiz (approved):
👍

kikoruiz (approved):

PR #92, opened 2017-05-25
feat(image/slider): simple gallery using react-slidy

betobe:
Simple gallery using react-slidy.


components/image/slider/src/index.js
```diff
@@ -0,0 +1,60 @@
+import React, {PropTypes} from 'react'
+import ReactSlidy from 'react-slidy'
+import SuiMultimedia from '@schibstedspain/sui-multimedia'
```
kikoruiz: Here we're using a legacy component, I suggest to use this one https://github.com/SUI-Components/sui-components/tree/master/components/image/lazyLoad

components/image/slider/package.json
```diff
@@ -0,0 +1,22 @@
...
+  "version": "1.0.0",
+  "description": "",
+  "main": "lib/index.js",
+  "scripts": {
+    "build": "rm -Rf ./lib && mkdir -p ./lib && npm run babel && npm run sass",
+    "babel": "../../../node_modules/.bin/babel --presets schibsted-spain ./src --out-dir ./lib",
+    "sass": "../../../node_modules/.bin/cpx \"./src/**/*.scss\" ./lib",
+    "preversion": "echo \"preversion\"",
+    "version": "npm run build",
+    "postversion": "git add ${PWD} && git commit -m \"release(image/slider): v$(node -p -e \"require('./package.json').version\")\" && git push origin master && npm publish"
```
JoanClaret: should we flag it as public?
`&& npm publish --access=public`

components/image/slider/src/index.js
betobe: I've checked lazyLoad and it doesn't cover all the required features. Thanks anyway!

components/image/slider/package.json
betobe: Yes, It seems I should do that 👍

components/image/slider/src/index.js
kikoruiz: Then we should discuss it more deeply in order to try to use only SUIStudio-based components.

naxhh (commented):
Care, `.COMPONENTS` is not required anymore. You should update sui studio to 3.X

components/image/slider/src/index.js
```diff
@@ -0,0 +1,64 @@
...
+
+/**
+ * @param {Array} images List given by props.images.
+ * @param {boolean} lazyLoad
+ * @return {Array} List of img elements.
+ */
+const getSlides = (images, lazyLoad) => {
+  if (images && images.length) {
+    return images.map( (image, index) => {
+      return (<img key={index} src={image.src} alt={image.alt} />)
```
kikoruiz: Not needed `a` element for each image?

components/image/slider/.gitignore
```diff
@@ -0,0 +1,3 @@
+
+lib
```
jcger, line 2: Have in mind that if you do this no 'lib' folder in any subfolder will be tracked. Not sure if we want that, as lib is a common subfolder name. '/lib' would be more accurate.

jcger (approved):

components/image/slider/src/index.js
betobe: Not anymore since onClick is managed at a higher level.

components/image/slider/.gitignore
betobe: You are right, but IMO it is not a real problem.

components/image/slider/src/index.js
```diff
@@ -0,0 +1,64 @@
...
+  images: PropTypes.arrayOf(
+    PropTypes.shape({
+      src: PropTypes.string.isRequired,
+      alt: PropTypes.string
+    }).isRequired
+  ),
+  /**
+   * Callback executed when clicking over the slider.
+   */
+  onClickCB: PropTypes.func
```
davidbarna: Why do you use "onClickCB" ?
The world uses "onClick", we have some components like ui-links that use "handleClick", and now yours is another trend :-)
I'd rather have onClick or onSlideClick. Or even handleClick to follow what we already have... But not the CB suffix .

davidbarna (approved):

components/image/slider/src/index.js
```diff
@@ -0,0 +1,64 @@
...
+ * @return {boolean}
+ */
+const hasMoreThanOneImage = (images) => (images && 1 < images.length)
+
+/**
+ * @param {Array} images List given by props.images.
+ * @param {boolean} lazyLoad
+ * @return {Array} List of img elements.
+ */
+const getSlides = (images, lazyLoad) => {
```
davidbarna: What's lazyload for? I don't see the use of it...
Will it be on the next PR ?

betobe: Unused

components/image/slider/src/index.js
```diff
@@ -51,11 +51,11 @@ ImageSlider.propTypes = {
   /**
    * Callback executed when clicking over the slider.
    */
-  onClickCB: PropTypes.func
+  handleClick: PropTypes.func
```
davidbarna: 💋

davidbarna (approved):

components/image/slider/src/index.js
kikoruiz: Ok, then

kikoruiz (approved):

components/image/slider/src/index.js
```diff
@@ -0,0 +1,63 @@
...
+  )
+}
+
+/**
+ * Says if there are more than one image in the given list.
+ * Implemented for semantic purposes.
+ * @param {Array} images List given by props.images.
+ * @return {boolean}
+ */
+const hasMoreThanOneImage = (images) => (images && 1 < images.length)
```
davecarter: Should this method be declared in line 7 instead?

naxhh: if you do that you would re define the method on each render.

This way you defined the method only once. Since there is no requirement to re-define the method on render i would say it's fine this way

components/image/slider/src/index.js
```diff
@@ -0,0 +1,63 @@
...
+ * @param {Array} images List given by props.images.
+ * @return {Array} List of img elements.
+ */
+const getSlides = (images) => {
+  if (images && images.length) {
+    return images.map( (image, index) => {
+      return (<img key={index} src={image.src} alt={image.alt} />)
+    })
+  }
+  else{
```
midudev: Is that passing the linter???

components/image/slider/src/index.js
```diff
@@ -0,0 +1,63 @@
...
+    }).isRequired
+  ),
+  /**
+   * Callback executed when clicking over the slider.
+   */
+  handleClick: PropTypes.func
+}
+
+ImageSlider.defaultProps = {
+  handleClick: (...args) => {}
```
midudev: This `...args` is not needed.

betobe: Yes. I've added it as a clue but clearly I could add a comment 🙈

JoanClaret (approved):

PR #95, opened 2017-06-01
feat(topbar/user): allo setting notifications in the cta

naxhh:


components/topbar/user/src/index.scss
```diff
@@ -118,6 +119,19 @@
             fill: $c-primary !important;
           }
         }
+
+        &Notifications {
+          background-color: $c-primary;
+          border-radius: 50%;
+          color: $c-white;
+          font-size: $fz-xs;
+          height: 17px;
```
kikoruiz: This magic number is asking for a Sass var 😝

components/topbar/user/src/index.scss
```diff
@@ -147,6 +161,21 @@
             $background-color: $bgc-topbar-user-cta,
             $icon-fill: $c-white
           );
+          position: relative;
+        }
+
+        &Notifications {
+          background-color: $c-primary;
```
kikoruiz: I think we can reuse some common code and create a mixin with it, then apply specific styles in each selector. Can help if you need.

components/topbar/user/src/index.scss
```diff
@@ -147,6 +161,21 @@
...
+          border-radius: 50%;
+          color: $c-white;
+          font-size: $fz-xs;
+          height: 17px;
+          line-height: 17px;
+          position: absolute;
+          right: -7;
```
midudev: Missing the `px`here and in the top. :+1:

naxhh: ups!

components/topbar/user/src/index.scss
naxhh: done!

components/topbar/user/src/index.scss
naxhh: and done!

I don't know if what i did is what you wanted, or just keep top & right outside the mixin?

kikoruiz (approved):
👏

JoanClaret (approved):
Like a lot the mixin solution to avoid duplicated code. congratulations!

carlosvillu (approved):

carlosvillu (approved):

PR #97, opened 2017-06-06
feat(form/autocompleted): enables closeOnOutsideClick

JoanClaret:
enables closeOnOutsideClick

### DIFF
--- a/components/form/autocompleted/README.md
+++ b/components/form/autocompleted/README.md
@@ -25,6 +25,7 @@ The component exposes the following props:
 * handleBlur (Function): This function is called everytime user exits the input.
 * handleFocus (Function): This function is called everytime user focus on the input.
 * selectFirstByDefault (Boolean): **Optional** It sets first position for the autocomplete default active option. Defaults to `true`.
+* closeOnOutsideClick (Boolean): **Optional** Enables close suggestions on click outside feature
 * focus (Boolean): **Optional** It trigger focus in the input. Defaults to `false`.
 * submit (object): **Optional** contains an optional icon and an optional text for the submit button
 
@@ -76,6 +77,7 @@ class MyFormAutocompleted extends React.Component {
         suggests={this.state.suggests}
         submit={{ text: 'Submit' }}
         collapsed
+				closeOnOutsideClick
         />
     )
   }
--- a/components/form/autocompleted/README.md
+++ b/components/form/autocompleted/README.md
@@ -25,7 +25,7 @@ The component exposes the following props:
 * handleBlur (Function): This function is called everytime user exits the input.
 * handleFocus (Function): This function is called everytime user focus on the input.
 * selectFirstByDefault (Boolean): **Optional** It sets first position for the autocomplete default active option. Defaults to `true`.
-* closeOnOutsideClick (Boolean): **Optional** Enables close suggestions on click outside feature
+* closeOnOutsideClick (Boolean): **Optional** Enables close suggestions on click outside. Elements in the array `excludeFromOutsideClick` (`input`, `submit`, `suggestList`) will not close the suggest list.
 * focus (Boolean): **Optional** It trigger focus in the input. Defaults to `false`.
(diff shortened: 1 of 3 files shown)


components/form/autocompleted/src/index.js
```diff
@@ -143,15 +148,28 @@ export default class FormAutocompleted extends Component {
...
     if (this.state.focus) {
       this.focusInput()
     }
+    this.excludeFromOutsideClick = [
+      this.input,
+      this.submit,
+      this.suggestList
+    ]
+    this.props.closeOnOutsideClick === true &&
+    window.addEventListener('click', this._handleOutsideClick, false)
```
davidbarna, line 175: Listener should be removed on componentDidUnMount

components/form/autocompleted/src/index.js
```diff
@@ -143,15 +148,28 @@ export default class FormAutocompleted extends Component {
...
+    !this.excludeFromOutsideClick.includes(event.target) &&
+    this.setState({
+      showSuggestsList: false
+    })
 
   componentDidMount () {
     if (this.state.focus) {
       this.focusInput()
     }
+    this.excludeFromOutsideClick = [
```
davidbarna, line 169: This solution is ok. But if another subelement were added, another ref should be created and added to excludeFromOutsideClick. Otherwise, we could have a very hardly detectable error. 

What about traversing up the target to see is one of the parents if the main node of the component ?
something like:

```js
var target = event.target
var isInnerClick = false
while(target.parentElement) {
   if( target === componentRef ) {
      isInnerClick = true
      break
   }
}

!isInnerClick && this.setState({ showSuggestsList: false})
```

With it you can remove part of your code and get rid of the risk explained above.
Outerclick could even be a component with this single responsibility.

components/form/autocompleted/src/index.js
```diff
@@ -127,7 +128,11 @@ export default class FormAutocompleted extends Component {
   }
 
   _renderSubmitButton = ({ text, icon: Icon }) => (
-    <button className='sui-FormAutocompleted-submit' onClick={this._handleSubmit}>
+    <button
+      className='sui-FormAutocompleted-submit'
+      onClick={this._handleSubmit}
+      ref={node => { this.submit = node }}
```
naxhh, line 137: could you add `this.submit` and `this.suggestList` to the constructor? (this.submit = null)

JoanClaret: Done, thx!

components/form/autocompleted/src/index.js
JoanClaret: done on `componentWillUnmount`. thanks for point it out!

components/form/autocompleted/src/index.js
davidbarna: @JoanClaret pointed out that the tree can be heavy and provoke too many traversing.
This PR should be merged but could be interesting to know if one of you guys knows if it's really an issue.

JoanClaret: This looks like a fancy solution, but since this event will be triggered anywhere in the page i'm not sure about the performance. What do you think @carlosvillu @kikoruiz ?

components/form/autocompleted/README.md
```diff
@@ -25,6 +25,7 @@ The component exposes the following props:
 * handleBlur (Function): This function is called everytime user exits the input.
 * handleFocus (Function): This function is called everytime user focus on the input.
 * selectFirstByDefault (Boolean): **Optional** It sets first position for the autocomplete default active option. Defaults to `true`.
+* closeOnOutsideClick (Boolean): **Optional** Enables close suggestions on click outside feature
```
kikoruiz: It would be great to document the elements that are not taken into account in order to close the suggestions list.

kikoruiz (approved):
A part from the documentation comment... 👍

davidbarna (approved):

PR #98, opened 2017-06-07
release(ad/smartbanner): removes fixed positioning in favor of static

davecarter:
In order to avoid unwanted interactions with the new mobile actionBar the SmartBanner positioning has to change from absolute to static.

Instead of simply remove one style positioning with the other, I'm adding a prop `static` to define the component layout.

### DIFF
--- a/components/ad/smartbanner/package.json
+++ b/components/ad/smartbanner/package.json
@@ -16,7 +16,6 @@
   "license": "ISC",
   "dependencies": {
     "@schibstedspain/sui-svgiconset": "1",
-    "@schibstedspain/suistudio-fatigue-deps": "github:SUI-Components/suistudio-fatigue-deps",
-    "classnames": "2.2.5"
+    "@schibstedspain/suistudio-fatigue-deps": "github:SUI-Components/suistudio-fatigue-deps"
   }
 }
--- a/components/ad/smartbanner/package.json
+++ b/components/ad/smartbanner/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@schibstedspain/sui-ad-smartbanner",
-  "version": "2.0.0",
+  "version": "1.2.0",
   "description": "",
   "main": "lib/index.js",
   "scripts": {
@@ -16,6 +16,7 @@
   "license": "ISC",
   "dependencies": {
     "@schibstedspain/sui-svgiconset": "1",
-    "@schibstedspain/suistudio-fatigue-deps": "github:SUI-Components/suistudio-fatigue-deps"
+    "@schibstedspain/suistudio-fatigue-deps": "github:SUI-Components/suistudio-fatigue-deps",
+    "classnames": "2.2.5"
   }
 }
--- a/components/ad/smartbanner/package.json
+++ b/components/ad/smartbanner/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@schibstedspain/sui-ad-smartbanner",
-  "version": "1.2.0",
+  "version": "1.1.0",
   "description": "",
   "main": "lib/index.js",
   "scripts": {
--- a/components/ad/smartbanner/package.json
+++ b/components/ad/smartbanner/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@schibstedspain/sui-ad-smartbanner",
-  "version": "1.1.0",
+  "version": "2.0.0",
   "description": "",
   "main": "lib/index.js",
   "scripts": {
--- a/components/ad/smartbanner/src/index.js
+++ b/components/ad/smartbanner/src/index.js
@@ -2,7 +2,7 @@ import React, {Component, PropTypes} from 'react'
 import IconCloseDefault from '@schibstedspain/sui-svgiconset/lib/X'
 import cx from 'classnames'
 
-export default class AdSmartbanner extends Component {
+class AdSmartbanner extends Component {
   IconClose = this.props.icon || IconCloseDefault
 
   _handleClick = (event) => {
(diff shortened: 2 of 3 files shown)


ruben-martin-lozano (approved):

aleanez (approved):

components/ad/smartbanner/src/index.js
```diff
@@ -17,8 +18,12 @@ export default class AdSmartbanner extends Component {
   }
 
   render () {
+    const positioning = cx('sui-AdSmartbanner', {
```
ruben-martin-lozano: Por standard, esta variable la llamaria `className` en lugar de positioning: lo primero que me viene por positioning seria "top, bottom, etc.", pero el valor de esta variable son precisamente className/s

components/ad/smartbanner/package.json
```diff
@@ -16,6 +16,7 @@
   "license": "ISC",
   "dependencies": {
     "@schibstedspain/sui-svgiconset": "1",
-    "@schibstedspain/suistudio-fatigue-deps": "github:SUI-Components/suistudio-fatigue-deps"
+    "@schibstedspain/suistudio-fatigue-deps": "github:SUI-Components/suistudio-fatigue-deps",
+    "classnames": "2.2.5"
```
kikoruiz: `suistudio-fatigue-deps` already set `classnames` as dependency.

components/ad/smartbanner/src/index.js
davecarter: me cuesta buscar un nombre apropiado a las cosas.. piensa que mi perro se llamaba Render

components/ad/smartbanner/src/index.scss
```diff
@@ -58,4 +58,8 @@
   &-hidden {
     margin-top: 0;
   }
+
+  &-static {
```
kikoruiz: It should be a modifier, or even better a simple state: `.is-static`.

davecarter: I used the previously name convention as in: `-hidden`

kikoruiz: I didn't notice it, but it should be changed as well.

components/ad/smartbanner/src/index.js
```diff
@@ -42,7 +47,8 @@ AdSmartbanner.propTypes = {
   imageUrl: PropTypes.string.isRequired,
   title: PropTypes.string.isRequired,
   text: PropTypes.string.isRequired,
-  buttonText: PropTypes.string.isRequired
+  buttonText: PropTypes.string.isRequired,
+  static: PropTypes.bool
```
danivm: You should to define defaultProps for all non-required props

components/ad/smartbanner/src/index.scss
```diff
@@ -51,11 +51,15 @@
...
 
-  &-visible {
+  &--visible {
     margin-top: 70px;
   }
 
-  &-hidden {
+  &--hidden {
     margin-top: 0;
   }
```
ruben-martin-lozano: Where are those "visible" and "hidden" classes used?

davecarter: anywhere 😲  !

kikoruiz: Well, another thing to discuss is... why are they called visible and hidden? But... "that's another war" 😄

kikoruiz (approved):

components/ad/smartbanner/package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "@schibstedspain/sui-ad-smartbanner",
-  "version": "1.1.0",
+  "version": "1.2.0",
```
carlosvillu: Please dont touch the version of the package. Is an automatic process

carlosvillu (changes requested):

components/ad/smartbanner/src/index.scss
davecarter: Now I know 😅 
Turns out this modifier is appended to the App Container DIV:
`<div id="App" class="sui-AdSmartbanner-visible">`

components/ad/smartbanner/src/index.js
```diff
@@ -17,8 +18,12 @@ export default class AdSmartbanner extends Component {
   }
 
   render () {
+    const className = cx('sui-AdSmartbanner', {
+      'sui-AdSmartbanner-static': this.props.static
```
JoanClaret: What do you think about defining the static class as a modifier? something like `is-static`.

carlosvillu (approved):

danivm (approved):
🚀

components/ad/smartbanner/src/index.scss
JoanClaret: Do you think this is a component responsability? i mean, looks like this class should be added somewhere outside the component.

ruben-martin-lozano: Tal cual @JoanClaret!

components/ad/smartbanner/src/index.scss
```diff
@@ -51,11 +51,11 @@
     }
   }
 
-  &-visible {
+  &--visible {
     margin-top: 70px;
```
JoanClaret: magic number? `70px`

JoanClaret: maybe you should use `$h-ad-smartbanner`

davecarter: yes! you're right. Done

PR #101, opened 2017-06-26
feat(breadcrumb/basic): restyling

JoanClaret:
feat(breadcrumb/basic): restyling

### DIFF
--- a/components/breadcrumb/basic/src/index.js
+++ b/components/breadcrumb/basic/src/index.js
@@ -4,10 +4,7 @@ import Chevronright from '@schibstedspain/sui-svgiconset/lib/Chevronright'
 import cx from 'classnames'
 
 export default class BreadcrumbBasic extends Component {
-  constructor (...args) {
-    super(...args)
-    this.state = {isExpanded: false}
-  }
+  state = { isExpanded: false }
 
   _expandBreadcrumb = () => {
     this.setState({isExpanded: true})
--- a/components/breadcrumb/basic/src/index.js
+++ b/components/breadcrumb/basic/src/index.js
@@ -1,27 +1,42 @@
 /* eslint-disable react/prop-types */
-import React, {PropTypes} from 'react'
+import React, {Component, PropTypes} from 'react'
 import Chevronright from '@schibstedspain/sui-svgiconset/lib/Chevronright'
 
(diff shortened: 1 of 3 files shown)


naxhh (changes requested):
NIce work!

But you should add a BREAKING CHANGE in there.. because component props have changed

JoanClaret:
yes, but at this moment, nobody is using this component. that's why i did not do it.

JoanClaret:
added sass variables in order to be able to customize from theme
https://github.com/SUI-Components/theme-basic/pull/68
https://github.schibsted.io/scmspain/frontend-mt--uilib-theme/pull/43

naxhh (approved):

components/breadcrumb/basic/src/index.js
```diff
@@ -1,43 +1,58 @@
...
-    </ul>
-  )
+export default class BreadcrumbBasic extends Component {
+  constructor (...args) {
+    super(...args)
+    this._breadcrumb = null
+  }
+
+  _expandBreadcrumb = () => {
+    this._breadcrumb.classList.add('is-expanded')
```
midudev: Why not using state for that?

components/breadcrumb/basic/src/index.js
```diff
@@ -1,43 +1,58 @@
...
-          </Link>
-          { index < numLinks && <IconAngle svgClass='sui-BreadcrumbBasic-icon' /> }
-        </li>
-      )}
-    </ul>
-  )
+export default class BreadcrumbBasic extends Component {
+  constructor (...args) {
+    super(...args)
+    this._breadcrumb = null
```
midudev: You just could do:

```
_breadcrumb = null
```

And you remove the constructor.

components/breadcrumb/basic/src/index.js
JoanClaret: i thought that using refs would be better option, but i might be wrong.
fixed!

components/breadcrumb/basic/src/index.js
JoanClaret: actually, since we are now using state, we don't need to initialize this value.

midudev (approved):

PR #103, opened 2017-06-29
feat(tab/basic): new tab basic component based on the old navigation tab

midudev:
Add new tab basic based on the old one (removed as it's not being used).

image


carlosvillu (approved):

naxhh (approved):

demo/tab/basic/playground
```diff
@@ -0,0 +1,20 @@
+const tabs = [
+  'Vincent',
+  'Jules',
+  'Marcellus',
+  'Mia',
+  'Winston',
+  'Butch'
```
JoanClaret, line 7: since they will be printed as a anchor links, should `tabs` contain an url?
maybe will be useful for SEO purposes

JoanClaret (approved):
Aside from a minor comment, looks fine to me!

ruben-martin-lozano (approved):

demo/tab/basic/playground
midudev: Truth is I was thinking to even use a button instead an anchor... :D I don't see the point to keep the `url` functionality anymore otherwise I should even consider React Router Links and I don't see that use right now.

PR #104, opened 2017-06-29
feat(tab/content): add new tab content component

midudev:
Add new tabbed content component to SUI Components.

image


components/tab/content/src/index.js
```diff
@@ -0,0 +1,90 @@
...
+          <section className={className} key={indexTab}>
+            {content}
+          </section>
+       ))
+      }
+      return contentToRender
+    }, [])
+  }
+
+  shouldComponentUpdate (nextProps, nextState) {
```
carlosvillu, line 33: 👍

components/tab/content/src/index.js
```diff
@@ -0,0 +1,90 @@
...
+      </div>
+    )
+  }
+}
+
+TabContent.displayName = 'TabContent'
+
+TabContent.defaultProps = {
+  activeTab: 0,
+  renderInactiveContent: true
```
carlosvillu, line 63: What does this ?!

carlosvillu (approved):

rackso (approved):

components/tab/content/src/index.js
midudev: For SEO reasons, we might want to render the HTML of the hidden tabs but usually we would want to avoid that and don't render the invisble content at all. This option offers you both possibilities.

PR #107, opened 2017-06-30
fix(alert/basic): import icons in a way that not all the lib is loaded

naxhh:
This way of import ended up in something like

```
var _lib = require('@schibstedspain/sui-svgiconset/lib')
var Info = _lib.Info
```

So we where importing all the lib just for this 3 icons


midudev (approved):

JoanClaret (approved):

midudev (approved):

midudev (approved):

midudev:
Hoy intento aprobarla de nuevo. :D xD

PR #108, opened 2017-07-03
Feature/breadcrumb adjust

JoanClaret:
- fixing button color
- adding `background color` and `box shadow` properties in order to fit with motor design requeriments

don't really know why tab-basic and tab-content is upgrading a new version.

### DIFF
--- a/components/breadcrumb/basic/package.json
+++ b/components/breadcrumb/basic/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@schibstedspain/sui-breadcrumb-basic",
-  "version": "1.3.0",
+  "version": "1.4.0",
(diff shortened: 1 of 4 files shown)


components/breadcrumb/basic/package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "@schibstedspain/sui-breadcrumb-basic",
-  "version": "1.3.0",
+  "version": "1.4.0",
```
naxhh: All package versions should be changed when release and not by hand, right?

naxhh (changes requested):
Remove package version changes

JoanClaret:
@naxhh fixed!

naxhh (approved):

PR #115, opened 2017-07-17
feat(thumbnail/basic): new thumbnail basic component

nguasch:
This PR creates the layout for this component:
screen shot 2017-07-17 at 13 22 44

Theme basic props defined here: https://github.com/SUI-Components/theme-basic/pull/73


components/thumbnail/basic/src/index.js
```diff
@@ -0,0 +1,25 @@
+import React, {Component} from 'react'
+import PropTypes from 'prop-types'
+
+class ThumbnailBasic extends Component {
+  render () {
+    const { image, captionText } = this.props
+    return (
+      <figure className='sui-ThumbnailBasic'>
+        <img className='sui-ThumbnailBasic-Image' src={image.src} alt={image.alt} />
```
JoanClaret: `sui-ThumbnailBasic-Image` should be `sui-ThumbnailBasic-image`.
same for `sui-ThumbnailBasic-Caption` => `sui-ThumbnailBasic-caption`

more info https://github.com/suitcss/suit/blob/master/doc/naming-conventions.md

components/thumbnail/basic/src/index.scss
```diff
@@ -0,0 +1,44 @@
+@import '~@schibstedspain/theme-basic/lib/index';
+
+.sui-ThumbnailBasic {
+  @include reset-basic;
+  background: $bgc-thumbnail;
```
JoanClaret: in order to be more consistent with other components, i would use `$bgc-thumbnail-basic`. same for others selectors.

components/thumbnail/basic/src/index.scss
```diff
@@ -0,0 +1,44 @@
...
+    color: $c-thumbnail-caption;
+    display: block;
+    font-size: $fz-thumbnail-caption;
+    left: 0;
+    max-height: 0;
+    opacity: .9;
+    overflow: hidden;
+    position: absolute;
+    text-align: center;
+    transition: $transition-thumbnail;
```
JoanClaret: what about using `trs-thumbnail-basic`?
more info https://docs.emmet.io/cheat-sheet/

components/thumbnail/basic/src/index.scss
```diff
@@ -0,0 +1,44 @@
...
+      display: block;
+      padding: $p-thumbnail-caption;
+    }
+  }
+
+  &:hover {
+    .sui-ThumbnailBasic-Caption {
+      max-height: 100%;
+    }
+  }
```
JoanClaret, line 43: i'm sure there is a reason for this, but i don't understand what does this max-height on hover.
could you please shed light on this?

components/thumbnail/basic/src/index.js
```diff
@@ -0,0 +1,25 @@
...
+        </figcaption>
+      </figure>
+    )
+  }
+}
+
+ThumbnailBasic.displayName = 'ThumbnailBasic'
+
+ThumbnailBasic.propTypes = {
+  image: PropTypes.object.isRequired,
```
naxhh: For better documentation would be a good idea to define the shape of the object

```
PropTypes.object.shape({
 src: PropTypes.string.isRequired,
 alt: PropTypes.string
})
```

components/thumbnail/basic/src/index.js
```diff
@@ -0,0 +1,25 @@
...
+  captionText: PropTypes.string
```
naxhh: Keep in mind that you have set this to default, but in the render you seem to expect the `captionText` to exist.

Or this is required or you should check if the var exists before creating the figcaption.

Check a similar example:  https://github.com/SUI-Components/sui-components/blob/master/components/card/basic/src/index.js#L38

JoanClaret (changes requested):
some minor comments, very good job anyways!
requesting changes only about the selector naming following SUIT naming convention.

demo/thumbnail/basic/context.js
```diff
@@ -0,0 +1,5 @@
+module.exports = {
```
naxhh: Why is this required?

midudev:
Good job! It looks nice for me besides the comment written here! :)

components/thumbnail/basic/src/index.scss
nguasch: 👍

components/thumbnail/basic/src/index.scss
nguasch: 👍

components/thumbnail/basic/src/index.scss
nguasch: It's because the transition. It's `max-height: 0` by default and goes to `max-height: 100%` on hover.

demo/thumbnail/basic/context.js
jcger: It's autogenerated by sui-studio.

In case you already knew it: I don't want to check if every file that sui-studio generates makes sense or not, I assume that they do or at least that they would tell me that they are optional (like the props in the src/index.js file), therefore this file still exists.

carlosvillu (approved):
Great work!

components/thumbnail/basic/src/index.js
jcger: You'r right, after talking with Norbert, it makes no sense for us to use this component without caption so the caption will be required

components/thumbnail/basic/src/index.scss
JoanClaret: aww, all right then! 👍

demo/thumbnail/basic/context.js
naxhh: context is only created if you pass the flag `-C` when generating the files.
I was curious to know about this one since the component doesn't really need it because is not using anything from context.

jcger: Ok, will remove it

naxhh (approved):

components/thumbnail/basic/src/index.js
nguasch: Ok. I getting a little crazy with this convention 😅

components/thumbnail/basic/.gitignore
```diff
@@ -0,0 +1,3 @@
+
+lib
+node_modules
```
davidbarna: No need as it's already in root https://github.com/SUI-Components/sui-components/blob/feature/thumbnail-basic/.gitignore

Also, i would remove this file and add lib to root's .gitignore.

davidbarna (approved):

JoanClaret (approved):

components/thumbnail/basic/.gitignore
jcger: I think this is something enablers/code owners should talk about and solve properly for all components, for now this file is being autogenerated by sui-studio.

I can create a sui-studio issue if it's ok for you.

davidbarna: "sui-studio generate" creates a general configuration. As sui-studio-create does not add gitignore, "sui-studio generate" can't assume you already added node_modules to the root folder.
But, if the root folder has a .gitignore, it makes sense you don't commit duplicated instruction in every single component.

davidbarna (approved):

PR #116, opened 2017-07-19
AB Testing components

davidbarna:
Migration of components from https://github.com/SUI-Components/react-ab-testing
Tests run was implmented for all the monorepo.


.babelrc
```diff
@@ -0,0 +1,3 @@
+{
```
jcger, line 1: Is there any special reason why you create this file instead of adding it in the package.json?

davidbarna: No. I don't remember but it may be so Jest uses if for sub packages, not only for current project.

components/abtesting/optimizelyXExperiment/src/index.js
```diff
@@ -0,0 +1,38 @@
+/* eslint handle-callback-err: "off" */
+import React, {Component, PropTypes} from 'react'
+import AbTestToggle from '@s-ui/abtesting-toggle'
+import {createExperimentUseCase} from './optimizely-x'
+
+class AbTestOptimizelyXExperiment extends Component {
+  constructor (props) {
```
midudev: Instead, you could just do:

```
state = { variation: undefined } // I think it's a good practice to initialize any state you're gonna use
```

components/abtesting/optimizelyXExperiment/src/index.js
```diff
@@ -0,0 +1,38 @@
...
+  }
+
+  componentDidMount () {
+    this._experiment = createExperimentUseCase(this.props.experimentId)
+    this._experiment.execute()
+      .then(variation => this.setState({variation}))
+  }
+
+  render () {
+    let {variation} = this.state
```
midudev: `const`.

components/abtesting/optimizelyXExperiment/src/optimizely-x.js
```diff
@@ -0,0 +1,65 @@
...
+   *  isActive: true,
+   *  isInExperimentHoldback: true,
+   *  variation: {
+   *      id: '101',
+   *      name: 'Variation #1'
+   *  }
+   * }
+   */
+  static async getInfo (experimentId) {
+    let sdk = await this.getSDK()
```
midudev, line 28: I personally prefer use `const` when the value of the variable is not gonna change, as it will enable your lint to help you to identify some problems (mutating for example variables that you don't want). In the long term it might have a performance boost but right now that's not important as all of that gets compiled. :D

Not important comment. :)

components/abtesting/toggle/package.json
```diff
@@ -0,0 +1,14 @@
+{
+  "name": "@s-ui/abtesting-toggle",
+  "version": "1.4.0",
+  "description": "",
+  "main": "lib/index.js",
+  "scripts": {
+    "build": "rm -Rf ./lib && mkdir -p ./lib && npm run babel",
+    "babel": "../../../node_modules/.bin/babel --presets sui ./src --out-dir ./lib"
```
midudev: `build:js` to keep a consistency.

components/abtesting/toggle/src/index.js
```diff
@@ -0,0 +1,43 @@
+/* eslint eqeqeq: "off" */
+
+import React, {Component, PropTypes} from 'react'
+
+class AbTestToggle extends Component {
+  render () {
+    let {variation, children} = this.props
+
+    var filterFunc
```
midudev: `let`.

components/abtesting/toggle/src/index.js
```diff
@@ -0,0 +1,43 @@
...
+    if (variation) {
```
midudev:
```
const filterFunc = variation
  ? child => child.props.variationId == variation // why not a strict one?
  : child => child.props.defaultVariation

File: components/abtesting/optimizelyXExperiment/src/index.js (hunk 2 above)
Comment (naxhh) [reply components/abtesting/optimizelyXExperiment/src/index.js]: +1
At least is nice to know in a single glance what is in the state.

File: package.json [hunk 8]
```diff
@@ -34,6 +38,7 @@
       "name": "SUIComponents"
     },
     "sui-mono": {
+      "access": "public",
```
naxhh, line 41: good catch

naxhh:
I have a few comments regarding this just to clarify points.

Logic in component:
This is the very first component with "business" logic inside it, while I understand why, is a way we want to take for other components? or this is just an exception? In that case maybe that should be very clear somewhere.

Jest for testing, and testing:
Until this PR there's no testing anywhere nor a definition on what we should test and how. Jest has been added globally in the studio. Does this means we should tests all components? Unit tests or E2E test? etc...

While I understand why we are adding those, maybe we should make clear somewhere if this are exceptions or is the new standard for components.

naxhh (approved):

davidbarna:
@naxhh thanks so much for your comments. Let me answer.

**Logic in component:** you're right. We're not going to do a domain for optimizely. Although, we could have done an abstraction or something. The thing is that we talked with @miduga about migrating from https://github.com/SUI-Components/react-ab-testing to here but what you say clearly makes me think about discarding this PR.

**Jest for testing, and testing**
My components were already tested because I use too. And, there was optimizely logic that had to be tested. There's no prohibition to test components I guess. Nevertheless, having jest in there may confuse the rest of the teams.

@miduga and @carlosvillu please give your opinion.

components/abtesting/optimizelyXExperiment/src/index.js
davidbarna: Much cleaner that constructor and good to mention variation prop. ✅

components/abtesting/optimizelyXExperiment/src/optimizely-x.js
davidbarna: Well it's a mutable object I prefer not to. I know the asignation is constant. Not relevant anyways.

components/abtesting/toggle/src/index.js
davidbarna: I hate multiline ternaries but this one is so cute. Changed.
Comparaison is abstract cause it can be string or number (check propTypes)

midudev:
About logic, I would say that's fine in this case, as it's a scoped component with logic attached to a external dependency. We're not going to create a domain for connecting with newRelic IMHO.

About testing, I would say that Jest is were we want to go. We definitely are not going to test all our components now with Jest, but if we have to create or define a new way to do it, it would be definitely Jest.

For me this component could be a fine exception of all the rules we have for the UI components.

components/abtesting/optimizelyXExperiment/src/optimizely-x.js
midudev: When you say it's a mutable object, it seems that you're changing the value of the variable and I would say that you're not doing it. Making it a `const` would give you an error in the linter if you try to reassign the `sdk` when you already know is something you want to avoid. It's a effortless change that could avoid problems. :P

davidbarna: BTW I changed it [here](https://github.com/SUI-Components/sui-components/commit/71b6b4a21b1250ca80fa04785577b765f9187ca3#diff-c271d8346f679e021e3b8f11a422901dL54). I don't have a strong opinion about this and I get that we're defining an immutable reference to an object. Maybe we should talk altogether and decide something as a convention. Actually I don't event like to name consts without uppercase. Maybe I'm just old school.

PR #118, opened 2017-07-24
Cover basic

betobe:
Display an image with width=100%. It also shows buttons if defined.


components/cover/basic/package.json
```diff
@@ -0,0 +1,17 @@
+{
+  "name": "@schibstedspain/sui-cover-basic",
+  "version": "1.0.0",
+  "description": "",
+  "main": "lib/index.js",
+  "scripts": {
+    "build": "rm -Rf ./lib && mkdir -p ./lib && npm run babel && npm run sass",
+    "babel": "../../../node_modules/.bin/babel --presets sui ./src --out-dir ./lib",
```
midudev: Be sure to use latest `suistudio` version, as it has different scripts name.

components/cover/basic/src/index.scss
```diff
@@ -0,0 +1,72 @@
...
+  }
+
+  &-gradient {
+    display: inline-block;
+    position: relative;
+    width: 100%;
+  }
+
+  &-gradient:after {
+    background: -moz-linear-gradient(top, rgba(0, 0, 0, 0) 0%, rgba(0, 0, 0, 0.65) 100%); /* FF3.6+ */
```
midudev: Autoprefixer will add automatically the prefixes you need for your supported browsers, so no need to add all here.

betobe: 👍

components/cover/basic/src/index.scss
```diff
@@ -0,0 +1,72 @@
+@import '~@schibstedspain/theme-basic/lib/index';
+
+.sui-CoverBasic {
+  max-height: 676px;
+  position: relative;
+  width: 100%;
+
+  &.clickable {
```
JoanClaret: since this is a state, should it be `is-clickable`?

betobe: 👍

components/cover/basic/package.json
betobe: Ok. I was using 4.10 and now I have 4.11.

Note that `sui-studio link` is failing because it is trying to use `babel` instead of `build:js`

components/cover/basic/src/index.scss
```diff
@@ -0,0 +1,70 @@
...
+    display: block;
+    height: 100%;
+    object-fit: cover;
+    width: 100%;
+  }
+
+  &-buttonList {
+    display: flex;
+    flex-direction: row;
+    margin: 30px 0 0 52px;
```
JoanClaret: if possible, we should try to avoid using magic numbers.
what about use theme margin declarations? `$m-xlarge`?

components/cover/basic/src/index.scss
```diff
@@ -0,0 +1,70 @@
...
+    position: absolute;
+    top: 0;
+
+    button {
+      background-color: $c-white;
+      border: 0;
+      border-radius: 4px;
```
JoanClaret: better using `$bdrs-base`

components/cover/basic/src/index.scss
```diff
@@ -0,0 +1,70 @@
...
+      color: $c-black;
+      font-size: $fz-cover-basic-button;
+      line-height: 1.5;
+      margin-right: 10px;
+      min-width: 90px;
+      min-height: 40px;
+      opacity: 0.75;
+      padding: 0 15px;
```
JoanClaret: if possible, all this declarations should use theme basic variables such as `$lh-base`
for margins, i would define `$m-small, $m-xsmall, $m-xxsmall` as @nguasch did with paddings in
https://github.com/SUI-Components/theme-basic/pull/73/files#diff-c4aed6ba387ce119c150c70302460cc6R123

components/cover/basic/src/index.scss
```diff
@@ -0,0 +1,70 @@
...
+
+      &:hover {
+        cursor: hand;
+        opacity: 1;
+      }
+
+      svg {
+        margin-right: 8px;
```
JoanClaret: same as commented above.

components/cover/basic/src/index.scss
```diff
@@ -0,0 +1,70 @@
+@import '~@schibstedspain/theme-basic/lib/index';
+
+$mah-cover-basic: 676px !default;
+$fz-cover-basic-button: 16px !default;
```
JoanClaret: since this is a sui-component, what do you think about defining this variables in `theme-basic`?

components/cover/basic/src/index.js
```diff
@@ -0,0 +1,90 @@
...
+
+const CoverBasic = (props) => {
+  const buttons = buildButtons(props)
+
+  // Custom css class is attached when component is clickable.
+  const isClickableClass = isClickable(props) ? ' is-clickable' : ''
+
+  return (
+    <div className={'sui-CoverBasic' + isClickableClass} onClick={buildClickHandler(props)}>
+      <div className={(props.gradient ? 'sui-CoverBasic-gradient' : '')}>
```
JoanClaret: we usually use `cx` to manage classNames. i think it would be more readable. See example below.
https://github.com/SUI-Components/sui-components/blob/master/components/breadcrumb/basic/src/index.js#L13

components/cover/basic/src/index.scss
JoanClaret: if this does not fit with your needs, we could define this variables in theme-basic.

components/cover/basic/src/index.scss
```diff
@@ -0,0 +1,70 @@
...
+
+  &-buttonList {
+    display: flex;
+    flex-direction: row;
+    margin: 30px 0 0 52px;
+    position: absolute;
+    top: 0;
+
+    button {
+      background-color: $c-white;
```
JoanClaret: could we use buttons defined in `theme-basic`? 
something like `@include sui-button('ghost', 'large', 'full');`
https://github.com/SUI-Components/theme-basic/blob/master/src/components/_button.scss

components/cover/basic/src/index.scss
```diff
@@ -0,0 +1,70 @@
+@import '~@schibstedspain/theme-basic/lib/index';
```
JoanClaret, line 1: in order to aviod linting errors with `is-clickable' class name, you should insert this line at the top of the document

`/* sass-lint:disable no-important, class-name-format*/`

components/cover/basic/src/index.js
```diff
@@ -0,0 +1,90 @@
...
+        <img className='sui-CoverBasic-image' src={props.src} />
+      </div>
+      <div className='sui-CoverBasic-buttonList'>
```
jcger: You could win some render time by checking if there really is some button to show

components/cover/basic/src/index.js
```diff
@@ -0,0 +1,90 @@
...
+    isClickable(props) && props.handleClick(event)
+  }
+}
+
+const isClickable = (props) => {
+  return !!props.handleClick
+}
+
+const buildButtons = (props) => {
+  return props.buttons.map((item, index) => {
```
julian-gernun: As a different component, why not in a new file?

components/cover/basic/src/index.js
```diff
@@ -0,0 +1,90 @@
...
+    // Build click handler for each button
+    function handler (event) {
+      event.stopPropagation()
+      item.handleClick(event)
+    }
+
+    return (
```
jcger: As a different component, why not in a different file?

components/cover/basic/src/index.js
betobe: 👍

components/cover/basic/src/index.js
betobe: "Nice" but not really needed. I think it is the simplest way.

components/cover/basic/src/index.scss
betobe: Is it really needed? It just displays a warning. And if it is really allowed, is the linter setup wrong?

components/cover/basic/src/index.js
betobe: 👍

components/cover/basic/src/index.scss
betobe: I would rather to leave it here because right now it is too component-especific and reusability is not clear at all.

components/cover/basic/src/index.scss
davidbarna: IMO controlling warnings and avoiding them to display is a good practice and with tiny cost. I usually do it.

components/cover/basic/src/index.scss
davidbarna: I think that should come from the designer. They should deliver mockups with normalized spacing... 
Developer should worry about pixel perfect. But, dev should also check if there is a way to do it with our current variables. At least, hardcoded paddings and margins should not happen.
Can't you normalize them with the designer?

components/cover/basic/src/index.scss
betobe: I think the opposite. I think it is a bad practice because 1)you might be hiding a possible risk 2)you might be overriding an obsolete rule in the wrong place 3)you might be damaging a future attempt to improve linting rules or linting reports.

components/cover/basic/src/index.scss
```diff
@@ -0,0 +1,50 @@
...
+    left: 0;
+    position: absolute;
+    top: 0;
+    width: 100%;
+  }
+
+  &-image {
+    display: block;
+    height: 100%;
+    object-fit: cover;
```
ronaldo-mh, line 43: object-fit is not supported in any version of IE.

JoanClaret: Well seen!

components/cover/basic/src/index.js
```diff
@@ -0,0 +1,97 @@
...
+  const buttons = buildButtons(props)
+
+  // Custom css class is attached when component is clickable.
+  const coverBasicClassNames = cx('sui-CoverBasic', {
+    'is-clickable': isClickable(props)
+  })
+
+  return (
+    <div className={coverBasicClassNames} onClick={buildClickHandler(props)}>
+      <div className={(props.gradient ? 'sui-CoverBasic-gradient' : '')}>
```
ronaldo-mh: Maybe use cx here?
```
<div className={cx({'sui-CoverBasic-gradient' : props.gradient})}>
```

components/cover/basic/src/index.scss
betobe: Yep, I know... I still need to find out how to solve it

components/cover/basic/src/index.js
betobe: 👍

components/cover/basic/src/index.scss
davidbarna: You're member the of WDO (warnings' defenders organization), but I get your point. I think you're right.

components/cover/basic/.gitignore
```diff
@@ -0,0 +1,3 @@
+
+lib
+node_modules
```
davidbarna: not needed as already in main gitignore.

components/cover/basic/src/index.scss
davidbarna: Check out who may have a solution for you :-)
https://github.com/tonipinel/object-fit-polyfill
It's not a good solution at all. But if you really can't find a way to do it with css, it may inspire...

components/cover/basic/src/index.scss
naxhh: "its just a warning" is also bad, since people will tend to accumulate them and at some point we will not be able to see errors in a glance.
Or just not event take care of them

So, I think (outside of this pr) we fix the linter rules, we make that specific line not trigger a warning or we came up with another solution

I think letting the lint add up a lot of warnings will tend users to not care about them.
And disable for the full file will have it's problems too as Beto said

davidbarna: Now I agree with Nacho. Big lack of personality. Will work on it XD

PR #119, opened 2017-07-25
feat(title/basic): first version of basic title and subtitle component.

ronaldo-mh:


components/title/basic/src/index.js
```diff
@@ -0,0 +1,52 @@
...
+export default function TitleBasic ({align, className, subtitle, title}) {
+  function buildAlignClassName (alignProp) {
+    let alignClassName = ''
+    if (alignProp) {
+      alignClassName = 'sui-TitleBasic--align' + alignProp.charAt(0).toUpperCase() + alignProp.slice(1)
+    }
+    return alignClassName
+  }
+
+  const titleBasicClassName = cx('sui-TitleBasic', className, {
```
betobe: It could be shortened, right?
```
const titleBasicClassName = cx('sui-TitleBasic', className, `${buildAlignClassName(align)}`)
```

components/title/basic/src/index.js
```diff
@@ -0,0 +1,52 @@
...
+   */
+  align: PropTypes.oneOf([
+    'left',
+    'center',
+    'right'
+  ]),
+  /**
+   * CSS classNames to apply to Title Basic container.
+   */
+  className: PropTypes.node,
```
betobe: I think it should be PropTypes.string instead.

components/title/basic/src/index.scss
```diff
@@ -0,0 +1,48 @@
+@import '~@schibstedspain/theme-basic/lib/index';
+
+$c-title-basic: $c-white;
+$c-subtitle-basic: $c-white;
+
+$ff-roboto-sans-serif: 'Roboto', sans-serif; // Will be moved to nc-theme-basic
```
betobe: We still need to clarify where the component vars should be really defined (here or themebasic). In any case I would name this var like  `$ff-title-basic: $ff-base` (actually $ff-base doesn't exist and the only similar var is $ff-sans-serif)

components/title/basic/src/index.scss
```diff
@@ -0,0 +1,48 @@
...
+.sui-TitleBasic {
+  display: flex;
+  flex-direction: column;
+  font-family: $ff-roboto-sans-serif;
+
+  &-title {
+    color: $c-title-basic;
+    font-size: $fz-title-basic;
+    font-weight: $fw-title-basic;
+    margin: 0;
```
betobe: Margins could be removed (by the moment) or parametrized.

components/title/basic/src/index.js
```diff
@@ -0,0 +1,52 @@
+import React, {PropTypes} from 'react'
+import cx from 'classnames'
+
+export default function TitleBasic ({align, className, subtitle, title}) {
+  function buildAlignClassName (alignProp) {
+    let alignClassName = ''
+    if (alignProp) {
+      alignClassName = 'sui-TitleBasic--align' + alignProp.charAt(0).toUpperCase() + alignProp.slice(1)
```
betobe: Consider simplifying this OR adding a comment :-p

components/title/basic/src/index.scss
JoanClaret: ~~agree with @betobe . also, the name of the variable should be `ff-title-basic`~~

components/title/basic/src/index.js
```diff
@@ -0,0 +1,52 @@
+import React, {PropTypes} from 'react'
+import cx from 'classnames'
+
+export default function TitleBasic ({align, className, subtitle, title}) {
+  function buildAlignClassName (alignProp) {
```
JoanClaret: since this is a private function, i would define this function as `_buildAlignClassName`

components/title/basic/src/index.scss
```diff
@@ -0,0 +1,48 @@
...
+  }
+
+  &-subtitle {
+    color: $c-subtitle-basic;
+    font-size: $fz-subtitle-basic;
+    font-weight: $fw-subtitle-basic;
```
JoanClaret: All SCSS variables should be defined with the main component name. for instance: 
`$fw-component-name-descendent-name`

JoanClaret: more examples: `$fw-title-basic-main-title`, `$fw-title-basic-subtitle` ...

components/title/basic/src/index.scss
JoanClaret: thinking about it, what about using the theme-basic titles definition?
https://github.com/SUI-Components/theme-basic/blob/master/src/_typography.scss

components/title/basic/src/index.js
ronaldo-mh: Yep, agreed and changed. Thanks!

components/title/basic/src/index.js
ronaldo-mh: Don't know sure why I had the idea that react supports strings and arrays in className property, but reviewing the documentation it seems it only supports strings. 
Anyway, I have just changed this, thanks!

components/title/basic/src/index.js
ronaldo-mh: This is just a method to build the modifier classname to apply to the component (according to SUIT naming convention, it must be separated with "--" prefix and written in camel case).
I can reduce this method, but I don't think it will help since obfuscate is not the same as simplify.
I will add a comment on this function to specify its main purpose.

components/title/basic/src/index.scss
ronaldo-mh: Ok, I will remove the margins defined, just we need to remember to define the proper margins in the theme basic we create in the future for nc, or create "reset-margins" mixin in it.

components/title/basic/src/index.scss
ronaldo-mh: Ok, I have fixed the variables naming to have the component name as prefix. Thanks!

components/title/basic/src/index.scss
ronaldo-mh: I have simplified the use of variables in this component, in order to be aligned with the theme-basic _typography definition (h1 and h2) and _settings (font-family, font-size and font-weight).
I will just maintain the variables for the color of the title and subtitle, which can be overwritten in each basic-theme as required.

JoanClaret: in my opinion would be more easy and more maintainable if we use `@include h1`, `@include h2`...
more opinions? @betobe @miduga ?

ronaldo-mh: Actually, the title element is an h1 and the subtitle is an h2. So it would be a bit redundant if we use the mixin as well.
I will keep the styles for h1 and h2 already defined in theme-basic, and the scss file will contain only the custom styles required:

```
.sui-TitleBasic {
  &-title {
    color: $c-title-basic-main-title;
    font-weight: $fw-semi-bold;
  }
  &-subtitle {
    color: $c-title-basic-subtitle;
    font-weight: $fw-regular;
  }
```

components/title/basic/src/index.js
ronaldo-mh: Done!

components/title/basic/src/index.js
```diff
@@ -0,0 +1,55 @@
+import React, {PropTypes} from 'react'
+import cx from 'classnames'
+
+export default function TitleBasic ({align, className, subtitle, title}) {
+  /**
+   * Builds the modifier className to apply to the component following the SUIT naming convention (sui-TitleBasic--modifierInCamelCase).
+   * @param alignProp {string} property value for Title alignment.
+   * @returns {string} SUIT formatted classname.
+   */
+  function _buildAlignClassName (alignProp) {
```
betobe: It is not a private function but a function defined inside another function. IMO _ is not needed

components/title/basic/src/index.scss
JoanClaret: then, there is no `font-size` definition or i'm missing something?

components/title/basic/src/index.scss
```diff
@@ -0,0 +1,29 @@
+@import '~@schibstedspain/theme-basic/lib/index';
+
+$c-title-basic-main-title: $c-white !default;
+$c-title-basic-subtitle: $c-white !default;
```
JoanClaret: white by default? looks a bit weird.

components/title/basic/src/index.scss
```diff
@@ -0,0 +1,29 @@
...
+.sui-TitleBasic {
+
+  &-title {
+    color: $c-title-basic-main-title;
+    font-weight: $fw-semi-bold;
+  }
+
+  &-subtitle {
+    color: $c-title-basic-subtitle;
+    font-weight: $fw-regular;
```
JoanClaret: no font-size nor line height? why not `@include h2`?

components/title/basic/src/index.js
davidbarna: It's a pure component, so you are creating this func on every render. You can move it before declaration as a const func.

ronaldo-mh: Done!

components/title/basic/src/index.scss
```diff
@@ -0,0 +1,31 @@
+@import '~@schibstedspain/theme-basic/lib/index';
+
+$c-title-basic-main-title: $c-primary !default;
+$c-title-basic-subtitle: $c-primary !default;
```
betobe: Will you move this to theme basic?

components/title/basic/src/index.js
```diff
@@ -0,0 +1,51 @@
+import React, {PropTypes} from 'react'
+import cx from 'classnames'
+
+/**
+ * Builds the modifier className to apply to the component following the SUIT naming convention (sui-TitleBasic--modifierInCamelCase).
+ * @param alignProp {string} property value for Title alignment.
+ * @returns {string} SUIT formatted classname.
+ */
+const _buildAlignClassName = (alignProp) => {
```
betobe: Still not a private :p

ronaldo-mh: I know it does not exist the "private" concept in js, and although I am against this as well as you, I have seen that other similar functions in other sui-components includes the underscore dangle at the beggining of the functions names, so I will maintain this naming convention.

components/title/basic/src/index.scss
ronaldo-mh: Yes, I think it will be better to override this var $c-title-basic-main-title in each theme-basic, but with these definitions we ensure to have at least a default value.

components/title/basic/src/index.js
```diff
@@ -0,0 +1,51 @@
...
+  return alignProp ? 'sui-TitleBasic--align' + alignProp.charAt(0).toUpperCase() + alignProp.slice(1) : ''
```
naxhh: Since you have a list of valid align values, why not have a map lookup instead of slices?

I think will be a bit faster? not sure and I don't think really important though

components/title/basic/src/index.js
```diff
@@ -0,0 +1,51 @@
...
+    'right'
+  ]),
+  /**
+   * CSS classNames to apply to Title Basic container.
+   */
+  className: PropTypes.string,
+  /**
+   * Title to display.
+   */
+  title: PropTypes.string.isRequired,
```
naxhh, line 35: This is required but you are checking if it exists on the component before rendering

I think if it's required no check should be done in the render.

naxhh (approved):

JoanClaret (approved):

components/title/basic/src/index.js
ronaldo-mh: I agree with this, it is a bit redundant, so I remove the validation in render function.

PR #120, opened 2017-07-25
feat(title/multisize): first version of multisize title component

ronaldo-mh:
... which supports up to 3 different title sizes.


components/title/multisize/src/index.scss
```diff
@@ -0,0 +1,60 @@
+@import '~@schibstedspain/theme-basic/lib/index';
+
+$c-title-multisize-primary: #a81c4f !default;
```
JoanClaret: what about using theme-basic colors?

components/title/multisize/src/index.scss
```diff
@@ -0,0 +1,60 @@
...
+
+$fz-title-multisize-pre-title: $fz-h2 !default;
+$fz-title-multisize-main-title: 48px !default;
+$fz-title-multisize-post-title: $fz-h2 !default;
+
+$maw-phone: 480px !default; // Will be moved to nc-theme-basic
+
+$p-h-title-multisize-between-elements: $p-h-small;
+
+$phone: 'screen and (max-width: $maw-phone)'; // Will be moved to nc-theme-basic
```
JoanClaret: rather than use this mq, i would use theme-basic mq:
https://github.com/SUI-Components/theme-basic/blob/master/src/utils/_breakpoints.scss

also, theme-basic defined breakpoints:
https://github.com/SUI-Components/theme-basic/blob/master/src/_settings.scss#L137

for instance:

```
@include media-breakpoint-down(s) {
    align-self: flex-start;
    padding: 0;
    text-align: left;
}
```

components/title/multisize/src/index.scss
```diff
@@ -0,0 +1,60 @@
...
+$c-title-multisize-secondary: $c-gray !default;
+
+$ff-roboto-sans-serif: 'Roboto', sans-serif; // Will be moved to nc-theme-basic
+
+$fw-title-multisize-pre-title: 500 !default;
+$fw-title-multisize-main-title: 500 !default;
+$fw-title-multisize-post-title: $fw-regular !default;
+
+$fz-title-multisize-pre-title: $fz-h2 !default;
+$fz-title-multisize-main-title: 48px !default;
```
JoanClaret: if possible, better using https://github.com/SUI-Components/theme-basic/blob/master/src/_typography.scss

components/title/multisize/src/index.js
```diff
@@ -0,0 +1,33 @@
...
+
+TitleMultisize.propTypes = {
+  /**
+   * CSS classNames to apply to Title Multisize container.
+   */
+  className: PropTypes.node,
+  /**
+   * Text to display after main title (h3).
+   */
+  postTitle: PropTypes.node,
```
carlosvillu: That is an array of strings and nodes

carlosvillu (approved):

components/title/multisize/src/index.scss
ronaldo-mh: Ok, I am changing the styles variables definitions in order to use the theme-basic colors defined.

components/title/multisize/src/index.scss
ronaldo-mh: Thanks a lot! I did not find these mq in the basic theme.
I am changing everything to adapt it to the defined breakpoints.

components/title/multisize/src/index.scss
ronaldo-mh: Done, I have adapted all font-size and font-weight to the ones defined in theme-basic _typography.

components/title/multisize/src/index.js
ronaldo-mh: Yep, this is an array of strings and elements, but this is exactly what PropTypes.node is:

PropTypes.node
// Anything that can be rendered: numbers, strings, elements or an array
// (or fragment) containing these types.

components/title/multisize/src/index.js
```diff
@@ -0,0 +1,79 @@
...
+/**
+ * Builds an HTML/React Element depending on the size declared (s: H3, m: H2, l: H1).
+ * @param size {string} Size (s, m, l) of title element.
+ * @param className {string} CSS classNames to apply to element.
+ * @param content {element} Children of title element. Could be an element (HTML/React Element), string, or an array of both.
+ * @returns {element} React Title Element.
+ * @private
+ */
+const _buildTitleElement = (size, className, content) => {
+  let element = null
```
midudev:
```
// const possibly outside the component
const TAG_FOR_SIZE = {
'l' : 'h1',
'm': 'h2',
's': 'h3'
}
// get the tag that you need
const Tag = TAG_FOR_SIZE[size]
// return the element
return content && <Tag className={className}>{content}</Tag>
```

JoanClaret (approved):

components/title/multisize/.gitignore
```diff
@@ -0,0 +1,3 @@
+
+lib
+node_modules
```
davidbarna, line 3: You may not need this as it's on the root gitgnore.

components/title/multisize/README.md
```diff
@@ -0,0 +1,12 @@
+
```
davidbarna, line 1: Thanks for documenting 👍

components/title/multisize/src/index.js
ronaldo-mh: I had a previous version similar to this, but changed it as an if with the 3 sizes was more explicit. I any case, I can do it like this as well, thanks!

components/title/multisize/src/index.js
```diff
@@ -0,0 +1,79 @@
...
+  className: PropTypes.string,
+  /**
+   * Text to display after main title (h3).
+   */
+  postTitle: PropTypes.node,
+  /**
+   * Size of post-title: s = h3, m = h2, l = h1.
+   */
+  postTitleSize: PropTypes.oneOf([
+    's', 'm', 'l'
```
davidbarna: Not important but you can define the sizes a a const array. They you'll have only on array instance for this.

davidbarna (approved):

components/title/multisize/src/index.js
```diff
@@ -0,0 +1,79 @@
...
+
+/**
+ * Builds an HTML/React Element depending on the size declared (s: H3, m: H2, l: H1).
+ * @param size {string} Size (s, m, l) of title element.
+ * @param className {string} CSS classNames to apply to element.
+ * @param content {element} Children of title element. Could be an element (HTML/React Element), string, or an array of both.
+ * @returns {element} React Title Element.
+ * @private
+ */
+const _buildTitleElement = (size, className, content) => {
```
jcger: This isn't any class method, no oop context, so it can't be private

components/title/multisize/src/index.js
```diff
@@ -0,0 +1,79 @@
...
+  // get the corresponding tag for requested size
+  const Tag = TAG_FOR_SIZE[size]
+  // return the element if has content title
```
jcger: Are this and the  comment in line 20 really needed?

components/title/multisize/src/index.js
ronaldo-mh: Ok, I will reuse the TAG_FOR_SIZE object for this purpose.

components/title/multisize/src/index.js
ronaldo-mh: Removed! I had not seen it, I guess it was autogenerated by the WebStorm's jsDoc when it detected the underscore dangle in this function name.

components/title/multisize/src/index.js
ronaldo-mh: Not really, I will remove them in the next commit.

demo/title/multisize/playground
```diff
@@ -0,0 +1,13 @@
+return (
+    <div>
+        <TitleMultisize preTitle="desde" title="250.000 €"  postTitle={['1.234 €/m', <sup>2</sup>]} titleSize="l"/>
```
jcger: Executing this, you got that ugly `Each child in an array or iterator should have a unique "key" prop. Check the render method of TitleMultisize. See https://fb.me/react-warning-keys for more information` warning, looks like adding key attribute to the `<sup>` tags fixes it

jcger (approved):

demo/title/multisize/playground
ronaldo-mh: Yep, but the "sup" tag is not part of the multisize component itself, it is now written in the playground's definition (to emulate the final lookup of nc project UX design).
I know that when we create the component's wrapper we must add the key prop in the array's elements.
But for the moment, I will add it explicitly in the playground to remove this warning.

PR #123, opened 2017-07-28
feat(list/masonry): init masonry component

JoanClaret:
ListMasonry is a React grid layout component. It works by placing elements in optimal position based on available vertical space.

```javascript
const breakpoints = [600, 900]

<ListMasonry
  breakpoints={breakpoints}
  className='extra-class'
/>
```

Check the playground in order to get a complete view of the possibilities of the component. Try resizing your browser and check how this component build columns based on the breakpoints array passed via props.

image


components/list/masonry/src/index.js
```diff
@@ -0,0 +1,63 @@
...
+import cx from 'classnames'
+
+class ListMasonry extends React.Component {
+  constructor (props) {
+    super(props)
+    this.state = {columns: 1}
+  }
+  componentDidMount () {
+    this._onResize()
+    window.addEventListener('resize', this._onResize)
```
carlosvillu: Is missing the removeEventListener

JoanClaret: fixed 30 sec ago
: )

carlosvillu (approved):

components/list/masonry/src/index.js
```diff
@@ -0,0 +1,67 @@
+import React, { PropTypes } from 'react'
+import cx from 'classnames'
+
+class ListMasonry extends React.Component {
+  constructor (props) {
```
midudev: Simpler without constructor:
`state = { columns: 1}`

Should be 1 a prop? At least a const?

components/list/masonry/src/index.js
```diff
@@ -0,0 +1,67 @@
...
+  _getColumns (width) {
+    return this.props.breakPoints.reduceRight((collector, current, index) => {
+      return current < width ? collector : index
+    }, this.props.breakPoints.length) + 1
+  }
+
+  _onResize = () => {
+    const columns = this._getColumns(this.refs.ListMasonry.offsetWidth)
+    if (columns !== this.state.columns) {
+      this.setState({columns: columns})
```
midudev: Same name of key and value, not needed the latter.

JoanClaret: 👍

components/list/masonry/src/index.js
JoanClaret: i don't see that as a prop so i defined this as a const. 
thanks for the feedback!

naxhh (approved):
Great work!

PR #125, opened 2017-08-02
Cover basic

betobe:
- Accept className as property.
- Basic responsive feature.

### DIFF
--- a/components/cover/basic/src/index.scss
+++ b/components/cover/basic/src/index.scss
@@ -2,6 +2,14 @@
 @import '~@schibstedspain/sui-button-basic/lib/index';
 
 .sui-CoverBasic {
+  @include media-breakpoint-down(s) {
+    height: $hmediabreaks-cover-basic;
+
+    .sui-CoverBasic-buttonList {
+      display: none;
+    }
+  }
+
   max-height: $mah-cover-basic;
   position: relative;
   width: 100%;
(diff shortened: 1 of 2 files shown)


components/cover/basic/src/index.scss
```diff
@@ -2,6 +2,14 @@
 @import '~@schibstedspain/sui-button-basic/lib/index';
 
 .sui-CoverBasic {
+  @include media-breakpoint-down(s) {
+    height: $hmediabreaks-cover-basic;
```
betobe: https://github.com/SUI-Components/theme-basic/pull/78

midudev (approved):

components/cover/basic/src/index.scss
JoanClaret: what do you think about using `h-cover-basic-small`?

JoanClaret (approved):

carlosvillu (approved):

components/cover/basic/src/index.scss
betobe: I'll change it (because that name I gave it is awful) but I'm still not sure what is best. I wanted to mean "height value applied only on breakpoint 's'" and I couldn't find a better any better name. On the other hand "h-cover-basic-small" is ambiguous for me because it sounds like the "general height value is 'small'" instead.

PR #129, opened 2017-08-16
chore(META): change travis config to execute on every PR

davidbarna:
Changed travis config so I can test current issue executing build in PR.
Also, if this config works, then we'll leave it as it is and all PR will have a building step

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -9,9 +9,6 @@ before_install:
   - npm i -g npm@^3.0.0
   - npm i -g surge
   - npm config set //registry.npmjs.org/:_authToken $NPM_TOKEN
-install:
-  - npm i
-  - sui-studio run-all npm i
 before_deploy:
   - npm run build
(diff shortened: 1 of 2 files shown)


.travis.yml
```diff
@@ -9,13 +9,16 @@ before_install:
   - npm i -g npm@^3.0.0
   - npm i -g surge
   - npm config set //registry.npmjs.org/:_authToken $NPM_TOKEN
-after_success:
+install:
+  - npm i
```
carlosvillu: you dont need this.

carlosvillu (approved):

.travis.yml
davidbarna: I had to replace default "install" to add the "run-all" step. But I'll remove it and add run-all to the postinstall. Makes sense.

carlosvillu (approved):

PR #134, opened 2017-08-23
Feature/tag selectable list

davidmartin84:
A React component that show a selectable tag list


components/tag/selectableList/src/index.js
```diff
@@ -0,0 +1,109 @@
...
+
+  constructor (...props) {
+    super(...props)
+    this.tagHandlers = this.props.tagsList.map((tag) => this.toggleValue.bind(this, tag.value))
+    this.allHandler = this.toggleAll.bind(this)
+  }
+
+  toggleValue (value) {
+    let selectedValues = this.state.selectedValues
+    if (selectedValues.indexOf(value) !== -1) {
```
naxhh: Small issue but you can use `includes`

components/tag/selectableList/src/index.js
```diff
@@ -0,0 +1,109 @@
...
+
+TagSelectableList.displayName = 'TagSelectableList'
+
+TagSelectableList.propTypes = {
+  /**
+   * Icon to check tags
+   */
+  checkIcon: PropTypes.func,
+  /**
+   * Optional string. When is setted show a button `All` at first position
```
naxhh: You should specify that the string content is the label of the button

components/tag/selectableList/src/index.js
```diff
@@ -0,0 +1,109 @@
...
+}
+
+export default class TagSelectableList extends Component {
+  state = {
+    selectedValues: []
+  }
+
+  constructor (...props) {
+    super(...props)
+    this.tagHandlers = this.props.tagsList.map((tag) => this.toggleValue.bind(this, tag.value))
```
naxhh: If you want to avoid this bind (slow)

you can define toggleValue with fat arrow and it will be bind to this by defaullt.

```
const toggleValue = (value) => { ... }
```

And if you need to return a function instead of the computation result, use currying:

```
const toggleValue = (value) => () => { ... }
```

components/tag/selectableList/src/index.js
```diff
@@ -0,0 +1,109 @@
...
+    let selectedValues = this.state.selectedValues
+    if (selectedValues.indexOf(value) !== -1) {
+      selectedValues = removeFromArray(selectedValues, value)
+    } else {
+      selectedValues.push(value)
+    }
+    this.setState({selectedValues})
+  }
+
+  toggleAll () {
```
naxhh: What about:

```
this.setState({
  selectedValues = this.isAllSelected()
    ? []
    : this.props.tagsList.map(item => item.value)
})
```

components/tag/selectableList/src/index.js
```diff
@@ -0,0 +1,109 @@
...
+    let selectedValues
+    if (this.isAllSelected()) {
+      selectedValues = []
+    } else {
+      selectedValues = this.props.tagsList.map(item => item.value)
+    }
+    this.setState({selectedValues})
+  }
+
+  _renderTags () {
```
naxhh: How do you feel regarding moving the map logic to the render function and here only taking care of rendering the tag?

something like:

```
const _renderTag = (tag, index) => (
 <TagSlectable
  key
  isSelected={this.state.selectedValues.contains(tag.value)}
  ...
}
```

 and in render

```
const {all, tagList} = this.props

return ...
 ....
 tagList.map(this._renderTag)
```

naxhh (approved):

components/tag/selectableList/src/index.js
```diff
@@ -0,0 +1,109 @@
...
+
+export default class TagSelectableList extends Component {
+  state = {
+    selectedValues: []
+  }
+
+  constructor (...props) {
+    super(...props)
+    this.tagHandlers = this.props.tagsList.map((tag) => this.toggleValue.bind(this, tag.value))
+    this.allHandler = this.toggleAll.bind(this)
```
kikoruiz: You can avoid this if you use an arrow function in the class method `toggleAll` ;)

components/tag/selectableList/src/index.js
```diff
@@ -0,0 +1,109 @@
...
+  checkIcon: PropTypes.func,
+  /**
+   * Optional string. When is setted show a button `All` at first position
+   */
+  all: PropTypes.string,
+   /**
+   * By clicking on every tag, `onChange` is triggered and sends an
+   * object with the selected items in the array.
+   */
+  onChange: PropTypes.func.isRequire,
```
kikoruiz: Typo `isRequire`

components/tag/selectableList/src/index.js
```diff
@@ -0,0 +1,109 @@
...
+
+TagSelectableList.propTypes = {
+  /**
+   * Icon to check tags
+   */
+  checkIcon: PropTypes.func,
+  /**
+   * Optional string. When is setted show a button `All` at first position
+   */
+  all: PropTypes.string,
```
kikoruiz: I would look for another name more descriptive: `allLabel` or simply `label`.

components/tag/selectableList/src/index.js
davidmartin84: I need toggleAll to be bound to the current instance.
Did you mean this in the class declaration ?

```
toggleAll = () => {
    this.setState({
      selectedValues: this.isAllSelected()
        ? []
        : this.props.tagsList.map(item => item.value)
    })
  }
```

components/tag/selectableList/src/index.js
davidbarna: https://developer.mozilla.org/es/docs/Web/JavaScript/Referencia/Objetos_globales/Array/includes
Be aware of compatibility.

davidbarna (approved):

components/tag/selectableList/src/index.js
kikoruiz: Yes!

kikoruiz (approved):

PR #139, opened 2017-08-25
fix(topbar/user): switch from set styles to set css classnames in order to avoid ios errors

kikoruiz:

### DIFF
--- a/components/topbar/user/src/index.js
+++ b/components/topbar/user/src/index.js
@@ -12,8 +12,6 @@ const DEFAULT_NAV_WRAP_STYLE = {
   height: 'inherit',
   width: 'inherit'
 }
-const HTML_DISABLED_SCROLL_STYLE = 'overflow: hidden; position: relative;'
-const BODY_DISABLED_SCROLL_STYLE = 'height: 100%; overflow: hidden; position: fixed; width: 100%;'
 
 /**
  * Topbar containing a dropdown with user data (login, logout, secured links...).
@@ -53,16 +51,16 @@ class TopbarUser extends Component {
    * Lock body element scroll.
    */
   _lockBodyScroll () {
-    window.document.documentElement.style = HTML_DISABLED_SCROLL_STYLE
-    window.document.body.style = BODY_DISABLED_SCROLL_STYLE
+    window.document.documentElement.classList.add('html-has-scroll-disabled')
+    window.document.body.classList.add('body-has-scroll-disabled')
   }
 
   /**
    * Unlock body element scroll.
    */
   _unlockBodyScroll () {
-    window.document.documentElement.style = ''
-    window.document.body.style = ''
+    window.document.documentElement.classList.remove('html-has-scroll-disabled')
+    window.document.body.classList.remove('body-has-scroll-disabled')
   }
 
   /**
--- a/components/topbar/user/src/index.js
+++ b/components/topbar/user/src/index.js
@@ -326,10 +326,7 @@ TopbarUser.defaultProps = {
   callToActionComponent: DefaultCallToAction,
   linkFactory: ({ href, className, children, title }) =>
     <a href={href} className={className} title={title}>{children}</a>,
-  elementsToKeepScrollOnToggleMenu: [
-    '.mt-Container',
-    '.mt-SharedFooter'
-  ]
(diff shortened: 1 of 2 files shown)


components/topbar/user/src/index.scss
```diff
@@ -21,6 +21,18 @@
   width: $notifications-size;
 }
 
+.html-has-scroll-disabled {
```
kikoruiz, line 24: I was trying to avoid this solution because it creates global classnames, and that's not a good idea, but regarding iOS Safari behaviour, it doesn't accept to write on read-only DOM elements `body` and `html`.

naxhh (approved):

components/topbar/user/src/index.js
```diff
@@ -52,17 +51,25 @@ class TopbarUser extends Component {
...
    */
-  _lockBodyScroll () {
-    window.document.documentElement.style = HTML_DISABLED_SCROLL_STYLE
-    window.document.body.style = BODY_DISABLED_SCROLL_STYLE
+  _lockBodyScroll = () => {
+    this._verticalScrollPosition = window.scrollY
+    window.document.documentElement.classList.add('html-has-scroll-disabled')
+    window.document.body.classList.add('body-has-scroll-disabled')
+    this.props.elementsToKeepScrollOnToggleMenu.forEach(selector => {
+      document.querySelector(selector).style.transform = `translate3d(0, -${this._verticalScrollPosition}px, 0)`
```
kikoruiz: Don't know if it's better to save document elements in memory before.

components/topbar/user/src/index.js
```diff
@@ -52,17 +51,25 @@ class TopbarUser extends Component {
...
-  _unlockBodyScroll () {
-    window.document.documentElement.style = ''
-    window.document.body.style = ''
+  _unlockBodyScroll = () => {
+    this.props.elementsToKeepScrollOnToggleMenu.forEach(selector => {
+      document.querySelector(selector).style.transform = ''
+    })
+    window.document.documentElement.classList.remove('html-has-scroll-disabled')
+    window.document.body.classList.remove('body-has-scroll-disabled')
+    window.scrollTo(0, this._verticalScrollPosition)
```
kikoruiz: Restore window scroll position.

components/topbar/user/src/index.js
```diff
@@ -306,14 +313,23 @@ TopbarUser.propTypes = {
...
 }
 
 TopbarUser.defaultProps = {
   toggleIcon: Menu,
   callToActionComponent: DefaultCallToAction,
   linkFactory: ({ href, className, children, title }) =>
-    <a href={href} className={className} title={title}>{children}</a>
+    <a href={href} className={className} title={title}>{children}</a>,
+  elementsToKeepScrollOnToggleMenu: [
+    '.mt-Container',
```
kikoruiz: Fuck, removing it! 😅

midudev: xD

midudev (approved):
Looks good to me :+1:

components/topbar/user/src/index.js
davidbarna: AFAIK it would not improve performance.
But, caching translate3d string may do.

components/topbar/user/src/index.js
```diff
@@ -52,17 +51,25 @@ class TopbarUser extends Component {
   /**
    * Lock body element scroll.
    */
-  _lockBodyScroll () {
-    window.document.documentElement.style = HTML_DISABLED_SCROLL_STYLE
-    window.document.body.style = BODY_DISABLED_SCROLL_STYLE
+  _lockBodyScroll = () => {
+    this._verticalScrollPosition = window.scrollY
+    window.document.documentElement.classList.add('html-has-scroll-disabled')
```
davidbarna: Why is 'html-has-scroll-disabled' not a constant as HTML_DISABLED_SCROLL_STYLE was?
If a string is repeated, I usually use constant for that.

davidbarna (approved):

naxhh (approved):

midudev (approved):

components/topbar/user/src/index.js
kikoruiz: Sure 👍

PR #142, opened 2017-08-30
fix(cover/basic): fix height and cursor issues for IE.

betobe:


davidbarna (approved):

components/cover/basic/src/index.scss
```diff
@@ -10,16 +10,16 @@
     }
   }
 
-  max-height: $mah-cover-basic;
+  height: $mah-cover-basic;
```
davidbarna, line 13: Are you sure of this? the behavior changes and max-height works in IE.
No need to answerd. Just want to call your attention on it

betobe: yep

davidmartin84 (approved):

ronaldo-mh (approved):

PR #145, opened 2017-09-04
feat(form/rangeDatepicker): new component Range datepicker

davidmartin84:


components/form/rangeDatepicker/README.md
```diff
@@ -0,0 +1,36 @@
...
+```javascript
+import FormRangeDatepicker from '@schibstedspain/sui-form-range-datepicker'
+```
+
+```html
+function handleClick (range) {
+  console.log('handleClick', range)
+}
+
+function handleChangeStart (date) {
```
davidbarna, line 26: It's a range, so I would put a single callbakc onChange(startDate, endDate) that would be triggered when user decides to SET the date (in this case, the click on the button)

components/form/rangeDatepicker/src/index.js
```diff
@@ -0,0 +1,128 @@
+import React, {Component, PropTypes} from 'react'
+
+import DatePicker from 'react-datepicker'
+import moment from 'moment'
```
davidbarna, line 4: If you're using moment, you should install it in the package.json.

components/form/rangeDatepicker/src/index.js
```diff
@@ -0,0 +1,128 @@
...
+import ButtonBasic from '@schibstedspain/sui-button-basic'
+import Chevronbottom from '@schibstedspain/sui-svgiconset/lib/Chevronbottom'
+
+import 'react-datepicker/dist/react-datepicker.css'
```
davidbarna: I would not include the css of the component.
We should create a scss for your component, and in it import https://github.com/Hacker0x01/react-datepicker/blob/master/src/stylesheets/datepicker.scss
IMO

components/form/rangeDatepicker/src/index.js
```diff
@@ -0,0 +1,128 @@
...
+
+const locale = 'es-es'
+
+class FormRangeDatepicker extends Component {
+  constructor (props) {
+    super(props)
+
+    this.state = {
+      startDate: !(this.props.startDate) ? null : moment(this.props.startDate),
```
davidbarna: No need of parentheses

components/form/rangeDatepicker/src/index.js
```diff
@@ -0,0 +1,128 @@
...
+   * Click event that will send the date range
+   */
+  handleClickButton: PropTypes.func,
+  /**
+   * Maximum date that the calendar allows to select
+   */
+  maxDate: PropTypes.object,
+  /**
+   * Init date of the selected range
+   */
```
davidbarna: IMO date should be a string to ease implementation of component. But as React is JS, maybe I'm wrong privileging raw props...
You should document the expected date format in the comments of props types.
If you wanna expect a Date object, at least use PropTypes.instanceOf(Date)

components/form/rangeDatepicker/src/index.scss
```diff
@@ -0,0 +1,50 @@
...
+  /* Overwrite react-datepicker styles */
+  .react-datepicker__day {
+
+    &--selected,
+    &--in-selecting-range,
+    &--in-range {
+      background-color: $c-primary;
+
+      &:hover {
+        background-color: darken($c-primary, 5%);
```
davidbarna: I don't know if darkens as allowed. You should check with @davecarter

davidbarna (changes requested):

components/form/rangeDatepicker/src/index.js
davidmartin84: good! i like it

components/form/rangeDatepicker/src/index.js
davidmartin84: i'm agree. I can overwrite variables of the react-datepicker component

components/form/rangeDatepicker/package.json
```diff
@@ -0,0 +1,22 @@
...
+    "build": "rm -Rf ./lib && mkdir -p ./lib && npm run build:js && npm run build:styles",
+    "build:js": "../../../node_modules/.bin/babel --presets sui ./src --out-dir ./lib",
+    "build:styles": "../../../node_modules/.bin/cpx './src/**/*.scss' ./lib"
+  },
+  "dependencies": {
+    "@schibstedspain/sui-component-dependencies": "latest",
+    "@schibstedspain/sui-svgiconset": "1",
+    "@schibstedspain/sui-button-basic": "1.1.0",
+    "react-datepicker": "0.52.0",
+    "moment": "2"
```
carlosvillu, line 16: That is a BIG dependencie. Are you sure to use this ?!

carlosvillu (changes requested):

components/form/rangeDatepicker/package.json
davidbarna: You're right.
But, the used package react-datepicker expects a moment instance as props. And it's already has moment as a dependency...
https://github.com/Hacker0x01/react-datepicker/blob/master/package.json#L96
To avoid moment, we would need to get rid of react-datepicker and find another one or develop the logic from scratch.

carlosvillu: Maybe we have two solutions:

1.- https://github.com/date-fns/date-fns/issues/275#issuecomment-264934189
2.- https://github.com/jmblog/how-to-optimize-momentjs-with-webpack

carlosvillu: We will try to delay the load of this component as much as in possible.

carlosvillu (approved):

davidbarna (approved):

PR #146, opened 2017-09-05
feat(list/image): first version of sui-list-image component.

ronaldo-mh:
Finally, I moved the list-image component from NC Components to SUI-components.
So, the changes correspond to the ones committed in the old closed PR:
https://github.schibsted.io/scmspain/frontend-nc--uilib-components/pull/6


carlosvillu (approved):

davidbarna (approved):

components/list/image/src/index.scss
```diff
@@ -0,0 +1,63 @@
...
+  @include media-breakpoint-only(xl) {
+    margin: -($m-v-list-image-item-half-gutter);
+  }
+  @include media-breakpoint-only(l) {
+    margin: (-($m-v-list-image-item-half-gutter)) (-($m-h-list-image-item-breakpoint-l-gutter));
+  }
+  @include media-breakpoint-down(m) {
+    justify-content: space-between;
+  }
+  display: flex;
```
jcger: you could use the list mixin https://github.com/SUI-Components/theme-basic/blob/master/src/utils/_list.scss

components/list/image/src/index.scss
```diff
@@ -0,0 +1,63 @@
...
+    display: flex;
+    height: 100%;
+    left: 0;
+    margin: 0;
+    opacity: $op-list-image-last-item;
+    position: absolute;
+    top: 0;
+    width: 100%;
+
+    span {
```
jcger: we try to avoid direct styling on elements

components/list/image/src/index.scss
```diff
@@ -0,0 +1,63 @@
+@import '~@schibstedspain/theme-basic/lib/index';
+@import '~@schibstedspain/sui-image-lazy-load/lib/index';
+
+$m-h-list-image-item-half-gutter: ceil($m-h-list-image-item-full-gutter / 2) !default;
+$m-v-list-image-item-half-gutter: ceil($m-v-list-image-item-full-gutter / 2) !default;
+$m-h-list-image-item-breakpoint-l-gutter: $m-h-list-image-item-full-gutter - 1px !default;
```
jcger: Isn't there a way to avoid that -1px magic number?

components/list/image/src/index.js
```diff
@@ -0,0 +1,97 @@
...
+  </li>
+)
+
+/**
+ * Creates an special list item, which has a regular image and an additional cover box over it (lack 65% opacity
+ * background) with a customisable message and onClick event callback.
+ * The HTML structure of this list-item is slightly differnet that the regular Image Item.
+ * @param image {Object} src, alt and onClick callback configuration of the image to display.
+ * @param index {Number} Order index of the li element to create.
+ * @param showMoreItemsBox {Obhect} Text and onClick callback configuration for the list-item to be created.
```
jcger: typo Object

components/list/image/src/index.js
```diff
@@ -0,0 +1,97 @@
...
+  ).isRequired,
+  /**
+   * Max number of items to show in ListImage.
+   */
+  maxItems: PropTypes.number,
+  /**
+   * When showMoreItemsBox is set and the length of "images" prop is greater than "maxItems" prop value, the last item of the list
+   * is displayed with an extra cover where the text label and onClick action can be customised.
+   */
+  showMoreItemsBox: PropTypes.shape({
```
jcger: this sounds like boolean, maybe moreItemsBox is enough

components/list/image/src/index.js
```diff
@@ -0,0 +1,97 @@
...
+ * @param index {Number} Order index of the li element to create.
+ */
+const renderImageItem = (image, index) => (
+  image && <li className='sui-ListImage-item' key={index} onClick={ensureFunction(image.onClick)}>
+    {renderSuiLazyloadImage(image)}
+  </li>
+)
+
+/**
+ * Creates an special list item, which has a regular image and an additional cover box over it (lack 65% opacity
```
jcger: Consider removing that 65% as it's not really set in this component

components/list/image/src/index.scss
ronaldo-mh: Ok. Included sui-list mixin and removed dsiplay and padding style properties in next commit.

components/list/image/src/index.scss
ronaldo-mh: Added a CSS class name to the span element.

components/list/image/src/index.js
ronaldo-mh: 👍

components/list/image/src/index.js
ronaldo-mh: Right, moreItemsBox property name fits better. 👍

components/list/image/src/index.js
ronaldo-mh: Done! 👍

components/list/image/src/index.scss
ronaldo-mh: For the sui-component, we will keep the horizontal gutter of 16px between cells.

And in our custom theme, we will overwrite this variable to accomplish our specific cell gutter only for screens between 1024 and 1240px (UX requirements).
So, for our project the fixed cell gutter for this screen sizes must be 30px (defined in UX designs as well), that is why we had the half gutter variable of 15px (```$m-h-list-image-item-full-gutter - 1px```), **but we will keep this only in our specific project's theme.**

jcger (approved):

PR #156, opened 2017-09-21
fix(modal/basic): use classname instead direct styles in body

ruben-martin-lozano:
@miduga @davecarter please review. Now instead of setting direct styles to body, use a css class

### DIFF
--- a/components/modal/basic/src/index.scss
+++ b/components/modal/basic/src/index.scss
@@ -1,5 +1,9 @@
 @import '~@schibstedspain/theme-basic/lib/index';
 
+body.is-modal-open {
+  overflow-y: hidden;
+}
+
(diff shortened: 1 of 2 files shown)


danivm (approved):

components/modal/basic/src/index.scss
```diff
@@ -1,5 +1,9 @@
 @import '~@schibstedspain/theme-basic/lib/index';
 
+body.is-modal-open {
```
ruben-martin-lozano, line 3: I decided to preppend directly to "body" tag, because we only want this class acts on body, and because it seems pretty generic (is-modal-open), to avoid colisions I think it's better.

aleanez (approved):

midudev (approved):

PR #159, opened 2017-09-27
SUI Atom Tag

jcger:
TODOs

- [x] README
- [x] Use theme-basic vars
- [x] Raw SVG or wrapped
- [x] QA issues

Definition here: https://paper.dropbox.com/doc/SUI-Tags-bOSL4L5TFYZqPaK0wnCHb


components/tag/basic/src/index.js
```diff
@@ -0,0 +1,114 @@
...
+          Icon &&
+            <span>
+              <Icon svgClass='sui-TagBasic-icon' />
+            </span>
+        }
+        <span className='sui-TagBasic-label' title={label}>
+          {label}
+        </span>
+        {
+          !onClick && onClose &&
```
naxhh: We don't do this anywhere... but it's a good idea to add a warning that's removed on prod build, like react does, when the user pass an `onClick` and `onClose` ? since the second one will be ignored?

components/tag/basic/src/index.scss
```diff
@@ -0,0 +1,51 @@
+@import '~@schibstedspain/theme-basic/lib/index';
+
+.sui-TagBasic {
```
jcger: I'm using hardcoded values until theme-basic is migrated

kikoruiz: Ok, I understand this point, but if we leave it like this other verticals won't be able to reuse this component, just with an ugly practice: overwritting variables.

I'd rather create some default variables for those values within this file or in a specific place inside the theme-basic, ok?

jcger: Yes, of course, I'm waiting for the final name and value of most of them in order to update this file

components/tag/basic/src/index.js
jcger: I guess you are talking about a global solution and not about implementing it in this component, right?

components/tag/basic/package.json
```diff
@@ -0,0 +1,17 @@
...
+  "version": "1.0.0",
+  "description": "",
+  "main": "lib/index.js",
+  "scripts": {
+    "build": "rm -Rf ./lib && mkdir -p ./lib && npm run build:js && npm run build:styles",
+    "build:js": "../../../node_modules/.bin/babel --presets sui ./src --out-dir ./lib",
+    "build:styles": "../../../node_modules/.bin/cpx './src/**/*.scss' ./lib"
+  },
+  "dependencies": {
+    "@schibstedspain/sui-component-dependencies": "latest"
```
davidbarna: Update your deps, so the sui-studio generate will add correct deps

components/tag/basic/src/index.js
naxhh: Yep. My question was more for @davidbarna & co.

This component could benefit on it, but the implementation should come from outside.

I think of it as QOL since a developer could find himself in a WTF moment when the onclose is ignored etc.

jcger: Yes, would be nice. For now it's explained in the docs and in the component's own definition. A clickable tag, cannot be closed.

demo/atom/tag/playground
```diff
@@ -0,0 +1,31 @@
...
+
+const CloseIcon = () => (
+  <svg viewBox='0 0 24 24'>
+    <path id='a' d='M13.42 12l4.79-4.8a1 1 0 0 0-1.41-1.41L12 10.58 7.21 5.79A1 1 0 0 0 5.8 7.2l4.79 4.8-4.79 4.79a1 1 0 1 0 1.41 1.41L12 13.41l4.8 4.79a1 1 0 0 0 1.41-1.41L13.42 12z'/>
+  </svg>
+)
+
+return (
+  <div style={{float: 'left'}}>
+    <h2>Standard Tags</h2>
```
beatrizip, line 15: Style embebed in div, better: <div clas="x">...</div>, in file.css/scss  .x {float: left; }

components/atom/tag/src/index.scss
```diff
@@ -0,0 +1,73 @@
...
+  }
+
+  &-icon {
+    margin-right: $p-s;
+  }
+
+  &-secondary-icon {
+    cursor: pointer;
+    margin: 0 (-$m-s) 0 $m-m;
+  }
```
beatrizip: a negative margin is necessary?  (-$m-s)

components/atom/tag/src/index.scss
```diff
@@ -0,0 +1,73 @@
...
+
+    // if both icons are set
+    .sui-AtomTag-hasIcon.sui-AtomTag-hasClose & {
+      max-width: $w-label-base - ($p-l + $m-s) * 2;
+    }
+  }
+
+  &-icon {
+    margin-right: $p-s;
+  }
```
beatrizip: p-... is a padding variable

components/atom/tag/src/index.scss
jcger: After talking about it "in person" with @beatrizip,  it's ok!

demo/atom/tag/playground
jcger: We can't import files from the playground

components/atom/tag/src/Actionable/Container.js
```diff
@@ -0,0 +1,20 @@
+import React from 'react'
+import PropTypes from 'prop-types'
+
+/**
+ * Component treated as an anchor when href is defined
+ */
+const ActionableTagContainer = ({Link, href, target, children, ...props}) => {
+  return href
+    ? <Link href={href} target={target} {...props}>{children}</Link>
```
carlesnunez, line 9: Whats the goal of pass a Link component instead of render a link here I don't understand this part really well :/

carlesnunez (dismissed):

components/atom/tag/src/Actionable/Container.js
carlesnunez: Maybe we could use always <a> tag here and dismiss pass a component as a prop that looks a bit strange :s

jcger: You might want to use another kind of link-alike component, for example, a react-router link

components/atom/tag/src/Actionable/index.js
```diff
@@ -0,0 +1,71 @@
+import React, {Component} from 'react'
+import PropTypes from 'prop-types'
+import classnames from 'classnames'
+import ActionableTagContainer from './Container'
+
+class ActionableTag extends Component {
```
JoanClaret: Since there is no "state", should we use a function instead of a class?

components/atom/tag/src/Actionable/index.js
```diff
@@ -0,0 +1,71 @@
...
+  static RIGHT_ICON_PLACEMENT = 'right'
+
+  get _classNames () {
+    const {className} = this.props
+    return classnames(
+      'sui-AtomTag-actionable',
+      className
+    )
+  }
```
JoanClaret: what do you think about using `cx` here instead of this function?
https://github.com/SUI-Components/sui-components/blob/master/components/card/product/src/index.js#L81
would look a bit more clear to me.

components/atom/tag/src/index.scss
```diff
@@ -0,0 +1,73 @@
...
+    .sui-AtomTag-hasIcon.sui-AtomTag-hasClose & {
+      max-width: $w-label-base - ($p-l + $m-s) * 2;
+    }
+  }
+
+  &-icon {
+    margin-right: $m-s;
+  }
+
+  &-secondary-icon {
```
JoanClaret: is this the "close" icon?

JoanClaret (approved):

components/atom/tag/src/index.scss
jcger: It's the close button for standard tags, but actionable tags can have an icon on the right side to

components/atom/tag/src/Actionable/index.js
jcger: It's using the same lib, I just named it classnames because the lib is called classnames instead of cx (I don't know why we import it as cx, what has cx to do with classnames?, do someone know? It looks pretty random)

components/atom/tag/README.md
```diff
@@ -0,0 +1,21 @@
...
+## Installation
+
+```sh
+npm install @schibstedspain/sui-atom-tag --save
+```
+
+## Usage
+
+```js
+import AtomTag from '@schibstedspain/sui-atom-tag'
```
davidbarna, line 16: Please add a real example.

components/atom/tag/package.json
```diff
@@ -0,0 +1,19 @@
...
+  "description": "",
+  "main": "lib/index.js",
+  "scripts": {
+    "build": "rm -Rf ./lib && mkdir -p ./lib && npm run build:js && npm run build:styles",
+    "build:js": "../../../node_modules/.bin/babel --presets sui ./src --out-dir ./lib",
+    "build:styles": "../../../node_modules/.bin/cpx './src/**/*.scss' ./lib"
+  },
+  "dependencies": {
+    "@s-ui/component-dependencies": "latest",
+    "@schibstedspain/sui-theme": "8"
```
davidbarna: Now component dependencies provides sui-theme. No need to include it:
https://github.com/SUI-Components/sui/blob/master/packages/sui-component-dependencies/package.json#L13

components/atom/tag/src/Actionable/index.js
```diff
@@ -0,0 +1,67 @@
...
+    </ActionableTagContainer>
+  )
+}
+
+ActionableTag.propTypes = {
+  className: PropTypes.string,
+  label: PropTypes.string.isRequired,
+  Icon: PropTypes.func,
+  href: PropTypes.string,
+  iconPlacement: PropTypes.string,
```
davidbarna: oneOf your constants

components/atom/tag/src/Actionable/index.js
```diff
@@ -0,0 +1,67 @@
...
+          </span>
+      }
+    </ActionableTagContainer>
+  )
+}
+
+ActionableTag.propTypes = {
+  className: PropTypes.string,
+  label: PropTypes.string.isRequired,
+  Icon: PropTypes.func,
```
davidbarna: Isn't it a node ?

components/atom/tag/src/Standard.js
```diff
@@ -0,0 +1,52 @@
...
+        }
+      </span>
+    )
+  }
+}
+
+StandardTag.propTypes = {
+  onClose: PropTypes.func,
+  CloseIcon: PropTypes.func,
+  Icon: PropTypes.func,
```
davidbarna: node ?

components/atom/tag/src/Standard.js
```diff
@@ -0,0 +1,52 @@
...
+            </span>
+        }
+      </span>
+    )
+  }
+}
+
+StandardTag.propTypes = {
+  onClose: PropTypes.func,
+  CloseIcon: PropTypes.func,
```
davidbarna: Node?

components/atom/tag/src/index.scss
```diff
@@ -0,0 +1,73 @@
...
+    margin: 0 (-$m-s) 0 $m-m;
+  }
+
+  &-secondary-closeable:hover {
+    background-color: $c-system;
+    border-radius: 50%;
+    fill: color-variation($c-gray, 3);
+  }
+
+  & [class$='icon'] {
```
davidbarna: FYI this kind of selectors a not efficient and should be avoided when possible.

davidbarna (approved):

components/atom/tag/src/Actionable/index.js
jcger: From our last frontend meeting I understood that we'll continue using the svg components that we were using before, those are no nodes, they are funcs. For example [tag chip](https://github.com/SUI-Components/sui-components/blob/master/components/tag/chip/src/index.js#L62)

davidbarna: You might be right. Don't remember and can't find the meeting note either.

jcger: @kikoruiz @carlosvillu do you remember what we agree with?

PR #161, opened 2017-09-29
Feature/form element input text

ronaldo-mh:


jcger:
Is this going to be in any production env soon? (same for the textarea)

I mean, UX is working on this, maybe it would be possible to coordinate it with them

components/form/textInput/src/index.js
```diff
@@ -0,0 +1,59 @@
...
+const FormTextInput = ({className, errorMessage, name, onChange, placeholder, value}) => (
+  <div className={cx('sui-FormTextInput', className)}>
+    <input
+      className='sui-FormTextInput-value'
+      name={name}
+      type='text'
+      placeholder={placeholder}
+      value={value}
+      onChange={onChange} />
+    {renderErrorMessage(errorMessage)}
```
JoanClaret: What do you think about checking if there is an error before to execute the function?
`error && renderErrorMessage(errorMessage)`

components/form/textInput/src/index.scss
```diff
@@ -0,0 +1,39 @@
...
+
+    &:focus {
+      outline-width: 0;
+    }
+  }
+}
+
+.sui-FormTextInput {
+  &-value {
+    @include sui-text-input;
```
JoanClaret: i don't get the point of having a mixin here. 
does we pass `$default-border` or `$default-border` somewhere?

components/form/textInput/src/index.js
ronaldo-mh: Yes, it is better on that way. I will change it in next commit! Thanks! 👍

components/form/textInput/src/index.scss
ronaldo-mh: Yes, there is not point as the default borders won't be used anymore.
I will change it in next commit.

components/form/textInput/src/index.js
JoanClaret: Great!

components/form/textInput/src/index.js
```diff
@@ -0,0 +1,59 @@
...
+
+const renderErrorMessage = error => (
+  error && (
+    <div className='sui-FormTextInput-errorMessage'>
+      <span className='sui-FormTextInput-errorMessageLabel'>{error}</span>
+    </div>
+  )
+)
+
+const FormTextInput = ({className, errorMessage, name, onChange, placeholder, value}) => (
```
jcger: You are adding this class in your other PR too, please remove wherever it has to be removed

ronaldo-mh: Yes, this is the right branch for these files, I have just removed from the other PR, thanks!

ronaldo-mh:
@julian-gernun @JoanClaret 
- Removed the unused mixin.
- Removed wrong files from PR (corresponding to textarea's PR).
- Added comment in README.md to clarify the "provisional" status of this component.

Please review! Thanks!

JoanClaret (approved):

jcger (approved):

PR #162, opened 2017-09-29
Feature/form element textarea

ronaldo-mh:


kikoruiz:
Hello, I just have a doubt: Are these components the same defined in https://paper.dropbox.com/doc/SUI-Form-Elements-WIP-dLZacE1Ye7Qxrp7n9Y0Hy?

ronaldo-mh:
@kikoruiz No, initially they are not the same defined in the dropbox paper for SUI form elements.
The design delivered by our team's UX designer has some differences, but in the moment he created it (have into account that this was designed thinking in the New Constructions' detail page contact form), he validated it with the SUI UX team, even though in that moment there was no specifications for the form elements.

Having said that, I see that the changes required to be fully compliant with the specification (it is only styles differences) are neither extensive nor complicated.

¿Do you think it would be better if I adapt this styles now (taking into account that the definition of form elements is still in progress, and that I don't have the sketch/zeplin, so I don't know the margins/paddings/colours/font-sizes/borders/etc.)?
Or could I deliver this and then adapt it when the form elements specification goes from WIP to DONE?

components/form/textarea/src/index.js
```diff
@@ -0,0 +1,66 @@
...
+   */
+  cols: PropTypes.number,
+  /**
+   * Title label to show above the text area element.
+   */
+  label: PropTypes.string,
+  /**
+   * Specifies a name for the text area.
+   */
+  name: PropTypes.string.isRequired,
```
jcger: Why is this required?

jcger (approved):

components/form/textInput/README.md
```diff
@@ -0,0 +1,7 @@
+### FormTextInput
```
jcger: Please check the FormTextInput files here, they might not be part of this PR

kikoruiz:
@frmedina Well, iterating over this component sounds the best option, but let other SUI-developers know that this component is not finished yet. README can be a good place for it.

kikoruiz (approved):

components/form/textarea/src/index.js
ronaldo-mh: Same as pointed in the checkbox's PR, I will remove this 👍

components/form/textInput/README.md
ronaldo-mh: Yes, my mistake, pushed to wrong branch.
I'll fix it now, thanks! 👍

ronaldo-mh:
@kikoruiz Great idea, I will write something in the README to clarify it!

PR #163, opened 2017-09-29
feat(form/checkbox): first commit of checkbox form element component.

ronaldo-mh:


components/form/checkbox/README.md
```diff
@@ -0,0 +1,9 @@
+### FormCheckbox
+
+Simple checkbox form element.
+
+You can pass a custom callback to be executed when the checkbox value is changed, through ```onChange``` property.
```
davidbarna: FYI you can use simple quotes instead of triple when it's not multiline

davidbarna (approved):

components/form/checkbox/src/index.js
```diff
@@ -0,0 +1,74 @@
...
+   */
+  svgIcons: PropTypes.shape({
+    checked: PropTypes.node.isRequired,
+    unchecked: PropTypes.node.isRequired
+  })
+}
+
+FormCheckbox.defaultProps = {
+  svgIcons: {
+    checked: <svg className='sui-FormCheckbox-icon' xmlns='http://www.w3.org/2000/svg' viewBox='0 0 16 16'>
```
kikoruiz: What about using this repo for these icons? https://github.com/SUI-Components/svg-iconset. I understand they are not official and have to be redefined by UX/Design, but it's better to work with graphic icons that exports React components. Working directly with SVG icons (its code) can be painful.

JoanClaret:
Hi!
just to know, is this component aligned with UX sui-forms specifications?

components/form/checkbox/README.md
ronaldo-mh: Thanks! 👍
I will update this in next commit.

components/form/checkbox/src/index.js
ronaldo-mh: Yes, no problem, I can move these 2 icons to the svg-iconset repo.
Besides, do you think it's still useful to allow the customisation of the checkbox icons through a component's propert? Or not anymore? (just to know what you think about removing this `svgIcons` prop after moving the icons to the other repo).

ronaldo-mh:
@JoanClaret These implementation is not fully compliant with the dropbox paper of the SUI form elements.
This is based in the design created for the checkboxes of the contact form of New Construction promotion's detail page, and by the time our UX design it, there was no specification or document for form elements in the Dropbox (but our UX validated the design he made with the SUI UX team).

Besides this, the specification of the document is very similar to the one developed in this component, but as the form elements definition is still WIP, I don't have yet specified colours/margins/paddings/font-sizes, etc., and the icons for checked/unchecked status are not final yet.

JoanClaret:
I get your point, and i understand we should deliver components as soon as possible, but as far as i know, we also should develop components following the sui-components ux specificactions, which must be approved and validated by the cross UX team. since this form components still in WiP, maybe we should ask to the UX team to get the final specifications in order to avoid later refactors.

jcger:
@JoanClaret Looks like obra nueva has the ok to do components even if they do not fit the SUI definitions as they are in a hurry working for a MVP

components/form/checkbox/src/index.js
```diff
@@ -0,0 +1,74 @@
...
+   */
+  errorMessage: PropTypes.string,
+  /**
+   * Text to display as a description message on the right side of the chechbox input.
+   */
+  label: PropTypes.node,
+  /**
+   * Specifies the name of the checkbox component.
+   */
+  name: PropTypes.string.isRequired,
```
jcger: Is there any special reason this is required?

jcger (approved):

naxhh:
Wouldn't be better to make those components in other repo instead? SUI components is about having SUI compliant components...

I know @carlosvillu point is to have all components here but I don't see why we are doing this...
Maybe a 5 min talk with everyone to decide where this kind of components should be?

davidbarna (approved):

JoanClaret (approved):

kikoruiz:
Absolutely agree with @naxhh, this repo is not the place to leave non-SUI-compliant-components if the main reason is "a specific site has the need right now". I bet for a talk in order to clarify this process.

components/form/checkbox/src/index.js
ronaldo-mh: Not really, I wanted to have an unique id/name to be used in parent's form, but using the form's state is enough. I will remove this.

ronaldo-mh:
From the last meeting we had, we decided to put our component in the sui-components repository.
In fact, if they are not fully compliant now with the UX design, it does not mean they will not be ever compliant.
Once the UX definition is completed, we can iterate this component to build the sui compliant, and I think there won't be any problems of backward compatibility and adaptation to our designs when the final version is delivered.

And as pointed in the other PR, I will add a comment in the README file to clarify this.

PR #165, opened 2017-09-29
chore(META): add CONTRIBUTING info with first convention rule

davidbarna:

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -1,0 +1,54 @@
+# Contributing to SUI Components
+
+The following is a set of guidelines for contributing with developping new components.
+
+## Getting started
+
+### Environment
+
+To develop new packages, you only need to install Node with NPM. Fin below the default versions this repo was built with:
+* Node: `v6+`
+* NPM: `v3`
+
+### Main commands
+
+This repo is managed with [sui-studio](https://www.npmjs.com/package/@s-ui/studio) tool.
+
+Some base commands (based on sui-studio) you should acknowledge :
+* `npm run generate <category> <name>` to create a new component.
+* `npm run co` to commit your changes with a wizard.
+* `npm run release` publish all your changes for al packages (once your PR is merge).
+
+## Naming Conventions
+
+### Event Handlers
+
+Many presentational components give feedback from user interaction calling provided callbacks.
+
+#### `on{Event}`
+
+```html
+<MyButton
+  onClick={onClickHandler}
+  onHover={onHoverHandler}
+/>
+<MyTextArea
+  onScroll={onScrollHandler}
+  onChange={onChangeHandler}
+/>
+```
+
+#### `on{SubContext}{Event}`
+For compound component, we may need to provide several callbacks.
+
+```html
+<MyDialogModal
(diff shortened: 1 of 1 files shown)


CONTRIBUTING.md
```diff
@@ -0,0 +1,54 @@
+# Contributing to SUI Components
+
+The following is a set of guidelines for contributing with developping new components.
+
+## Getting started
+
+### Environment
+
+To develop new packages, you only need to install Node with NPM. Fin below the default versions this repo was built with:
```
midudev: `Fin` -> `Find`

CONTRIBUTING.md
```diff
@@ -0,0 +1,54 @@
+# Contributing to SUI Components
+
+The following is a set of guidelines for contributing with developping new components.
```
midudev: `developping` -> `developing`

CONTRIBUTING.md
```diff
@@ -0,0 +1,54 @@
...
+* Node: `v6+`
+* NPM: `v3`
+
+### Main commands
+
+This repo is managed with [sui-studio](https://www.npmjs.com/package/@s-ui/studio) tool.
+
+Some base commands (based on sui-studio) you should acknowledge :
+* `npm run generate <category> <name>` to create a new component.
+* `npm run co` to commit your changes with a wizard.
```
midudev, line 19: With a wizard?

image

CONTRIBUTING.md
```diff
@@ -0,0 +1,54 @@
...
+* `npm run release` publish all your changes for al packages (once your PR is merge).
```
midudev: `al` -> `all`

CONTRIBUTING.md
```diff
@@ -0,0 +1,54 @@
...
+  onChange={onChangeHandler}
+/>
+```
+
+#### `on{SubContext}{Event}`
+For compound component, we may need to provide several callbacks.
+
+```html
+<MyDialogModal
+  onClick={onClickHandler}
```
midudev, line 46: If we do that, then we're assuming that the `onClick` for which handler exactly? For the submit? I don't know if maybe, if it's a component with several, we should name all of them.

CONTRIBUTING.md
davidbarna: https://en.wikipedia.org/wiki/Wizard_(software) LOL will change.
I also mispelled the examples

midudev: hahahah don't worry! I know `wizard` is fine is just that I thought in Gandalf while reading it. xD

midudev (approved):

CONTRIBUTING.md
davidbarna: I don't have a strong opinion on that. To me, it's "no subcontext for default behavior". 
In the next example, onChange is "any change" and onFilterChange would be on a specific filter...
Let's see what the rest of the guys think.

midudev: But `any change` is even harder to determine, isn't? Because the onFilterChange is not going to overwrite the `onChange`.

kikoruiz (approved):

PR #169, opened 2017-10-02
feat(atom/button): AtomButton

davidbarna:


components/atom/button/README.md
```diff
@@ -0,0 +1,39 @@
...
+
+**Find full description and more examples in the [demo page](https://sui-components.now.sh/workbench/atom/button).**
+
+
+
+
+
+
+<!--
+TODO: Fix Design Problems
```
davidbarna: Waiting for UX to fix some stuff

components/atom/button/src/index.scss
```diff
@@ -0,0 +1,139 @@
+@import '~@schibstedspain/sui-theme/lib/index';
```
davidbarna: Not installed. Looking for the best solution

components/atom/button/README.md
```diff
@@ -0,0 +1,39 @@
+# AtomButton
+
+> Atom Element: SUI button
+
+![](./assets/screenshot.png)
+
+## Installation
+
+```sh
+npm install @schibstedspain/sui-atom-button --save
```
jcger, line 10: I thought that our sui-components would be under @s-ui

components/atom/button/src/index.scss
```diff
@@ -0,0 +1,139 @@
...
+
+$c-button: $c-primary !default;
+$c-button-light: color-variation($c-primary, 4) !default;
+$c-button-dark: $c-primary-dark !default;
+$c-button-light-contrast: color-variation($c-primary, 3) !default;
+$c-button-white: #ffffff !default;
```
jcger: I'm not sure if we should call it `...-white`, maybe in a vertical's theme they use another color as the foreground color.  Also, you will move this vars to sui-theme, right?

components/atom/button/src/index.scss
```diff
@@ -0,0 +1,139 @@
...
+$bw-button: 1px !default;
+
+// Set all color properties for text and icons
+@mixin button-foreground-color ($color) {
+  color: $color;
+
+  & svg path {
```
jcger: I would love to do this too but someone once told to me that we do not apply styles directly on tags and also that we do no exceptions so we don't have to memorize that exceptions. Maybe it could be reopened to discussion...

components/atom/button/README.md
davidbarna: No. Only tools. Not components.

components/atom/button/src/index.scss
davidbarna: Probably will be migrated. But It's one of the discussions I wanna have with the team

components/atom/button/src/index.scss
davidbarna: We reference out tag with classes. But in this case, It has to be because it's not part of our markup.

components/atom/button/src/index.js
```diff
@@ -0,0 +1,92 @@
+import React, {PropTypes} from 'react'
+import cx from 'classnames'
+
+const CLASS = 'sui-AtomButton'
+const TYPES = ['primary', 'accent', 'secondary', 'tertiary']
+const MODIFIERS = ['disabled', 'small', 'large', 'fullWidth', 'focused']
+
+const appendClass = (suffix) => `${CLASS}--${suffix}`
+const includes = (array, item) => array.indexOf(item) !== -1
+const getTypes = (props) => {
```
jcger: If I'm not wrong, doing it this way, last type wins. Although the developer should avoid having more than one type set as prop, consider that it might be dangerous as maybe some type in between applies styling that is not overwritten afterwards.

jcger (approved):

components/atom/button/src/index.js
```diff
@@ -0,0 +1,98 @@
+import React, {PropTypes} from 'react'
+import cx from 'classnames'
+
+const CLASS = 'sui-AtomButton'
+const TYPES = ['primary', 'accent', 'secondary', 'tertiary']
+const MODIFIERS = ['disabled', 'small', 'large', 'fullWidth', 'focused']
+
+const appendClass = (suffix) => `${CLASS}--${suffix}`
+const includes = (array, item) => array.indexOf(item) !== -1
+const cleanProps = (props, Component) => {
```
davidbarna: This is risky cause the API is fully coupled to button API.
I think it's good so we don't have to declare everything like onClick, accessiblityLabel, value, etc.
On the other hand, if we want out button to be an other element (image, svg, link) someday we can't ensure retrocompatibility. But In this case we can trust it's not gonna be the case.
What are your thoughts about it?

components/atom/button/src/index.js
```diff
@@ -0,0 +1,98 @@
...
+  return (<button {...newProps} className={classNames} disabled={disabled}>
+    {leftIcon && <span className={`${CLASS}LeftIcon`}>{leftIcon}</span>}
+    {children}
+    {rightIcon && <span className={`${CLASS}RightIcon`}>{rightIcon}</span>}
+  </button>)
+}
+
+AtomButton.displayName = 'AtomButton'
+AtomButton.propTypes = {
+  /**
```
davidbarna, line 50: I chose an API based on aggregable props. button-basic has a different approach "type", "color", ...
Important to give your opinion as we can define new conventions to make our set of components more consistent.

components/atom/button/src/index.js
```diff
@@ -0,0 +1,98 @@
...
+  const classNames = cx(
+    CLASS,
+    getTypes(props).map(appendClass),
+    getModifiers(props).map(appendClass),
+    className
+  )
+  const newProps = cleanProps(props, AtomButton)
+  return (<button {...newProps} className={classNames} disabled={disabled}>
+    {leftIcon && <span className={`${CLASS}LeftIcon`}>{leftIcon}</span>}
+    {children}
```
davidbarna: The text, contrary to button-basic, is the children. This gives more flexibility fits with the regular behaviour of native buttons.

components/atom/button/src/index.js
jcger: I guess that removing propTypes for prod env wouldn't be an option anymore

components/atom/button/src/index.scss
```diff
@@ -0,0 +1,139 @@
...
+    margin-right: $m-m;
+  }
+
+  &RightIcon {
+    margin-left: $m-m;
+  }
+}
+
+// Variations in contrast context
+.sui-Global--contrast {
```
JoanClaret: is `contrast` a web app property or a component property?

components/atom/button/README.md
```diff
@@ -0,0 +1,62 @@
...
+
+### Flexible props
+
+All props available from regular buttons an be used.
+
+```js
+import Button from '@schibstedspain/sui-atom-button'
+
+return (<div>
+  <AtomButton onClick={() => alert('Primary with onClick')}>
```
JoanClaret, line 35: does make sense to add the context in the components? IMHO i would use only the component name.

kikoruiz:
I would like to add the same as @JoanClaret, don't see the point to add the context within the name of the components. What do you think about?

components/atom/button/src/index.js
```diff
@@ -0,0 +1,98 @@
+import React, {PropTypes} from 'react'
+import cx from 'classnames'
+
+const CLASS = 'sui-AtomButton'
+const TYPES = ['primary', 'accent', 'secondary', 'tertiary']
+const MODIFIERS = ['disabled', 'small', 'large', 'fullWidth', 'focused']
```
kikoruiz: I don't think all they are the same. My point is: `disabled` and `focused` are states, and the other ones are modifiers (related to the button size).

components/atom/button/src/index.js
```diff
@@ -0,0 +1,98 @@
...
+const AtomButton = (props) => {
+  const {disabled, leftIcon, rightIcon, children, className} = props
+  const classNames = cx(
+    CLASS,
+    getTypes(props).map(appendClass),
+    getModifiers(props).map(appendClass),
+    className
+  )
+  const newProps = cleanProps(props, AtomButton)
+  return (<button {...newProps} className={classNames} disabled={disabled}>
```
kikoruiz: What about all those links (`a` elements) that we want to give the button look-and-feel? There are many cases

components/atom/button/README.md
davidbarna: What do you mean? I don't understand.

JoanClaret: it's about the component name. would be a better option to name it as `Button` instead of `AtomButton`?

components/atom/button/src/index.js
davidbarna: touché !!! :_(

components/atom/button/src/index.scss
davidbarna: That's a thing I was about to discuss.
It makes no sense to add contrast as a prop y every component cause its a context.
So, this is a proposal. But I won't merge it. We'll discuss it in the meeting about sui-theme.

components/atom/button/README.md
davidbarna: It's a mistake of the readme...
I'll change it to Button here. But in the index.js, the displayName will remain beting AtomButton. Agree?

components/atom/button/src/index.js
davidbarna: It's a matter of naming. TYPES are exclusive. So if none if none is set, primary is added by default.
MODIFIERS are just agrregable classes. Event if disable is a state for native button, in this context it's simply a CSS modifier to change the style. The component has no state.

components/atom/button/src/index.js
davidbarna: Good point. button-basic outputs a button as well.
Maybe designers should define a link or do you think we should add an option here ?

components/atom/button/src/index.scss
JoanClaret: Ok! Fine for me.

components/atom/button/src/index.js
davidbarna: Changed it so we'll only have one type no matter how many are defined.

davidbarna:
Closing review to build a new one as it's very old and the code has evolved a lot

PR #180, opened 2017-10-11
feat(atom/statusSpot): new component:  atom status spot

davidmartin84:


davidbarna (approved):

components/atom/statusSpot/src/index.js
```diff
@@ -0,0 +1,26 @@
...
+  }
+}
+
+AtomStatusSpot.displayName = 'AtomStatusSpot'
+
+AtomStatusSpot.propTypes = {
+  /**
+   * The string content is the status spot
+   */
+  status: PropTypes.string
```
naxhh: Shouldn't this be `oneOf` instead of free string?

davidmartin84: You're right!!! "oneOf" is better.

naxhh (approved):

kikoruiz (approved):

PR #182, opened 2017-10-16
feat(card/collapsible): first version of card collapsible component.

ronaldo-mh:
Pending to move scss vars to sui-theme.


components/card/collapsible/src/index.js
```diff
@@ -0,0 +1,146 @@
...
+    this.setState({
+      collapsed: !this.state.collapsed
+    })
+  }
+
+  /**
+   * Render a single text info item, that will be displayed in the card's header positioned depending on the expanded/collapsed state.
+   * @param label {string} Text to display.
+   * @param link {string} If set, item will be displayed as a text link.
+   * @param highlighted {boolean} Highlight the item (bold style).
```
jcger: you could add the return tag

components/card/collapsible/src/index.js
```diff
@@ -0,0 +1,146 @@
...
+   */
+  _renderInfoItem ({label, link, highlighted}, index) {
+    const Element = link ? 'a' : 'span'
```
jcger: Might be nothing but as `Element` is a native var too it might be dangerous

components/card/collapsible/src/index.scss
```diff
@@ -0,0 +1,82 @@
...
+
+          &:last-child {
+            margin: $m-m 0 0;
+          }
+        }
+      }
+    }
+  }
+
+  .sui-CollapsibleBasic {
```
jcger, line 79: I'm not sure if it's the right approach. You'r changing a component style directly. Maybe it's not the case, but if you end up having two collapsibleBasic in your site, both will have `border: 0` and you might not want to. I think you should be able to set a new className to the collapsibleBasic and use that selector to set your border from here.

components/card/collapsible/src/index.js
ronaldo-mh: Ok! I will improve the jsdoc comments in next commit.

components/card/collapsible/src/index.js
ronaldo-mh: Didn't count on that, I will rename the var to `InfoItemElement`.

components/card/collapsible/src/index.scss
ronaldo-mh: Ok, we will add the collapsible component's `border` parametrization feature in another PR, and fix this case removing this style.

betobe (approved):

PR #184, opened 2017-10-19
Feature/create sui links basic

midudev:
Move sui-links to sui-link-basic component in the Studio.

image


davecarter (approved):

components/links/basic/src/index.js
```diff
@@ -0,0 +1,82 @@
...
+          title={title}
+          to={url}
+        >
+          {this._renderContent()}
+        </Link>
+      )
+    }
+
+    return (
+      <a
```
rackso, line 53: Missing hreflang, onfocus and onblur, in case someone needs it.

components/links/basic/src/index.js
```diff
@@ -0,0 +1,82 @@
...
+LinkBasic.propTypes = {
+  className: PropTypes.string,
+  disabled: PropTypes.bool,
+  handleClick: PropTypes.func,
+  icon: PropTypes.element,
+  literal: PropTypes.string,
+  rel: PropTypes.string,
+  target: PropTypes.string,
+  title: PropTypes.string,
+  url: PropTypes.string.isRequired,
```
rackso, line 82: Missing default value.

rackso (approved):

components/links/basic/src/index.js
midudev: I only have just copied the old one. :) I rather prefer people to add them in case they need. Anyway none of the three mentioned make sense for the purposes of this component.

components/links/basic/src/index.js
midudev: :+1:

PR #185, opened 2017-10-19
feat(collapsible/accordion): new accordion collapsible component

davidmartin84:


components/collapsible/accordion/README.md
```diff
@@ -0,0 +1,23 @@
...
+$ npm install @schibstedspain/sui-collapsible-accordion --save
+```
+
+## Usage
+
+### Basic usage
+```js
+import CollapsibleAccordion from '@schibstedspain/sui-collapsible-accordion'
+
+return (<CollapsibleAccordion />)
```
davidbarna: If is requires props, put a real example with props.

components/collapsible/accordion/src/index.js
```diff
@@ -0,0 +1,54 @@
...
+class CollapsibleAccordion extends Component {
+  constructor (props) {
+    super(props)
+    this.state = {items: props.items}
+    this._handleClick = this._handleClick.bind(this)
+    this._collapseItems = this._collapseItems.bind(this)
+  }
+
+  _handleClick (id) {
+    return (collapsed) =>
```
davidbarna: You can write it in one line and remove the parens

```
return collapsed => this._collapseItems(collapsed, id)
````

Comment (davidbarna) [inline components/collapsible/accordion/src/index.js]: You can write it in one line and remove the parens

```
return collapsed => this._collapseItems(collapsed, id)
````

File: components/collapsible/accordion/src/index.js [hunk 3]
```diff
@@ -0,0 +1,54 @@
...
+      this._collapseItems(collapsed, id)
+  }
+
+  _collapseItems (collapsed, id) {
```
davidbarna: For performance, You could remember was the las id and then you wont have to iterate on every render.
And, please implement the shouldComponentUpdate hook.

components/collapsible/accordion/src/index.js
```diff
@@ -0,0 +1,54 @@
...
+      item.collapsed = (index === id) ? collapsed : true
+      return item
+    })})
+  }
+
+  _renderCollapsible () {
+    const { items } = this.state
+    return items.map((item, index) => {
+      return (
+        <CollapsibleBasic key={index} handleClick={this._handleClick(index)} collapsed={item.collapsed} label={<div style={{padding: '8px'}}>{item.label}</div>} >
```
davidbarna: Usually we put one props per line for readability.
Please never style with `style` prop. Use css classes. Do you really need this div ?
Id "div" the most semantic tag? Perhaps <p> of <section> would be more SEO friendly.

Also, to avoid breaking API in a future, you can do:

```js
return items.map((item, index) => {
      return (
        <CollapsibleBasic {...item} key={index} handleClick={this._handleClick(index)} >
```
Then if in the future CollapsibleBasic has more props, you won't have to change this component to make it work.

components/collapsible/accordion/src/index.js
```diff
@@ -0,0 +1,54 @@
...
+          {item.content}
+        </CollapsibleBasic>
+      )
+    })
+  }
+
+  render () {
+    return (
+      <div>
```
davidbarna: Do you really need to wrap it in a div ??? Seems a bad decision.

components/collapsible/accordion/src/index.js
```diff
@@ -0,0 +1,54 @@
...
+  }
+}
+
+CollapsibleAccordion.displayName = 'CollapsibleAccordion'
+
+CollapsibleAccordion.propTypes = {
+  /**
+   * Items array
+   */
+  items: PropTypes.array.isRequired
```
davidbarna: You can use ArrayOf to be more accurate.

davidbarna (changes requested):

davidmartin84:
@davidbarna  done all changes requested

naxhh (approved):
Nice work!

JoanClaret (approved):

davidbarna (approved):

PR #190, opened 2017-10-20
feat(image/placeholder)

jcger:
Shows a placeholder (svg) until an image is loaded. You can also set a fallback image to be loaded in case it fails loading

Is not defined by UX, needed by ThumbnailBasic but could be useful for other components.


demo/image/placeholder/playground
```diff
@@ -0,0 +1,34 @@
...
+        image={{
+          src: 'https://satyr.io/1000?delay=3g',
+          alt: 'Some alt'
+        }}
+        placeholder={placeholder}
+      />
+    </div>
+    <h2>Fallback example</h2>
+    <div style={{width: '20%'}}>
+      <ImagePlaceholder
```
beatrizip, line 27: Better style in CSS class

demo/image/placeholder/playground
```diff
@@ -0,0 +1,34 @@
...
+  <svg viewBox="0 0 1 1">
+      <rect width="100%" height="100%" />
+  </svg>
+)
+
+return (
+  <div>
+    <h2>Image Placeholder</h2>
+    <div style={{width: '20%'}}>
+      <ImagePlaceholder
```
beatrizip: Better style in CSS class

components/image/placeholder/README.md
```diff
@@ -0,0 +1,23 @@
+# ImagePlaceholder
+
+> Description
+
+<!-- ![](./assets/preview.png) -->
+
```
beatrizip: commentary??

demo/image/placeholder/playground
jcger: It's just the demo so its easier to see what happens.

Edit: we are not able to load files from the demos.

demo/image/placeholder/playground
davidbarna, line 27: I this a new SUI component from UX ?
I'm not a big fan of passing the image as properties. 
What about if my image has a title or other image properties that are not mention in ths shape you define ?
I'd rather imagine ImagePlaceholder as a enhence IMG tag and would do: 

```
<ImagePlaceholder
  src= 'http://404'
  alt= 'Some alt'
  placeholder={placeholder}
  fallbackImage={<img src="https://satyr.io/500" alt="" />}
/>
```
What do you think ?
I find weird that placeholder is a component but fallback an object.

jcger: No, its for our logos component (without extra UX definition). I like your proposal

Edit: implemented

davidbarna (approved):

PR #196, opened 2017-10-24
Feature/add name prop to form elements

ronaldo-mh:
- Add name property to form elements (in order to have a name reference for each component in parent's form).
- Moved form element scss vars from sui-theme's settings-compat-v7 file to each component.


davidbarna (approved):
Nice !!!

betobe (approved):

PR #197, opened 2017-10-25
feat(dropdown/basic): make component render also react router links by sending a config flag

kikoruiz:

### DIFF
--- a/components/dropdown/basic/src/index.js
+++ b/components/dropdown/basic/src/index.js
@@ -168,7 +168,7 @@ DropdownBasic.propTypes = {
       /**
        * Flag to know if uses React Router Link.
        */
-      useReactRouterLink: PropTypes.string
+      useReactRouterLink: PropTypes.bool
     }))
   })).isRequired,
   /**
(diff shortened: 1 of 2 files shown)


components/dropdown/basic/src/index.js
```diff
@@ -173,8 +183,10 @@ DropdownBasic.propTypes = {
 
 DropdownBasic.defaultProps = {
   expandOnMouseOver: false,
-  linkFactory: ({ href, className, children, title }) =>
-    <a href={href} className={className} title={title}>{children}</a>
+  linkFactory: ({ href, children, useReactRouterLink = false, ...attrs }) =>
+    useReactRouterLink
```
naxhh, line 187: Just me or this if is backwards?

naxhh (changes requested):
Personal opinion:

I don't think this is the right approach.
Components should not know about react-router at all.
They should be able to be used outside the react-router scope. and this will add a hard-dependency on 
it. That's why we send linkFactory as a prop

For the specific MT use case. You can already accomplish the task sending the `linkFactory` in `News` page as the `linkFactory` prop in the `TopBar`

jcger:
Agree with @naxhh , the factory function you pass through could know if it should render an anchor tag or an react router link, right?

kikoruiz:
Not needed, closing it.

kikoruiz:
We have a solution that takes into account if the `linkFactory`has to use a React Link or not. But anyway, we should rethink of this solution: https://github.com/SUI-Components/sui-components/blob/master/components/link/basic/src/index.js#L3

PR #202, opened 2017-10-26
Feature/ngcp 52 share buttons

carlesnunez:


components/button/share/src/index.js
```diff
@@ -0,0 +1,83 @@
...
+        src: `http://twitter.com/share?url=${parsedShareUrl}&tti=1&pagination=1&RowGrid=2&text=${shareText}`,
+        icon: Socialtwitter
+      }
+    case 'whatsapp':
+      return {
+        src: `whatsapp://send?text=${shareText} - ${parsedShareUrl}`,
+        icon: Commentsquare
+      }
+    case 'email':
+      return {
```
davidbarna, line 30: Shouldn't we do a mailto with the shareText and the url as body in this case ??

components/button/share/src/index.js
```diff
@@ -0,0 +1,83 @@
...
+      }
+    default:
+      return {}
+  }
+}
+
+const ButtonShare = ({type, icon, onClick, shareText, url}) => {
+  const shareDefinition = getShareDefinition(url, shareText, type)
+  const Icon = icon || shareDefinition.icon
+  return <button
```
davidbarna, line 42: Why not a link with href, target=_blank and let the native behaviour being executed ?
To me in makes no sense to use a button for a window.open.

components/button/share/src/index.js
```diff
@@ -0,0 +1,83 @@
...
+  /**
+   * A class definition, not instance, of our react icon component.
+   */
+  icon: PropTypes.func,
+  /**
+   * Callback function to be called onClick button. Default type onClick is handled if no onClick is provided
+   */
+  onClick: PropTypes.func,
+  /**
+   * The text to put as message on the share. Only twitter and whatsapp have this feature.
```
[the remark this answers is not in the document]
carlesnunez: I don't understand about which test should be applied to emails

components/button/share/src/index.js
carlesnunez: We didn't create this as a <a> because we thought in the email case. The email case in 'obra nueva' open a modal instead of redirect to an URL. This modal should be opened passing an onClick callback to the sharingButton.

It doesn't looks very pretty to opening a modal with an onClick handler in a <a> tag. What do you think, we keep the button, we change it to <a> or we make a mix between <a> and <button> tag wether if we have a onClick callback or not.

components/button/share/src/index.js
carlesnunez: Yup totally agree! Our email in 'obra nueva' opens a modal an we didn't thought about use mailTo as the default value here. Thank you!

components/button/share/src/index.js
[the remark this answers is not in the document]
carlesnunez: Oh, are you meaning TEXT? In this case, yeah, we can combine the mailto change with add a body on the mail option too

kikoruiz (approved):

components/button/share/src/index.js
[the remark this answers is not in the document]
davidbarna: That's what I meant. Currently the email only opens a blank window.

components/button/share/src/index.js
davidbarna: It's ok. If you don't need the email implementation in your project you can skip it.

davidbarna (approved):

components/button/share/src/index.js
carlesnunez: We can do the mailto implementation is a good idea to have in that component.

components/button/share/src/index.js
```diff
@@ -0,0 +1,83 @@
+import React, {PropTypes} from 'react'
+import cx from 'classnames'
+import {Socialtwitter, Socialfacebook, Commentsquare, Envelopeclosed} from '@schibstedspain/sui-svgiconset'
+
+const getOnClickHandle = src => () => window.open(src)
+
+const parseHttpSchemeOfUrl = (urlToEval) => {
+  const isUrlValid = urlToEval.match(/http:\/\/|https:\/\//)
+  return !isUrlValid ? `http://${urlToEval}` : urlToEval
```
naxhh: I think string functions should be much quicker here.

```
return urlToEval.startsWith('http://') || urlToEval.startsWith('https://')
 ? urlToEval
 : `http://${urlToEval}`
```

I think the polyfill is in the default list but if not indexOf can make the same thing

components/button/share/src/index.js
carlesnunez: Done!

components/button/share/src/index.js
carlesnunez: Done using indexOf!!

ronaldo-mh (approved):

PR #204, opened 2017-10-27
fix(dropdown/basic): fix spacing between links

devtoni:
fix spacing between links when dropdown basic is in mobile size

### DIFF
--- a/components/dropdown/basic/src/index.scss
+++ b/components/dropdown/basic/src/index.scss
@@ -100,7 +100,7 @@
       background-color: $bgc-dropdown-menu;
       border-top: $bdt-dropdown-menu;
       display: block;
-      padding: $gutter-s;
+      padding: $p-m;
     }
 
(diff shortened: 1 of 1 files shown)


JoanClaret (approved):

components/dropdown/basic/src/index.scss
```diff
@@ -100,36 +100,32 @@
       background-color: $bgc-dropdown-menu;
       border-top: $bdt-dropdown-menu;
       display: block;
-      padding: $gutter;
-      padding-bottom: $gutter-s;
+      padding: $gutter-s;
```
kikoruiz: Makes sense that it keeps setting a gutter value?

devtoni: As we've commented before, I've modified this shape variables to standard naming

kikoruiz (approved):

PR #206, opened 2017-10-30
feat(tag/chip): adapt new prop rel to render if it is necessary

devtoni:


JoanClaret (approved):

JoanClaret:
Parece que falla travis

PR #212, opened 2017-10-31
fix(cookie/banner): avoid to parse document global var by the server

kikoruiz:

### DIFF
--- a/components/cookie/banner/src/index.js
+++ b/components/cookie/banner/src/index.js
@@ -87,11 +87,9 @@ class CookieBanner extends Component {
 
   render () {
     const { showCookieBanner } = this.state
-    const hasAcceptedCookies = this._getHasAcceptedCookie()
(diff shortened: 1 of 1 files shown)


components/cookie/banner/src/index.js
```diff
@@ -87,11 +87,9 @@ class CookieBanner extends Component {
 
   render () {
     const { showCookieBanner } = this.state
-    const hasAcceptedCookies = this._getHasAcceptedCookie()
-    if (hasAcceptedCookies || !showCookieBanner) {
+    if (!showCookieBanner || this._getHasAcceptedCookie()) {
```
kikoruiz, line 90: When server renders, it won't execute `_getHasAcceptedCookie()` method that tries to read `document` var.

siscu (approved):

midudev (approved):

PR #214, opened 2017-11-02
fix/image-placeholder

jcger:
As svg resize is broken in IE11 this component does not allow <svg> node as placeholder/fallback anymore. For now on, use placeholder/fallback.src in order to set your svg's (see playground file)

### DIFF
--- a/components/image/placeholder/README.md
+++ b/components/image/placeholder/README.md
@@ -19,12 +19,14 @@ return (
     src='https://satyr.io/1000'
     alt='1000x1000'
     placeholder={
-      <svg viewBox='0 0 1 1'>
-        <rect width='100%' height='100%' />
-      </svg>
+      src: '...',
+      alt: '...'
+      ...imageTagAttributes
     }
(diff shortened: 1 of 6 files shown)


PablitoGS (approved):

davidbarna:
What do you been by svg resize broken ??

jcger:
I mean that svg with width/height do not resize/scale

davidbarna:
But SVGs meant to be used as icons should not have width and height. Only viewbox. It's a convention for SVG scaling.
The problem is that I've have a different approach with buttons and we must have the same approach. Can you send me a SVG example so I've checkout it there is a better solution? You can merge this if you need it. Don't want to be a stopper.

davidbarna (approved):

jcger:
SVG wasn't used as icon in this PR, it was used as a placeholder and the placeholder has to resize in order to fill the same space as the image that is being loaded but as IE11 does not resize the svg, it was broken.

PR #216, opened 2017-11-03
feat(map/basic): here maps component first version

carlesnunez:
**What does this PR do?**
This PR implements a reusable component to put HERE maps on any react web page

The demo is using HERE DEMO app_id and app_code.

**How should this be manually tested (steps)?**
Download de branch and play with it on sui-studio. You have full implementation info on the readme

screen shot 2017-11-03 at 13 08 06


components/map/basic/src/index.js
```diff
@@ -0,0 +1,227 @@
...
+      defaultLayers: null
+    }
+  }
+
+  _setCurrentLoadedStyles () {
+    this.currentStylesLoaded = Array.from(document.querySelectorAll('link')).map(style => style.href)
+  }
+
+  _setCurrentLoadedLibraries () {
+    this.currentLibrariesLoaded = Array.from(document.querySelectorAll('script')).map((library) => library.src)
```
davidbarna, line 29: Always optimize your selectors and be as specific as possible.
Look at the performance of `querySelectorAll` when you could use `getElementsByTagName` instead

https://jsperf.com/queryselectorall-vs-getelementsbytagname

components/map/basic/src/index.js
```diff
@@ -0,0 +1,227 @@
...
+        lineWidth: PropTypes.number,
+        lineCap: PropTypes.string,
+        lineJoin: PropTypes.string
+      })
+    }))
+  })
+}
+
+MapBasic.defaultProps = {
+  libraries: ['http://js.api.here.com/v3/3.0/mapsjs-core.js', 'http://js.api.here.com/v3/3.0/mapsjs-service.js', 'https://js.cit.api.here.com/v3/3.0/mapsjs-ui.js', 'https://js.cit.api.here.com/v3/3.0/mapsjs-mapevents.js'],
```
davidbarna, line 221: I have several comments about libraries management:

1. Some of the libs are with the http protocol, which would provoke problems in https pages
A good practice is to define urls os CDNs protocol relative like this:
```js
'//js.api.here.com/v3/3.0/mapsjs-core.js', 
'//js.api.here.com/v3/3.0/mapsjs-service.js'
'//js.cit.api.here.com/v3/3.0/mapsjs-ui.js'
'//js.cit.api.here.com/v3/3.0/mapsjs-mapevents.js'
```

2. Your are checking the libs loaded on every single instance obtaining them from the DOM.
You easily take track of if it's the first instance or not and do theses actions ONLY THE FIRST TIME. It's a matter or performance.

3. To me it doesn't make sense to allow the injection of libraries to load. Theses libraries are your dependencies. You choose what version your comptaible with. It should be hidden.

4. You are add scripts in the page, multiple requests. Have you considered using the npm version? https://www.npmjs.com/package/here-js-api
You could delay the loading of the library with require.ensure I guess.

5. If you, for some reason, maintain the loading logic, please move it to another file, to keep responsibilities separated

components/map/basic/src/index.js
```diff
@@ -0,0 +1,227 @@
...
+      this._initStylesInjection(styles)
+    }
+  }
+
+  _initLibraryInjection (libraries, appId, appCode, useCIT, useHTTPS) {
+    const libraryLength = libraries.length
+    const libraryUrl = libraries.shift()
+
+    if (!this.currentLibrariesLoaded.includes(libraryUrl) && libraryLength) {
+      const script = document.createElement('script')
```
davidbarna, line 61: You are appending it to body. It's a blocking script. Your component is blocking the full page delaying the render to the load of these libraries. The script should be async.

components/map/basic/src/index.js
```diff
@@ -0,0 +1,227 @@
...
+      this.map.instance.addObject(
+        new window.H.map.Circle({lat: element.lat, lng: element.lng},
+        element.radius,
+        { style: element.style || this.props.circleShape.styleGeneric })
+      )
+    })
+  }
+
+  _buildMapInstance () {
+    const DOMElementToBeFilled = document.getElementById('sui-here-map')
```
davidbarna, line 113: If you use the ''sui-here-map'' literal more than once, you should store it as a contant.
Plus, you are using an id to reference your element. This means that you component is "singleton". Only one instance of it would work on a page, not two.

Use unique ids or react ref prop for that

components/map/basic/src/index.js
```diff
@@ -0,0 +1,227 @@
...
+  isInteractive: PropTypes.bool,
+  /**
+   * Use CIT ONLY SHOULD BE ENABLED if we are going to use a customer integration testing app id and app code
+   * More info about CIT here https://developer.here.com/documentation/venue-maps/topics/request-cit-environment.html
+   */
+  useCIT: PropTypes.bool,
+  /**
+   * Enables or disables HTTPS communication with the platform.
+   */
+  useHTTPS: PropTypes.bool,
```
davidbarna, line 174: I really this it's reponsiblity of the component to determine if it's in an https context or not.
Don't you think?

demo/map/basic/playground
```diff
@@ -0,0 +1,25 @@
+return (<div style={{ width: '50%', height: '400px' }}>
```
davidbarna, line 1: Can you add another example with the simplest config ?

davidbarna (changes requested):

PR #219, opened 2017-11-07
Feature/callback collapsible accordion

davidmartin84:


components/collapsible/accordion/src/index.js
```diff
@@ -16,6 +16,9 @@ class CollapsibleAccordion extends Component {
 
   _collapseItems (collapsed, id) {
     this.setState({ openIndex: id === this.state.openIndex ? null : id })
+    if (this.props.onItemChange) {
```
midudev: Better if you put this.props.onItemChange value as an empty function in the defaultProps and avoid using a conditional.

davidmartin84: @miduga good! Done!

naxhh (approved):

PR #221, opened 2017-11-09
SUI Components: atom/button

davidbarna:
SPEC: https://paper.dropbox.com/doc/SUI-Buttons-AvdtvjMAqbsFkTYSz3egT
DEMO: https://sui-components-atom-button.now.sh/workbench/atom/button/demo

This is the implementation of the new button component.


components/atom/button/src/index.js
```diff
@@ -0,0 +1,106 @@
...
+ */
+const getModifiers = (props) => {
+  return Object.keys(props)
+    .filter(name => MODIFIERS.includes(name))
+}
+
+const AtomButton = (props) => {
+  const {disabled, leftIcon, rightIcon, children, className, type, size} = props
+
+  console.log(size, CLASSES[size])
```
jcger: console.log

components/atom/button/src/index.js
```diff
@@ -0,0 +1,106 @@
+import React from 'react'
+import PropTypes from 'prop-types'
+import cx from 'classnames'
+
+const CLASS = 'sui-AtomButton'
+const TYPES = ['primary', 'accent', 'secondary', 'tertiary']
+const SIZES = ['small', 'large']
```
jcger, line 7: what about exporting these props so they can be used to feed the component without magic strings?

davidbarna: Thinking about remaining as similar as HTML spec I thought it wasn't necessary.
We agreed not to force the use of constants for easier implementation. But maybe we still want to expose them for those who prefer to use them ?

jcger: I guess it's up to you

components/atom/button/src/index.scss
```diff
@@ -0,0 +1,180 @@
...
+  border: 1px solid;
+  border-radius: $bdrs-m;
+  box-sizing: border-box;
+  font: {
+    family: $ff-sans-serif;
+    size: $fz-button;
+    weight: $fw-semi-bold;
+  }
+  height: $sz-button;
+  min-width: $sz-button;
```
JoanClaret, line 76: Does we need a min-width? i can't find it in component's definition.

jcger (approved):

components/atom/button/src/index.scss
```diff
@@ -0,0 +1,180 @@
...
+
+  & + & {
+    margin-left: $m-m;
+  }
+
+  // Icons
+  &-leftIcon,
+  &-rightIcon {
+    display: inline-block;
+    transform: translateY(-2px);
```
JoanClaret, line 89: i don't get the point of having here a magic number.

components/atom/button/src/index.scss
```diff
@@ -0,0 +1,180 @@
...
+      border-color: $c-accent-dark;
+    }
+    background-color: $c-accent;
+    border-color: $c-accent;
+  }
+
+  // Modifiers
+  &--disabled {
+    cursor: not-allowed;
+    opacity: .3;
```
JoanClaret, line 149: should we define a global opacity variable? even if we does not have an opacity specification, i would find it useful.

components/atom/button/src/index.scss
davidbarna: It's in the zepplin linked in the paper

image

components/atom/button/src/index.scss
davidbarna: There's something with the icon that make them non perfectly center verically as exposed in the zepplin. It's to be pixel perfect

components/atom/button/src/index.scss
```diff
@@ -0,0 +1,180 @@
...
+  font: {
+    family: $ff-sans-serif;
+    size: $fz-button;
+    weight: $fw-semi-bold;
+  }
+  height: $sz-button;
+  min-width: $sz-button;
+  outline: 0;
+  padding: 0 $p-button;
+  transition: background-color .3s ease, border-color .3s ease;
```
JoanClaret: For a consistent animations, i think we could define a timing variable in sui-theme.

components/atom/button/src/index.scss
JoanClaret: great! thanks for point it out!

components/atom/button/src/index.scss
davidbarna: I proposed it to designers and they said they'll treat the animation thing later on.
Actually the transition was not defined by them. I just did it because I saw it in the IJ buttons. But maybe I should remove it as it was not clearly defined. What do you think?

components/atom/button/src/index.scss
davidbarna: I don't think so. This opacity value was specially defined for disable state of buttons.

components/atom/button/src/index.scss
```diff
@@ -0,0 +1,180 @@
...
+  & .sui-AtomButton-rightIcon {
+    margin-left: $margin;
+  }
+}
+
+.sui-AtomButton {
+  @include reset-button;
+  @include button-icon($icon-sz-button, $icon-m-button);
+  border: 1px solid;
+  border-radius: $bdrs-m;
```
beatrizip, line 68: I would put a border width variable

davidbarna: Borders are not a global variable and here it's used only once.
Why would you use a variable ? For future maintenance ?

components/atom/button/src/index.scss
JoanClaret: Looks fine to me, but i would define a global variable:

```
$trsdu-base: .3;
$trsdu-button: $trsdu-base !default;
```

So every vertical can assign their own speed.
IMHO, it's the same case as we should manage opacity.

davidbarna: Removing animation as not defined by design.

JoanClaret (approved):
great job mate

PR #223, opened 2017-11-13
feat(map/basic): Leaflet maps basic component first version

carlesnunez:
**What does this PR do?**
This PR's add's the SUI-MAP-BASIC component taked from fc-web-server and refactorized.

Some of the new features of the map include **new file structure** **some code changed and deleted some unused properties and functions** and the ability to **have more than one map instance at the same time on a single page**.

**The FC Map was decoupled from FC repo and refactorized. The old structure was:**

- leaflet-map.js
- search-map.js
- search-polygons.js

The new structure and code abstract the 'search' word from our code and split the map responsability in three files:

- map.js _// Main file who owns the managers and the map itself_
- markers-manager.js _// CRUD markers responsability_
- layers-manager.js _// CRUD Layers responsability_
- shapes/polygons.js _// Polygon shape responsability_

screen shot 2017-11-13 at 18 41 38
screen shot 2017-11-13 at 18 36 30
screen shot 2017-11-13 at 18 36 20


components/map/basic/leafletLocalInstall.js
```diff
@@ -0,0 +1,25 @@
+var fs = require('fs')
```
carlesnunez, line 1: File made to avoid problems on copy time of NON scss files and images.

components/map/basic/src/leaflet/marker-manager.js
```diff
@@ -0,0 +1,179 @@
...
+
+  createMarker (item) {
+    const events = [
+      { eventName: 'click', eventHandler: (e) => this.isPoiClicked(e) },
+      { eventName: 'mouseover', eventHandler: (e) => this.onMouseOver(e) },
+      { eventName: 'mouseout', eventHandler: (e) => this.onMouseOut(e) }
+    ]
+    const {latitude, longitude, isSelected, markerType, propertyInfo} = item
+    const marker = L.marker([latitude, longitude], {icon: this.getIconFor({item})})
+    marker.propertyInfo = propertyInfo
```
carlesnunez, line 34: This part is completely coupled to FC contract for POI's this should be changed on the future.

components/map/basic/package.json
```diff
@@ -0,0 +1,19 @@
...
+  "main": "lib/index.js",
+  "scripts": {
+    "build": "rm -Rf ./lib && mkdir -p ./lib && npm run build:js && npm run build:styles",
+    "build:js": "../../../node_modules/.bin/babel --presets sui ./src --out-dir ./lib",
+    "build:styles": "../../../node_modules/.bin/cpx './src/**/*.scss' ./lib",
+    "postinstall": "node ./leafletLocalInstall"
+  },
+  "dependencies": {
+    "@s-ui/component-dependencies": "1",
+    "leaflet": "^1.2.0"
```
midudev: Remove the ^ caret and use the fixed version 1.2.0.

components/map/basic/src/index.js
```diff
@@ -0,0 +1,248 @@
...
+    this.mapInstance = undefined
+    this.isHeatmapVisible = false
+    this.isSatelliteView = false
+  }
+
+  setMapEventDefinition () {
+    this.mapEventList = [
+      {
+        name: 'leaflet_map_click',
+        handleFunction: (evt) => this.props.onMapClick && this.props.onMapClick(evt.detail)
```
midudev: In order to keep simple the component, I would put the `onMap...` method as a defaultProp with an empty function.

```
const NO_OP = () => {}
Map.defaultProps = {
  onMapClick: NO_OP
}
```

So you simplify your code like this as the method will exist always.
```
handleFunction: (evt) => this.props.onMapClick(evt.detail)
```

I even would like to see that:
```
handleFunction: this.props.onMapClick
```

And extract the `detail` from the `evt` on the prop instead. But I assume that you're doing this to keep compatibility.

carlesnunez: Totally agree with the first point! not so much on the second one.

As you said I've modified as much as possible without break any contract on FC current map implementation. In the near future there are a bunch of things and contracts that should be simplified to dismiss big amounts of code to be executed.

components/map/basic/src/leaflet/layer-manager.js
```diff
@@ -0,0 +1,74 @@
...
+  removeLayersFromGroup (layersToDelete, groupName) {
+    !this.layers[groupName] && (this.layers.markers = this.getFullLayerGroup())
+    this.layers[groupName].eachLayer(layer => {
+      const found = layersToDelete.find(pointToDelete => layer.Id === pointToDelete.Id)
+      found && this.layers.markers.removeLayer(layer)
+    })
+  }
+
+  getBaseMapView (value) {
+    switch (value) {
```
midudev: It doesn't seem to have more options?

`return value === mapViewModes.SATELLITE ? 'aerial' : 'base'`

components/map/basic/src/leaflet/map.js
```diff
@@ -0,0 +1,268 @@
...
+
+  getCenter () {
+    return {
+      latitude: this._map.getCenter().lat,
+      longitude: this._map.getCenter().lng
+    }
+  }
+
+  getCenterWithZoom () {
+    let centerWithZoom = this.getCenter()
```
midudev:
```
return {
  ...this.getCenter(),
  zoom: this.getZoomLevel()
}
```

midudev:
Great work @carlesnunez 👏 👏 I'm in vacation but managed to take a quick look to this. Looks nice!! :)

components/map/basic/src/leaflet/layer-manager.js
carlesnunez: Not by the moment but is a really good one candidate to be scaled. In any case is more clear this approach.

carlesnunez:
PR Changes done @miduga

components/map/basic/src/leaflet/layer-manager.js
midudev: Smells to YAGNI. :P

components/map/basic/src/index.js
```diff
@@ -0,0 +1,257 @@
...
+  componentWillReceiveProps (nextProps) {
+    const {heatMapUrl, pois, showHeatmap, showSatelliteView} = nextProps
+    pois.length && this.mapInstance.displayPois(pois)
+    this.checkIfHeatMapShouldBeDisplayed(showHeatmap, heatMapUrl)
+    this.checkWhichViewShouldBeDisplayed(showSatelliteView)
+  }
+
+  componentDidMount () {
+    this.subscribeToMapEvents()
+    this.mapInstance = new LeafletMap(this.getMapConfig())
```
midudev, line 127: At this point, I think that you have the `this.mapDOMInstance` with the DOM element. It would be great to use this instead getting manually the element later in https://github.com/SUI-Components/sui-components/pull/223/files#diff-f4a8e7a6d35f7bb6b104560342f8ef5eR14

Also, you could pass down to the marker-manager.

midudev (approved):

components/map/basic/src/index.js
```diff
@@ -0,0 +1,257 @@
...
+      this.mapInstance.removeHeatMapLayer()
+    }
+  }
+
+  componentWillUnmount () {
+    this.unsubscribeFromMapEvents()
+  }
+
+  shouldComponentUpdate () {
+    // The component itself have no changes. All changes are managed through leaflet maps api.
```
davidbarna: typo "has"

components/map/basic/src/index.js
```diff
@@ -0,0 +1,257 @@
...
+    return false
+  }
+
+  componentWillReceiveProps (nextProps) {
+    const {heatMapUrl, pois, showHeatmap, showSatelliteView} = nextProps
```
davidbarna: You can directly destructure in the params parens

```js
  componentWillReceiveProps ({heatMapUrl, pois, showHeatmap, showSatelliteView}) {

```

components/map/basic/src/index.js
```diff
@@ -0,0 +1,257 @@
...
+    pois.length && this.mapInstance.displayPois(pois)
```
davidbarna: What if there were 2 pois and now there's none. You should execute displayPois to remove them right ?

components/map/basic/src/index.js
```diff
@@ -0,0 +1,257 @@
...
+
+  subscribeToMapEvents () {
+    this.mapEventList.forEach(mapEvent => this.mapDOMInstance.addEventListener(mapEvent.name, mapEvent.handleFunction))
+  }
+
+  unsubscribeFromMapEvents () {
+    this.mapEventList.forEach(mapEvent => this.mapDOMInstance.removeEventListener(mapEvent.name, mapEvent.handleFunction))
+  }
+
+  checkWhichViewShouldBeDisplayed (showSatelliteView) {
```
davidbarna, line 90: You're not checking, you're showing the correct view... The name is not good to me.

components/map/basic/src/index.js
```diff
@@ -0,0 +1,257 @@
...
+    if (showSatelliteView && !this.isSatelliteView) {
+      this.isSatelliteView = true
+      this.mapInstance.setView(mapViewModes.SATELLITE)
+    } else if (!showSatelliteView && this.isSatelliteView) {
+      this.isSatelliteView = false
+      this.mapInstance.setView(mapViewModes.NORMAL)
+    }
+  }
+
+  checkIfHeatMapShouldBeDisplayed (showHeatmap, url) {
```
davidbarna, line 100: You're not checking, you're display the heapmap... maybe `toggleHeatMap`of  `showHeatMap`

components/map/basic/src/leaflet/layer-manager.js
```diff
@@ -0,0 +1,67 @@
...
+  createMapLayers ({mapViewModes, maxZoom, minZoom, appId, appCode}) {
+    let tileLayers = []
+    mapViewModes.forEach((value, index) => {
+      const baseMapView = this.getBaseMapView(value)
+      const tileLayer = L.tileLayer('https://{s}.{base}.maps.api.here.com/maptile/2.1/maptile/{mapID}/' + value + '/{z}/{x}/{y}/256/png8?app_id={app_id}&app_code={app_code}', {
+        app_code: appCode,
+        app_id: appId,
+        attribution: 'Map &copy; 1987-2017 <a href="http://developer.here.com">HERE</a>',
+        base: baseMapView,
+        mapID: 'newest',
```
davidbarna: hardcoded ID ?

components/map/basic/src/leaflet/marker-manager.js
davidbarna: Not a fan of having this in sui-components... I thing it should not be allowed.
The follwoing lines are event more coupled with IsFavorite, isPromotion, etc.

components/map/basic/src/leaflet/shapes/Polygons.js
```diff
@@ -0,0 +1,88 @@
...
+    layer.on({
+      dblclick: SearchMapPolygons.zoomIn,
+      mouseout: SearchMapPolygons.resetHighlight,
+      mouseover: SearchMapPolygons.highlightFeature
+    })
+  }
+
+  printPolygonOnMap ({ map, polygon }) {
+    const params = {
+      className: 'scm-map__area',
```
davidbarna, line 54: What does scm stand for ?

components/map/basic/leafletLocalInstall.js
carlesnunez: The thing is that leaflet doesn't have any SCSS file. Talking with @carlosvillu we concluded that this was the less painful solution. Modify .css dist file to .scss and delete the non used background-images to dismiss image load problems.

components/map/basic/src/index.js
carlesnunez: This is not true at all. This fn will check if there's a change on the view to be loaded. If there's, will load it, but if for example we call again this FN and nothing changes the function won't do anything.

components/map/basic/src/index.js
carlesnunez: Same here. Is not a pure toggling function. I thought the same on my first function view but later I realised that is more a check&execute function than a pure toggle one.

components/map/basic/src/index.js
carlesnunez: Wow I didn't saw that! 🔝

components/map/basic/src/index.js
carlesnunez: Totally true, changed.

components/map/basic/src/index.js
carlesnunez: Fully agree with that. Next time I must check all the code on finish to re-refactor it.

Changed!

components/map/basic/leafletLocalInstall.js
davidbarna: I may be wrong but I disagree.
Background images will only be loaded if they are actually used in the DOM. But, if you add css statement to override waht you want you're done.

About the css file, you should be able to import css code into sass, right ??? 
http://sass-lang.com/documentation/file.SASS_CHANGELOG.html#CSS__import_Directives
All valid css is valid scss, by definition.

davidbarna (approved):
Make sure the coupled code will be abstracted in the future

components/map/basic/src/leaflet/layer-manager.js
carlesnunez: Is not exactly and ID, i've changed the naming to have a more clear view about what it is. This is the 'version' of the tile. Newest will get the newset 'topographic' version of the tile

components/map/basic/src/index.js
```diff
@@ -0,0 +1,257 @@
+import React, {Component} from 'react'
+import PropTypes from 'prop-types'
+import LeafletMap from './leaflet/map'
+import { mapViewModes, NO_OP } from './leaflet/constants'
+
+class MapBasic extends Component {
+  constructor (props) {
+    super(props)
+    this.setMapEventDefinition(props)
```
betobe: Unused parameter

carlesnunez: Which one is not used?

PR #225, opened 2017-11-14
Feature/update default variables rangedatepicker

davidmartin84:


components/form/rangeDatepicker/src/index.scss
```diff
@@ -19,6 +23,20 @@ $datepicker__selected-color: $bgc-range-datepicker-selected;
     align-items: center;
     display: flex;
     padding-right: $p-xsmall;
+    width: $w-range-datepicker-input;
```
davidbarna, line 26: Is `!important` necessary in this case ????
Please avoid `!important` and use css weight instead

davidbarna (changes requested):

components/form/rangeDatepicker/src/index.scss
davidmartin84: I need !important when i update the fill property

davidbarna (approved):

PR #233, opened 2017-11-16
SUI Color Panel

jcger:


components/atom/colorPanel/README.md
```diff
@@ -0,0 +1,25 @@
+# AtomColorPanel
+
+> A color panel is a container with a color on its background.
+
+## Installation
+
+```sh
+$ npm install @schibstedspain/sui-atom-colorPanel --save
```
davidbarna: Do we use uppercase for package names ???

components/atom/colorPanel/src/index.js
```diff
@@ -0,0 +1,49 @@
...
+    'sui-AtomColorPanel',
+    `sui-AtomColorPanel-${type}`
+  )
+
+const getStyle = ({opacity}) => ({
+  opacity
+})
+
+const AtomColorPanel = ({children, ...props}) => (
+  <div className={getClassNames(props)} style={getStyle(props)}>
```
davidbarna: Is the opacity supposed to be for the component or for its background ?

davidbarna (approved):

components/atom/colorPanel/README.md
jcger: nope, its the autogenerated readme that does this and I didn't check it. will be fixed

components/atom/colorPanel/src/index.js
jcger: background, thanks!

davidbarna: Then I request the change !! :-)

JoanClaret:
Not sure if we need a component to assign a background-color.
i would propose to define the variables in `sui-theme`, and do not develop a React component.
but maybe i'm missing something!

```scss
$c-color-panel-canvas: $c-white !default;
$c-color-panel-default: color-variation($c-gray, 4) !default;
$c-color-panel-base: color-variation($c-gray, 3) !default;
$c-color-panel-contrast: $c-black !default;
$c-color-panel-corporate: $c-primary !default;
$c-color-panel-highlight: color-variation($c-accent, 4) !default;
```

nguasch:
@JoanClaret it make sense. I think that @ julian-gernun has the idea to merge `Color Image` component with this one, but maybe one good option would be had the `Color Panel` as a mixin in `sui-theme` and use it in `Image Panel` component

jcger:
Closing this because we will create a mixing as color panel and use it in a "sui-atom-panel" component which will be "image-panel" + "color-panel"

PR #235, opened 2017-11-17
fix(map/basic): fix error in leafletLocalInstall.js script.

ronaldo-mh:
There is an error installing the sui-map-basic component as dependency of another component.

Basically, when executing the postInstall script `leafletLocalInstall.js`, the node_modules path where leaflet package is looked up is not always right.
This path is calculated with `path.join(__dirname, '/node_modules/leaflet/dist/leaflet.css')`, but this is what happens:

- When installing it in sui-components repo, it works correctly as leaflet is installed directly in the node_modules directory in the root path of sui-map-basic.

- But when installing as dependency of another component, the `leafletLocalInstall.js` has as __dirname `main-component/node_modules/@schibstedspain/sui-map-basic`, and the leaflet package is in the node_modules folder 3 levels upper.

So, this fix simply uses **require.resolve** to calculate the leaflet module path.

### DIFF
--- a/components/map/basic/leafletLocalInstall.js
+++ b/components/map/basic/leafletLocalInstall.js
@@ -1,9 +1,7 @@
 var fs = require('fs')
-var path = require('path')
 
-var basePath = __dirname.split('node_modules')[0]
-var originPath = path.join(basePath, '/node_modules/leaflet/dist/leaflet.css')
-var destinationPath = path.join(basePath, '/node_modules/leaflet/dist/leaflet.scss')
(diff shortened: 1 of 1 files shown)


carlesnunez:
We should use require.ensure. I forgot to add that on my component. Sorry :/

carlesnunez (approved):
Cool! Thanks for the fix and sorry for forget it!! 👎

davecarter (approved):

davidbarna (approved):

davidbarna (approved):

PR #238, opened 2017-11-20
Component/atom image panel

PablitoGS:


jcger:
Am not sure if we should kind of merge color panel and image panel and just create an atom panel component

components/atom/imagePanel/src/index.js
```diff
@@ -0,0 +1,110 @@
...
+  RIGHT: 'right'
+}
+
+const VERTICAL_ALIGNMENTS = {
+  TOP: 'top',
+  CENTER: 'center',
+  BOTTOM: 'bottom'
+}
+
+const TYPES = {
```
jcger: not used

components/atom/imagePanel/src/index.js
```diff
@@ -0,0 +1,110 @@
...
+  horizontalAlign: VERTICAL_ALIGNMENTS.CENTER,
+  overlayColor: '#000',
+  overlayAlpha: 0
+}
+
+export default AtomImagePanel
+export {
+  HORIZONTAL_ALIGNMENTS as atomImagePanelHorizontalAlign,
+  VERTICAL_ALIGNMENTS as atomImagePanelVerticalAlign,
+  TYPES as atomItemPanelTypes
```
jcger: should be removed as TYPES is bool

davidbarna:
Please provide links to the definition and/or design

demo/atom/imagePanel/playground
```diff
@@ -0,0 +1,69 @@
+const imageSrcLong = 'http://lorempicsum.com/simpsons/1000/100/1'
```
davidbarna, line 1: * We should add examples of all use cases exposed in the zepplin.

components/atom/imagePanel/src/index.js
```diff
@@ -0,0 +1,110 @@
...
+  /**
+   * Gradient opacity
+   * @type float between 0 and 1
+   */
+  overlayAlpha: PropTypes.number,
+  /**
+   * Background position x
+   */
+  horizontalAlign: PropTypes.oneOf(Object.values(HORIZONTAL_ALIGNMENTS)),
+  /**
```
davidbarna, line 89: The text is always black. 
What happened we don't provide the overlayColor  and it has it's default #000. It's black text on dark background. Have tried ?

```js
<AtomImagePanel src={imageSrcLong} overlayAlpha={0.9} />
```

components/atom/imagePanel/src/index.js
```diff
@@ -0,0 +1,110 @@
...
+  placeholderColor: PropTypes.string.isRequired,
+  /**
+   * Gradient color
+   */
+  overlayColor: PropTypes.string,
+  /**
+   * Gradient opacity
+   * @type float between 0 and 1
+   */
+  overlayAlpha: PropTypes.number,
```
davidbarna, line 84: According to zepplin, alpha is always `0.5`.
We should only make customizable what's meant to be by design.

components/atom/imagePanel/src/index.js
```diff
@@ -0,0 +1,110 @@
...
+    resized && `sui-AtomImagePanel-resized`,
+    classNames
+  )
+}
+
+const getStyles = function ({placeholderColor, overlayColor, src, overlayAlpha}) {
+  const url = `url(${src})`
+  if (overlayColor) {
+    const overlayColorRgb = hexToRgb(overlayColor)
+    const rgba = `rgba(${overlayColorRgb.r}, ${overlayColorRgb.g}, ${overlayColorRgb.b}, ${overlayAlpha})`
```
davidbarna: Do we really to add a gradient that is not a gradient but background with alpha color ? 
I get that we're trying to minimize the number of DOM nodes. But, are we measuring which is better between one more nodes or repainting a gradient twice ?
Have you consider using Before or After pseudo elements ?

components/atom/imagePanel/src/index.js
jcger: It's being exported and therefore "used"

components/atom/imagePanel/src/index.js
PablitoGS: Finally we will use the :: before pseudo element

components/atom/imagePanel/src/index.js
PablitoGS: we'll create a  sass function to fix this

components/atom/imagePanel/src/index.js
davidbarna: We found it out later LOL

components/atom/imagePanel/src/index.js
PablitoGS: But acording to the defintion: 
It’s optional and has a certain amount of opacity that can be configured.

davidbarna: Ok. I thought they meant the opacity could be configure by theme, like the color.

components/atom/imagePanel/src/index.js
jcger: Shouldn't the child component be able to render itself? I mean, we do not just set an image as background for an label, there can be any kind of component inside this container, which means that it could happen that if the component itself has a white background, but the overlay is dark, that the user sees nothing as we would set the font color to white.

In your example, I would add an class to the component that sets the color to white as we already know that the background is dark

demo/atom/imagePanel/playground
jcger: What example do you mean exactly? If it is the switched/replace example, it does not apply to this version as mentioned as comment in https://paper.dropbox.com/doc/SUI-Image-Panel-AjDu7OEMGBGaJaHiK8Vlf

I'll add a white text over black background example cause I don't see what else is missing

davidbarna: The white text for instance, yes. Putting all the examples you make sure to see the behavior.
Every behavior/style that you develop should be documented by example.

jcger:
Closing this as we'll add create a new PR with color + image panel merged as atom panel component

PR #248, opened 2017-11-23
feat(tag/selectableList): add option: initialValues to selectableList

davidmartin84:


components/tag/selectableList/src/index.js
```diff
@@ -8,7 +8,7 @@ const removeFromArray = (arr, value) => {
 
 export default class TagSelectableList extends Component {
   state = {
-    selectedValues: []
+    selectedValues: this.props.initialValues
```
jcger: dunno if it's the most descriptive name, is initialSelectedValues to long or what?

davidmartin84: i change it!

jcger (approved):

PR #256, opened 2017-11-28
feat(profile/rating) Sui profile rating component to show user rating data

carlesnunez:
**What does this PR do?**
This pr creates the sui-profile-rating component, created to allow us to show rating information to the user in a fancy way.
**What are the relevant tickets?**
No SUI tickets related with this issue
Created by need on vibbo widget 0
mygif

**Does the documentation need an update?**
N/A


components/profile/rating/src/index.js
```diff
@@ -0,0 +1,92 @@
+import React, { Component } from 'react'
+import PropTypes from 'prop-types'
+
+class ProfileRating extends Component {
+  constructor (props) {
```
midudev: Init the state like this:
```
state = { ratingFilledWidth: 0 }
```

components/profile/rating/src/index.js
```diff
@@ -0,0 +1,92 @@
...
+      })
+    }
+  }
+
+  render () {
+    const { maxValue, icon, fillColor, emptyColor } = this.props
+
+    return <div className='sui-ProfileRating'>
+      <div className='sui-ProfileRating-empty' style={{ fill: emptyColor }}>
+        {this.buildRating(maxValue, icon)}
```
midudev: You're calculating this twice in the render. Store it in a var and access later. 

```
const ratingBuilt = this.buildRating(maxValue, icon)
{ratingBuilt}
```

midudev:
The component looks amazing, congrats! The context of the rating doesn't sound good as it could be potentially used outside a Profile. I think might be more appropriated something like `@schibstedspain/sui-rating-icons` as we could have different rating systems (with numbers, with top/down arrows, with strings...).

carlesnunez:
@miduga we can change the scope of the rating and set as a category itself. I like the idea. something like 

rating/icons

components/profile/rating/src/index.scss
```diff
@@ -0,0 +1,30 @@
+@import '~@schibstedspain/theme-basic/lib/index';
+
+$fill-profile-rating-filled: #f7af33 !default;
```
davecarter: We used to set default Sass values [here](https://github.com/SUI-Components/sui-theme/blob/master/src/settings-compat-v7/_components.scss) and then the corporate theme rewrite these values.
Since we are not doing this any more it would be nice to have Sass Vars declarations inside a separate partial file like:
`_settings.scss`

components/profile/rating/src/index.js
carlesnunez: Is it a convention? Is more clean for me to have it as a constructor declaration but I don't have any problem to change it.

components/profile/rating/src/index.scss
```diff
@@ -0,0 +1,30 @@
...
+$fill-profile-rating-empty: #d7dbdf !default;
+
+.sui-ProfileRating {
+  height: 16px;
+  margin-top: 7px;
```
davecarter: Margins and Paddings should be base8. Thera are a set of declarations [here](https://github.com/SUI-Components/sui-theme/blob/master/src/settings/_spacing.scss) you should use instead

components/profile/rating/src/index.js
```diff
@@ -0,0 +1,92 @@
...
+  }
+
+  calculateWidthByRating () {
+    const { maxValue, iconSize, rating, spacingBetween } = this.props
+
+    return (maxValue * (iconSize + spacingBetween) / maxValue) * rating
+  }
+
+  getIcon (Icon, index) {
+    return <div key={index} className='sui-ProfileRating-iconContainer' style={{ minWidth: this.props.iconSize, marginLeft: this.props.spacingBetween }}>
```
davecarter: you might deconstruct iconSize and spacingBeteen here too:
`const { iconSize, spacingBetween } = this.props`

components/profile/rating/src/index.scss
carlesnunez: I asumed talking with @JoanClaret and @frmedina that there's a new way to declare component variables that is inside the component.

carlesnunez:
PR Changes done, also styling variables rename and component renaming

midudev (approved):

demo/rating/icons/themes/_vibbo.scss
```diff
@@ -0,0 +1,2 @@
+@import '../../../../themes/vibbo';
+@import '../../../../components/rating/icons/src/index.scss';
```
davecarter: Sass files does not need to specify file extension when importing

carlesnunez: Is changed! BTW I think that all the components that have themes in sui are importing the index in this way

components/profile/rating/src/index.scss
davecarter: This is exactly what I'm saying. Since we have a new way to declare Sass variables inside the component, use at least a separate partial file called `_settings.scss`

davecarter (changes requested):
please, use a `_settings.scss` separate sass partial file to add variable declarations as used [here](https://github.schibsted.io/scmspain/frontend-fc--uilib-components/blob/master/components/card/property/src/_settings.scss)

kikoruiz:
Hi @carlesnunez, I reallized there is something similiar within the AdSmartbanner component: https://github.com/SUI-Components/sui-components/blob/master/components/ad/smartbanner/src/rating-star.js (this only shows a rating, but your new component also does that), could we do a merge of these components before keep going with this PR?

components/rating/icons/src/index.js
```diff
@@ -0,0 +1,90 @@
...
+   * The color used for the filled mask. If you don't provide any color, it will fallback to our component scss variable $fill-profile-rating-filled
+   */
+  fillColor: PropTypes.string,
+  /**
+   * The color used for the empty mask. If you don't provide any color, it will fallback to our component scss variable $fill-profile-rating-empty
+   */
+  emptyColor: PropTypes.string
+}
+
+RatingIcons.defaultProps = {
```
davidbarna, line 87: What about defining more default values.
Default icon size, default maxValue, etc

davidbarna (approved):

components/rating/icons/src/index.js
carlesnunez: I prefer to keep this as is, if it start to be painful for us to write the props I'll change it.

carlesnunez:
@kikoruiz I prefer to don't merge by the moment this two 'rating' components. The main thing to that is that we currently have a 1 icon state rating instead of the 3 icon states that the smartbanner currently need. Maybe in the future we can merge the two types of ratings or UX+UI can define one that will be SUI compliant.

davecarter (approved):
Good Job! 👍

davecarter:
@davidbarna => We decided to do it this way a while ago, and actually same pattern is followed by SUI-Components

PR #257, opened 2017-11-28
feat(profile/sidebarcard): created the sidebarcard for the profile user

carlesnunez:
**What does this PR do?**
This pr creates the sui-profile-sidebarcard component, created to allow us to show user related information on as a card.
**What are the relevant tickets?**
No SUI tickets related with this issue
Created by need on vibbo widget 0
screen shot 2017-11-28 at 20 08 16

**Does the documentation need an update?**
N/A


components/profile/sidebarcard/src/Avatar.js
```diff
@@ -0,0 +1,14 @@
+import React from 'react'
```
midudev: I would like that's acceptable right now, but it would be a good idea to think that this component would potentially be an independent Sui Avatar component. :)

carlesnunez: Iup we've talked today with our UX+UI guys about the same thing. In the future as we are going to work closely with UI-UX to be sui compliant we will made a definition of this component and export it as an atom component.

midudev (approved):

carlosvillu (approved):

components/profile/sidebarcard/README.md
```diff
@@ -0,0 +1,52 @@
+# ProfileSidebarcard
```
JoanClaret: Since we already have a `card` context defined in `sui-components` i would define this component as `card/profile`.
In my opinion, we should not reference the location where it's gonna be placed (Sidebar), becasuse we don't know where will we use it in the future.

carlesnunez: Done, renamed the component

components/card/profile/src/_settings.scss
```diff
@@ -0,0 +1,11 @@
+$bd-card-profile-avatar-halo: 4px solid $c-accent !default;
+$bgc-card-profile: $c-gray-lightest !default;
+$c-profile-info: $c-gray !default;
```
JoanClaret: all variables should contain the context: `$c-card-profile-info`

components/card/profile/src/Avatar.js
```diff
@@ -0,0 +1,14 @@
+import React from 'react'
+import PropTypes from 'prop-types'
+
+const Avatar = ({ profilePic }) => <div className='sui-CardProfile-topContainerAvatarHalo'>
```
JoanClaret: Can we use a shorter classname?
for instance: `sui-CardProfile-avatarHalo`

components/card/profile/src/Avatar.js
```diff
@@ -0,0 +1,14 @@
...
+  <div className='sui-CardProfile-topContainerAvatarRadiusWrapper'>
```
JoanClaret: i would use a shorter name here too: `sui-CardProfile-avatarRadius`

components/card/profile/src/CounterBox.js
```diff
@@ -0,0 +1,15 @@
+import React from 'react'
+import PropTypes from 'prop-types'
+
+const CounterBox = ({number, label, link}) => <a href={link} className='sui-CardProfile-counterBox'>
+  <div className='sui-CardProfile-counterBoxNumber'>{number}</div>
```
JoanClaret: what do you think about: `sui-CardProfile-counterNumber` ?

components/card/profile/src/_settings.scss
```diff
@@ -0,0 +1,11 @@
...
+$ff-profile-rating-info-box-name: 'Open Sans', sans-serif !default;
+
+$h-card-profile: 280px;
+$w-card-profile: 172px;
```
JoanClaret: what do you think about the component adapt to its parent? IMO we should try to avoid fixing withs and heights.

components/card/profile/src/index.js
```diff
@@ -0,0 +1,64 @@
...
+   */
+  maxRatingValue: PropTypes.number.isRequired,
+  /**
+   * The icon to be painted as your rating icon.
+   */
+  icon: PropTypes.func.isRequired,
+  /**
+   * The icon size to allow the system to make their math calcs to build fill width's.
+   */
+  iconSize: PropTypes.number.isRequired
```
JoanClaret, line 61: icon sizes are already defined in sui-theme:
https://github.com/SUI-Components/sui-theme/blob/master/src/settings/_size.scss
What about to stick at this definition?

components/card/profile/src/index.scss
```diff
@@ -0,0 +1,76 @@
...
+        border-radius: 50%;
+        display: block;
+        height: $w-card-profile-avatar;
+        margin-left: auto;
+        margin-right: auto;
+        margin-top: $m-s;
+        overflow: hidden;
+        width: $w-card-profile-avatar;
+
+        img {
```
JoanClaret: i would suggest to add a class instead of point the html tag.

for instance: `sui-CardProfile-avatarImage`

JoanClaret (commented):
Some comments about classnames too long and other minor things. good job overall!

components/card/profile/README.md
```diff
@@ -0,0 +1,52 @@
+# CardProfile
+
+> Description
+
+ProfileSideBar card serves a card with user information like avatar, name, rating bar and stats information.
```
JoanClaret: `ProfileSideBar card` -> `Card Profile` ?

components/card/profile/README.md
```diff
@@ -0,0 +1,52 @@
...
+Can be putted everywhere but its main purpose is to be added as a sidebar element.
+
+## Installation
+
+```sh
+$ npm install @schibstedspain/sui-card-profile --save
+```
+
+## Usage
+The sidebar card need a few properties to be setted up:
```
JoanClaret: `The sidebar card` -> `The profile card` ?

components/card/profile/src/_settings.scss
```diff
@@ -0,0 +1,11 @@
+$bd-card-profile-avatar-halo: 4px solid $c-accent !default;
+$bgc-card-profile: $c-gray-lightest !default;
+$c-profile-info: $c-gray !default;
+$ff-profile-rating-info-box-name: 'Open Sans', sans-serif !default;
```
JoanClaret: Missing context in variable name. 
Anyways, is this definition really needed? shouldn't be the font family defined at page level?  or this componente has a different font?

components/card/profile/src/index.js
carlesnunez: The icon size is not used for a stilying purpose, is for a calc purpose. The filling bar need to know the width of their icon before being created if no iconsize is provided the calc can't be done and the mask can't be created. We could check for a better solution on a next iteration of the component if you want 👍

components/card/profile/src/CounterBox.js
```diff
@@ -0,0 +1,15 @@
+import React from 'react'
+import PropTypes from 'prop-types'
+
+const CounterBox = ({number, label, link}) => <a href={link} className='sui-CardProfile-counter'>
+  <div className='sui-CardProfile-counterNumber'>{number}</div>
```
davidbarna: Let me know if I'm wrong but divs are not "allowed" inside <a> tags I think

components/card/profile/src/index.js
davidbarna: I would say could modify the implementation with no need of the size.

davidbarna (approved):

PR #258, opened 2017-11-29
feat(element/breakpointToggle): add inverse mode to breackpoint toggle

davidmartin84:


components/element/breakpointToggle/src/index.js
```diff
@@ -23,9 +23,10 @@ class BreakpointToggle extends Component {
   }
 
   onViewportResize (width, height) {
-    let {breakpoint} = this.props
+    let {breakpoint, inverse} = this.props
```
carlesnunez: use const instead of let

davidmartin84: @carlesnunez  done!

jcger (approved):

PR #261, opened 2017-12-01
Feature/buttons link

JoanClaret:
Adding attribute to render button links:

image

### DIFF
--- a/components/atom/button/README.md
+++ b/components/atom/button/README.md
@@ -64,6 +64,6 @@ output:
 <a class="sui-AtomButton sui-AtomButton--link" href="http://www.schibsted.com/">Link</a>
 ```
 
-Note: Since this `disabled` attribute is not valid for a link, this prop will not be assigned to the html tag.
+Note: Since `disabled` attribute is not valid for a link, this prop will not be assigned to the html tag, but a `sui-AtomButton--disabled` class will be added.
 
 > **Find full description and more examples in the [demo page](https://sui-components.now.sh/workbench/atom/button).**
--- a/components/atom/button/README.md
+++ b/components/atom/button/README.md
@@ -27,7 +27,7 @@ return (<div>
 
 ### Flexible props
 
-All props available from regular buttons an be used.
+All props available from regular buttons can be used.
 
 ```js
 import Button from '@schibstedspain/sui-atom-button'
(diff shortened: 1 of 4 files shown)


components/atom/button/README.md
```diff
@@ -46,5 +46,24 @@ return (<div>
...
 ```
 
+### Rendering a link
+When `link` property is passed, the component will render an html link.
+
+```js
+
+return (
+  <Button link url='http://www.schibsted.com/'>Link</Button>
```
kikoruiz: Why not unify props in just one? E.g.: `<Button link='http://www.schibsted.com/'>Link</Button>` So that, we would use just one string prop, instead of a string and a boolean.

components/atom/button/src/index.js
```diff
@@ -45,19 +56,38 @@ const AtomButton = (props) => {
...
+  )
 }
 
 AtomButton.displayName = 'AtomButton'
 
 AtomButton.propTypes = {
+  /**
+   * HTML element: if true, render a link. Otherwise render a button.
+   */
+  link: PropTypes.bool,
```
midudev: I would say that if the `url` string is passed, the `link` prop is implicit. Isn't?

davecarter (approved):

components/atom/button/README.md
danivm: +1

davidbarna: Agree with Kiko,
But, on the other hande, knowing that a target props can be applied, we may do 

```html
<Button link href='http://www.schibsted.com/'>Link</Button>
<Button link href='http://www.schibsted.com/' target='_blank'>Link</Button>
```

`link` prop would just mean "convert to link" and then you can use whatever `<a>` specific props.

components/atom/button/src/index.js
```diff
@@ -45,19 +56,38 @@ const AtomButton = (props) => {
...
-    {leftIcon && <span className={`${CLASS}-leftIcon`}>{leftIcon}</span>}
-    {leftIcon || rightIcon
-      ? <span className={`${CLASS}-text`}>{children}</span>
-      : children
-    }
-    {rightIcon && <span className={`${CLASS}-rightIcon`}>{rightIcon}</span>}
-  </button>)
+
+  const Button = ({ children, url, disabled, ...attrs }) => link
+    ? <Link {...attrs} href={url}>{children}</Link>
```
davidbarna: What do we add the "link" type AND the linkFactory on top of it.
I would either provide a elementFactory to what we already had OR implement the link prop.
Do you see cases where linkFactory would be really needed if we already inherited all the props... ?

components/atom/button/README.md
JoanClaret: well seen! done!

JoanClaret: ok, i get your point @davidbarna. then, what if we send an object?
```html
<Button link={ href: 'http://www.schibsted.com/', target:'_blank'}>Link</Button>
```

davidbarna: We are trying to be as HTML standards compliant as possible. I really think if should be defined by props.

To me, the best API would be either:

```js
<Button link href='http://www.schibsted.com/' target='_blank'>Link</Button>
````
or 

```js
import {Link} from 'sui-atom-button'
<Link href='http://www.schibsted.com/' target='_blank'>Link</Link>
````

Comment (JoanClaret) [reply components/atom/button/README.md]: ok, take a look @davidbarna. developed like so:

```html
<Button link href='http://www.schibsted.com/' target='_blank'>Link</Button>
```

davidbarna (approved):

PR #263, opened 2017-12-04
feat(atom/button): fix icons alignment

JoanClaret:
align icons properly using flexfox instead of line-height:

```scss
.sui-AtomButton-inner {
    align-items: center;
    display: inline-flex;
    height: 100%;
}
```
Note that we've applied the flexbox in to the layer `inner`. 
That is because some browsers does not support flex in `button` elements.
https://stackoverflow.com/questions/35464067/flexbox-not-working-on-button-or-fieldset-elements

btw, we've forced the icon size, in order to avoid flexbox resizing it
```scss
[class*="Icon"] {
    height: $size;
    width: $size;
}
```

### DIFF
--- a/components/atom/button/src/index.js
+++ b/components/atom/button/src/index.js
@@ -63,12 +63,14 @@ const AtomButton = (props) => {
 
   return (
     <Button {...newProps} className={classNames} title={title} disabled={disabled}>
-      {leftIcon && <span className={`${CLASS}-leftIcon`}>{leftIcon}</span>}
-      {leftIcon || rightIcon
-        ? <span className={`${CLASS}-text`}>{children}</span>
-        : children
-      }
-      {rightIcon && <span className={`${CLASS}-rightIcon`}>{rightIcon}</span>}
+      <div className={`${CLASS}-inner`}>
+        {leftIcon && <span className={`${CLASS}-leftIcon`}>{leftIcon}</span>}
+        {leftIcon || rightIcon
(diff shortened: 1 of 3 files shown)


danivm (approved):

devtoni (approved):

davidmartin84 (approved):

jcger:
What was the issue?

components/atom/button/src/index.js
```diff
@@ -63,12 +63,14 @@ const AtomButton = (props) => {
 
   return (
     <Button {...newProps} className={classNames} title={title} disabled={disabled}>
-      {leftIcon && <span className={`${CLASS}-leftIcon`}>{leftIcon}</span>}
-      {leftIcon || rightIcon
-        ? <span className={`${CLASS}-text`}>{children}</span>
-        : children
-      }
-      {rightIcon && <span className={`${CLASS}-rightIcon`}>{rightIcon}</span>}
+      <div className={`${CLASS}-inner`}>
```
davidbarna, line 66: Div not allowed in button or a as only phrasing content is allowed.

JoanClaret:
@julian-gernun icons was not aligned properly due to a line height definition. so we decided to use flex to align items properly.

components/atom/button/src/index.js
JoanClaret: ok, we should use an span instead.

PR #267, opened 2017-12-11
feat(collapsible/basic): remove max-height limitation in collapsible content.

ronaldo-mh:
When we have a very large content in the collapsible component, this content is currently limited to the current height of screen (max-height: 100%). Any overflowing elements are hidden.

I'm removing this limitation, since I don't see its possible usefulness.

### DIFF
--- a/components/collapsible/basic/src/_settings.scss
+++ b/components/collapsible/basic/src/_settings.scss
@@ -11,3 +11,6 @@ $c-collapsible-basic: $c-gray-dark !default;
 $bdc-collapsible-basic: $c-gray-lightest !default;
 $bd-collapsible-basic: 1px solid $bdc-collapsible-basic !default;
 $trs-collapsible-basic-trigger-icon: transform $trs-base !default;
+$trs-collapsible-basic-content: .3s ease-out;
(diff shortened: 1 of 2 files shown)


JoanClaret:
i think this max-height was set to animate on show content. 
it's related to:
```scss
&.is-collapsed {
    #{$root-parent}-collapsibleContent {
      max-height: 0;
    }
}
```

if we remove it, we'll loose the animation, isn't it?

ronaldo-mh:
@JoanClaret yep, you are right, it has sense.
Mi problem now is that I don'w want that the content is limited to the height of the screen (in our domain component, we may have a very long list of elements as content). I'll figure out another solution. Thanks!

ronaldo-mh:
In the end, I didn't find a proper solution when the content of the collapsible is too large and is cut-off by the `max-height: 100%` style.

So I am keeping the max-height in order not to lose the animation, and adding a customisable overflow property (set by default to `overflow: scroll`).

I am also making the max-height customisable, since our solution will be to use pagination when our list of results is too long, and we might need to reduce the max-height of the content to a fixed value for animation.

PR #268, opened 2017-12-11
feat(card/collapsible): expose click handler func of collapsible.

ronaldo-mh:
- Add prop `onChangeHandler` to allow customisation of `handleClick` prop of collapsible.
- Add box-shadow style to card only when it is expanded.
- Move scss vars to _settings.scss partial file.

### DIFF
--- a/components/card/collapsible/src/index.scss
+++ b/components/card/collapsible/src/index.scss
@@ -2,16 +2,14 @@
 @import '~@schibstedspain/sui-theme/lib/index';
 @import '~@schibstedspain/sui-button-basic/lib/index';
 @import '~@schibstedspain/sui-collapsible-basic/lib/index';
-
-$bgc-card-collapsible: $c-white !default;
-$h-card-collapsible-header-image: 100px;
-$c-card-collapsible-info-item: $c-gray-dark !default;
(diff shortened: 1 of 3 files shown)


components/card/collapsible/src/index.scss
```diff
@@ -2,17 +2,15 @@
...
-$c-card-collapsible-info-item: $c-gray-dark !default;
-$fz-card-collapsible-info-item-collapsed: $fz-l + 2 !default;
-$fz-card-collapsible-info-item-expanded: $fz-m !default;
-$lh-card-collapsible-info-item: 1.2 !default;
+@import 'settings';
 
 .sui-CardCollapsible {
   background-color: $bgc-card-collapsible;
 
+  &.has-shadow {
```
JoanClaret: What about using `is-expanded` class? so we will be able to add other properties.

ronaldo-mh: Yes, sure, I can reuse `is-expanded` class. I think it has sense since `has-shadow` is not set directly by a prop itself.

JoanClaret (approved):

davidbarna (approved):

PR #271, opened 2017-12-12
SUI Input

jcger:


components/atom/input/README.md
```diff
@@ -0,0 +1,114 @@
...
+
+### Basic usage
+```js
+import MaskInput from '@schibstedspain/sui-atom-input/lib/inputs/MaskInput'
+import AtomInput from '@schibstedspain/sui-atom-input'
+
+// render the same
+return (
+  <div>
+    <MaskInput label='postal code' name='date' mask={postalCodeMask}/>
```
davidbarna, line 22: What about this ?

```js
import {Mask} from '@schibstedspain/sui-atom-input'
```

the option of `/lib/inputs/MaskInput` is only for optimization purpose I guess ?

Anyways I would only show one way to do it.

components/atom/input/src/InputWrapper.js
```diff
@@ -0,0 +1,23 @@
+import React from 'react'
+import PropTypes from 'prop-types'
+import InputLabel from './Label'
+
+const InputWrapper = (props) => {
+  const {name, label, children} = props
+  return (
+    <div className='sui-AtomInput'>
```
davidbarna, line 8: I think the best practice is to enclose inputs inside their label for accessibility. Maybe Pablo has a better opinion.

davidbarna, line 8: Are you sure you need a div. By default a span is the wrapper with less layout effects.
If you take into account my previous comment you could event return the label element as a wrapper as you're not using the sui-AtomInput class....

components/atom/input/src/inputs/MaskInput.js
```diff
@@ -0,0 +1,47 @@
+import React, {Component} from 'react'
+import PropTypes from 'prop-types'
+import IMask from 'imask'
+import InputWrapper from '../InputWrapper'
+
+class MaskInput extends Component {
+  componentDidMount () {
+    const mask = this.props.mask
```
davidbarna: Choose a way to obtain a single prop :-)
either line 8 or line 16

components/atom/input/src/inputs/MaskInput.js
```diff
@@ -0,0 +1,47 @@
...
+          onChange={this.onChange}
+        />
+      </InputWrapper>
+    )
+  }
+}
+
+MaskInput.displayName = 'MaskInput'
+
+MaskInput.propTypes = {
```
davidbarna, line 39: Why are we limiting props here?
For instance, why can't I use title or tabindex.
We need a solution to allow "global" input attributes at least.

components/atom/input/src/inputs/TagInput/index.js
```diff
@@ -0,0 +1,76 @@
...
+    this.props.onChange && this.props.onChange({value: this.value})
+  }
+
+  render () {
+    const {name, label} = this.props
+    return (
+      <InputWrapper name={name} label={label} >
+        {
+          this.state.tags.map((label, idx) =>
+            <AtomTag label={label} key={idx} onClose={() => this.removeTag(idx)} />
```
davidbarna, line 61: I see in the paper nothing specifies if you can remove tag by backspace, select them, create them with ENTER or something :-(

davidbarna (changes requested):

components/atom/input/README.md
jcger: I have no preference about it, maybe someone else want to give his opinion about this

Yes, it's "optimization"

components/atom/input/src/inputs/TagInput/index.js
jcger: There is this line: "To see how Tags behave inside an Input, please check Autosuggest Component." with a link to https://paper.dropbox.com/doc/SUI-Autosuggest-VHYPgfOTsq4pt50EdimCL. In here there is a tag type in which it says that tags are generated by commas. To remove one Tag, you'll have to click on the tag delete icon

components/atom/input/src/index.js
```diff
@@ -10,6 +10,7 @@ import InputWrapper from './InputWrapper'
 const NATIVE_TYPES = ['text', 'number', 'date']
 
 const changeHandler = (ev, cb) => {
+  // eslint-disable-next-line
```
davidbarna: You should not disable rules until it's necessary.

components/atom/input/src/index.scss
```diff
@@ -1,5 +1,73 @@
...
-.sui-AtomInput {
+$validation: success $c-success, error $c-error;
+$size: medium 40px, small 30px;
+
+@mixin input {
+  border-color: $c-gray-light;
+  border-style: solid;
+  border-width: 1px;
+  font-size: $fz-base;
+  height: 40px;
```
davidbarna, line 12: Isn't it a way to express this height with our theme variables ?

components/atom/input/src/index.scss
```diff
@@ -1,5 +1,73 @@
...
+  padding-left: $p-l;
+  padding-right: $p-l;
+  width: 100%;
+
+  &[placeholder] {
```
davidbarna, line 17: Attribute selectors very innefficient. Moreover, here we have doubles like '[disabled][placeholder]'. Better use classes to identify sub elements

davidbarna (changes requested):

components/atom/input/src/index.js
jcger: Sorry, I told him to do it so he could commit while I'm looking why the linter is failing

components/atom/input/src/InputWrapper.js
jcger: It's ok as long as you use the label `for` attribute. Your approach is recommended if it's not possible to set the `for` attribute

More info here https://www.w3.org/WAI/tutorials/forms/labels/ where you can read that `Generally, explicit labels are better supported by assistive technology.`

PR #277, opened 2018-01-04
SUI AtomPanel

jcger:
We merged AtomImagePanel and AtomColorPanel as AtomPanel

Definitions
- https://paper.dropbox.com/doc/SUI-Image-Panel-AjDu7OEMGBGaJaHiK8Vlf
- https://paper.dropbox.com/doc/SUI-Color-Panel-viiQMX2koZ102o0Jukb3W


components/atom/panel/src/ImagePanel.js
```diff
@@ -0,0 +1,86 @@
...
+  return cx(
+    `${IMAGE_PANEL_CLASS}--vertical-${verticalAlign}`,
+    `${IMAGE_PANEL_CLASS}--horizontal-${horizontalAlign}`,
+    overlayColor && `${BASE_CLASS}--${overlayColor}-overlay-${overlayAlphaValue}`,
+    `${BASE_CLASS}-color--${color}`,
+    resized && `${IMAGE_PANEL_CLASS}--resized`
+  )
+}
+
+const getStyles = function ({src}) {
```
midudev, line 41: I would avoid using `const` to store anonymous functions but create named functions instead.

`function getStyles ({ src }) {`

davidbarna: I'm interested in knowing why.

midudev: It's a personal preference, I wouldn't like to be a must, just to be considered. :)

First, anonymous functions are hard to read in an Error stack (I think, only in latest Chrome editions seems that they're getting automagically named by the const). And that's because the Function.name is not being set. Normally, the only exception I'm up is with inline functions inside an Array.prototype method.

Also, there's no need to identify the function as a let, const or var as it could work as a function expression. It's not like we need to know that the value is immutable or that we're going to modify it. Nor the hoisting is important as clearly we want these functions to be used in general in the file.

At finally, for me at least, is more readable having a function expression than a const with a function. But, as I said, it's my personal choice! :) Anyway, I would be interested as well to know why using a `const` for storing an anonymous function could be considered better.

components/atom/panel/src/index.scss
```diff
@@ -0,0 +1,62 @@
...
+$vertical-alignments-atom-panel: ('top', 'center', 'bottom') !default;
+
+.sui-atom-panel {
+  &-content {
+    position: relative;
+  }
+
+  @each $type in $bgc-panels {
+    $key: nth($type, 1);
+    $value: nth($type, 2);
```
JoanClaret: This seems an smart solution, but looks a bit weird to me. 
i'm more used to see something like:

```scss
$bgc-panels: (
 'canvas': $c-white,
 'contrast': $c-black,
 'accent': $c-accent
) !default;
 
.sui-atom-panel {
  @each $type, $attr in $bgc-panels {
    &--#{$type} {
      @include bgcolor-overlay-opacity($attr);
    }
  }
}
```

what do you think about that? it's a bit more clear?

PablitoGS: Yes, I like it,  is much cleaner and clear than before.

components/atom/panel/src/ImagePanel.js
midudev: It's a personal preference, I wouldn't like to be a must, just to be considered. :)

First, anonymous functions are hard to read in an Error stack (I think, only in latest Chrome editions seems that they're getting automagically named by the const). And that's because the Function.name is not being set. Normally, the only exception I'm up is with inline functions inside an Array.prototype method.

Also, there's no need to identify the function as a let, const or var as it could work as a function expression. It's not like we need to know that the value is immutable or that we're going to modify it. Nor the hoisting is important as clearly we want these functions to be used in general in the file.

At finally, for me at least, is more readable having a function expression than a const with a function. But, as I said, it's my personal choice! :) Anyway, I would be interested as well to know why using a `const` for storing an anonymous function could be considered better.

davidbarna: The function can be named for stacktrace even declared by const. And the stack only works when code is not minified.

I used to prefer function declarations as you do. But since arrow functions appeared I prefer expressions over declaration. This way I make sure all my functions are declared the same way (for readibility) even when I mixed regular and bound functions.

On the other hand sometimes I wondered if I should by default declared all my functions as arrow funcs, also for readibility...

I was just curious in case you'd change my mind but you didn't ;-)

davidbarna (approved):

nguasch (approved):
👍

JoanClaret (approved):

PR #280, opened 2018-01-10
feat(form/rangeDatepicker): add new prop for customizable icon

davidmartin84:
Add new prop for customizable icon

### DIFF
--- a/components/form/rangeDatepicker/src/index.js
+++ b/components/form/rangeDatepicker/src/index.js
@@ -50,6 +50,7 @@ class FormRangeDatepicker extends Component {
   }
 
   render () {
+    const InputIcon = this.props.icon
     return (
       <div className='sui-FormRangeDatepicker'>
         <div className='sui-FormRangeDatepicker-item'>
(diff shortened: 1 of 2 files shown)


components/form/rangeDatepicker/src/index.js
```diff
@@ -122,7 +123,15 @@ FormRangeDatepicker.propTypes = {
   /**
    * Init date of the selected range
    */
-  startDate: PropTypes.instanceOf(Date)
+  startDate: PropTypes.instanceOf(Date),
+  /**
+   * Icon of select inputs
+   */
+  icon: PropTypes.func
```
davidbarna, line 130: Shouldn't it be "node" instead of func?

davidbarna (approved):

components/form/rangeDatepicker/src/index.js
davidmartin84: I used the same method used in: https://github.com/SUI-Components/sui-components/blob/master/components/tag/selectable/src/index.js#L42

PR #285, opened 2018-01-16
feat(dropdown/user): fix notifications position

JoanClaret:
Fix incorrect notification position, since we removed `flex:1` in `sui-DropdownUserMenu-listText` to avoid an IE bug.


devtoni (approved):

davidbarna (approved):

kikoruiz:
Don't like it 😞 @JoanClaret Just to know, which IE version is the bug being reproduced?

JoanClaret:
IE11 @kikoruiz

PR #286, opened 2018-01-16
feat(image/picker): new image-picker component

rackso:


components/image/picker/src/index.js
```diff
@@ -0,0 +1,141 @@
+import React, { Component } from 'react'
+import PropTypes from 'prop-types'
+import cx from 'classnames'
+
+import ImagePlaceholder from '@schibstedspain/sui-image-placeholder'
+
+class ImagePicker extends Component {
+  constructor (props) {
+    super()
```
davidbarna:
```
super(props)
```

components/image/picker/src/index.js
```diff
@@ -0,0 +1,141 @@
...
+
+    const { images } = props
+    let filteredImages = []
+    let selected = -1
+
+    if (images.length > 0) {
+      filteredImages = images.map(image => {
+        if (typeof image.src !== 'undefined' && typeof image.thumb !== 'undefined') { return image }
```
davidbarna: Why would you map and then filter when you can just filter ?
```
images.filter(img => img.thumb && img.src)
```
Use "truthy values" so you don't validate a 0 or "" or null as a value...

components/image/picker/src/index.js
```diff
@@ -0,0 +1,141 @@
...
+        return false
+      }).filter(image => image !== false)
+
+      if (filteredImages.length > 0) {
+        selected = filteredImages.findIndex(image => image.selected === true)
```
davidbarna: You don't need the equality as I supposed it's either true or false anyways
```
filteredImages.findIndex({selected}=> selected)
````

File: components/image/picker/src/index.js [hunk 4]
```diff
@@ -0,0 +1,141 @@
...
+
+  _getListImages () {
+    const { images, selected } = this.state
+
+    const imageList = images.map((image, index) => {
+      const itemClass = cx('sui-ImagePicker-thumbItem', {
+        'sui-ImagePicker-thumbItem--selected': index === selected
+      })
+      return (
+        <div className={itemClass} key={image.src} onClick={() => this._handleClick(image)}>
```
davidbarna: if you wrap with `ul`, use `li`, not `div`

components/image/picker/src/index.js
```diff
@@ -0,0 +1,141 @@
...
+    const placeHolder = {
+      alt: 'Placeholder',
+      src: emptyImage
+    }
+
+    const src = (images.length === 0) ? emptyImage : images[selected].src
+    const alt = (images.length === 0) ? 'No text available' : images[selected].alt
+
+    return (
+      <div className='sui-ImagePicker-image--vcenter'>
```
davidbarna: Can't you reduce the markup with flex or other appraoches ?
Actually it's more a general comment about using less dom nodes in this component.

components/image/picker/src/index.scss
```diff
@@ -0,0 +1,63 @@
+@import '~@schibstedspain/sui-theme/lib/index';
+
+.sui-ImagePicker {
+
+  &-headImage {
+    background-color: color-variation($c-black, 4);
```
davidbarna: Can't you use any of the greys ?
https://github.com/SUI-Components/sui-theme/blob/master/src/settings/_color.scss#L29

components/image/picker/src/index.js
rackso: Really good point!!

components/image/picker/src/index.js
rackso: No. It can even not exist. But I'm taking some idea from your line.

components/image/picker/src/index.js
rackso: Good point too. Legacy ideas.

components/image/picker/src/index.scss
rackso: Sure, and it's open to any modification. Colors are not important at this stage.

components/image/picker/src/index.js
davidbarna: To work with undefined:
```
filteredImages.findIndex({selected}=> !!selected)
```

components/image/picker/src/index.js
rackso: Reducing

components/image/picker/src/index.js
```diff
@@ -0,0 +1,134 @@
...
+
+class ImagePicker extends Component {
+  constructor (props) {
+    super(props)
+
+    const { images } = props
+    let filteredImages = []
+    let selected = -1
+
+    if (images.length > 0) {
```
davidbarna: I think this condition is unnecessary

components/image/picker/src/index.js
```diff
@@ -0,0 +1,134 @@
...
+      selected,
+      images: filteredImages
+    }
+  }
+
+  _getHeadImage () {
+    const { images, selected } = this.state
+    const { emptyImage } = this.props
+
+    const placeHolder = {
```
davidbarna: If it's really a constant, IMO you could declare it ONCE at the top of the file.
It's a tradeoff between performance and memory. Up to you.

components/image/picker/src/index.js
```diff
@@ -0,0 +1,134 @@
...
+      alt: 'Placeholder',
+      src: emptyImage
+    }
+
+    const src = (images.length === 0) ? emptyImage : images[selected].src
+    const alt = (images.length === 0) ? 'No text available' : images[selected].alt
```
davidbarna: If the image exists and it has no alt, then the alt text (indexed) would actually be 'No text available'
Do you really want that.
You could also declare the shape of images to force alt texts as an accessibility requirement.

components/image/picker/src/index.js
```diff
@@ -0,0 +1,134 @@
...
+    const { onClick } = this.props
+
+    const selected = images.findIndex(image => image.src === clickedImage.src)
+    this.setState({ selected })
+    onClick(clickedImage)
+  }
+
+  _buttonAction = (callback) => {
+    const { images, selected } = this.state
+    if (images.length === 0 || selected === -1) { return }
```
davidbarna: According to the code above, if there's no images, `selected` if define as -1 anyways. So, the first condition is unnecessary.

davidbarna (approved):

midudev (approved):

ruben-martin-lozano (approved):

PR #287, opened 2018-01-17
feat(sticky/content): create new sticky content component wrapper.

ronaldo-mh:
New component which adds a sticky or fixed position behavior (depending on the set of props) to the node/element/component that wraps.

You can view the demo in: http://ugliest-alley.surge.sh/workbench/sticky/content/demo


demo/sticky/content/playground
```diff
@@ -0,0 +1,39 @@
...
+  </div>
+)
+
+return (
+  <div style={{ height: '2000px', width: '100%', backgroundColor: 'white'}}>
+    { emptyLines(10) }
+
+    <StickyContent
+      sticky
+      scrollableElement={document.body}
```
davidbarna: Can you put an example with a node that is not body but another element ??
Just to check the behavior when it's not the body.

components/sticky/content/src/index.js
```diff
@@ -0,0 +1,114 @@
...
+  }
+
+  _isStuck = false
+  _elementOffset = null
+  _DOMElement = null
+  _scrollableElement = null
+
+  _shouldStickContent = () => {
+    const windowTop = this._scrollableElement.scrollTop
+    const { top: elementTop } = this._DOMElement.getBoundingClientRect()
```
davidbarna: You may want to optimize this by memoization or debbouncing
http://dcousineau.com/blog/2013/09/03/high-performance-js-tip/

components/sticky/content/src/index.js
```diff
@@ -0,0 +1,114 @@
...
+    if (!this._elementOffset) {
+      this._elementOffset = elementTop
+    }
+    return windowTop >= this._elementOffset
+  }
+
+  _handleScroll = () => {
+    const { onDisplayChange } = this.props
+
+    if (!this._isStuck && this._shouldStickContent()) {
```
davidbarna: memoize the result of this._shouldStickContent() to avoid double calculation when not necesary

components/sticky/content/src/index.js
```diff
@@ -0,0 +1,114 @@
...
+const NO_OP = () => {}
+
+export default class StickyContent extends Component {
+  static displayName = 'StickyContent'
+
+  static propTypes = {
+    /**
+     * Component's children node.
+     */
+    children: PropTypes.node,
```
betobe: You could use the ".isRequired" attribute, right?

components/sticky/content/src/index.js
```diff
@@ -0,0 +1,114 @@
...
+      onDisplayChange && onDisplayChange({ isStuck: false })
+    }
+  }
+
+  componentDidMount () {
+    const { sticky, scrollableElement = document.documentElement } = this.props
+
+    if (sticky) {
+      this._scrollableElement = scrollableElement
+      window.addEventListener('scroll', this._handleScroll)
```
davidbarna: scroll is an event that is dispatched many many times.
You should optimize it with passive mode:
https://blog.chromium.org/2016/05/new-apis-to-help-developers-improve.html
Chrome may applied it by default but we don't know about other browsers like safari mobile.

components/sticky/content/src/index.js
```diff
@@ -0,0 +1,114 @@
...
+     */
+    onDisplayChange: PropTypes.func
+  }
+
+  static defaultProps = {
+    position: Object.keys(STICKY_CONTENT_POSITION_CLASSNAME)[0],
+    onDisplayChange: NO_OP
+  }
+
+  _isStuck = false
```
betobe: Wouldn't _isFixed be a better name?

components/sticky/content/src/index.js
```diff
@@ -0,0 +1,114 @@
...
+    /**
+     * Modifier: indicates where the component will be fixed ('top' or 'bottom').
+     */
+    position: PropTypes.oneOf(Object.keys(STICKY_CONTENT_POSITION_CLASSNAME)),
+    /**
+     * Modifier: full width (100%).
+     */
+    fullWidth: PropTypes.bool,
+    /**
+     * Element which will be used as a references to calculate how many distance its content has been scrolled vertically.
```
betobe: I would explain here what you said to me about body and html nodes.

components/sticky/content/src/index.js
```diff
@@ -0,0 +1,114 @@
...
+
+    if (this._isStuck && !this._shouldStickContent()) {
+      this._isStuck = false
+      this._DOMElement.classList.remove('sui-StickyContent-fixed')
+      onDisplayChange && onDisplayChange({ isStuck: false })
+    }
+  }
+
+  componentDidMount () {
+    const { sticky, scrollableElement = document.documentElement } = this.props
```
betobe: Wouldn't it be better to add document.documentElement as default value using defaultProps?

components/sticky/content/src/index.js
```diff
@@ -0,0 +1,114 @@
...
+    return windowTop >= this._elementOffset
+  }
+
+  _handleScroll = () => {
+    const { onDisplayChange } = this.props
+
+    if (!this._isStuck && this._shouldStickContent()) {
+      this._isStuck = true
+      this._DOMElement.classList.add('sui-StickyContent-fixed')
+      onDisplayChange && onDisplayChange({ isStuck: true })
```
betobe: The part `onDisplayChange && ` will always be true because it has a default value, right? So it could be removed. Same on line 78

components/sticky/content/src/index.js
ronaldo-mh: Yep, I could move it there, but not sure if defining this in defaultProps affects SSR.

components/sticky/content/src/index.js
ronaldo-mh: Didn't know that `getBoundingClientRect()` made so many calculations. I will call it only once when the component ref is saved to avoid so many calls to this method on every scoll method.
Thanks!

components/sticky/content/src/index.js
ronaldo-mh: From what I read, scroll event is by default passive,  but anyway I will add the proper `EventListenerOptions` in case other browser requires them.

davidbarna: I knew it was [default from Chrome 56](https://developers.google.com/web/updates/2017/01/scrolling-intervention) (which was polemic as a breaking change) but I don't think so for rest of browsers. For instance mobile browsers are a concern.

demo/sticky/content/playground
ronaldo-mh: For the moment, the implementation don't support listening to scroll in any custom element.
This `scrollableElementSelector` property refers only to the DOM element to inspect in order to calculate the total distance scrolled in page (by getting its scrollTop value).

components/sticky/content/src/index.js
```diff
@@ -0,0 +1,127 @@
...
+      window.addEventListener('scroll', this._handleScroll, {
+        capture: true,
+        passive: true
+      })
+    }
+  }
+
+  componentWillUnmount () {
+    if (this.props.sticky) {
+      window.removeEventListener('scroll', this._handleScroll, {
```
davidbarna: You know unregister must have the same options for it to remove the actual handler.
That's why I would have the options as a constant.

components/sticky/content/src/index.js
```diff
@@ -0,0 +1,127 @@
...
+    this._isFixed = !this._isFixed
+    this._DOMElement.classList.toggle('sui-StickyContent-fixed')
+    this.props.onDisplayChange({ isFixed: true })
+  }
+
+  _handleScroll = (e) => {
+    e.stopPropagation()
+    const shouldStickContent = this._shouldStickContent()
+
+    if ((this._isFixed && !shouldStickContent) || (!this._isFixed && shouldStickContent)) {
```
davidbarna: A single comparaison can be enough here :-)
```js
this._isFixed !== shouldStickContent && this._toggleFixedStatus()
```

components/sticky/content/src/index.js
```diff
@@ -0,0 +1,127 @@
...
+    this._DOMElement = ref
+  }
+
+  componentDidMount () {
+    const { sticky, scrollableElementSelector } = this.props
+
+    if (sticky) {
+      this._scrollableElement = scrollableElementSelector ? document.querySelector(scrollableElementSelector) : document.documentElement
+
+      window.addEventListener('scroll', this._handleScroll, {
```
davidbarna: FYI there're ways to control better the calls to handleScroll

https://lodash.com/docs/#throttle
https://lodash.com/docs/4.17.4#debounce

davidbarna (approved):

components/sticky/content/src/index.js
ronaldo-mh: Done!

components/sticky/content/src/index.js
```diff
@@ -0,0 +1,130 @@
...
+     */
+    scrollableElementSelector: PropTypes.string,
+    /**
+     * Callback to execute when sticky=true and the wrapper position style attribute changes between "relative" and "fixed".
+     */
+    onDisplayChange: PropTypes.func
+  }
+
+  static defaultProps = {
+    position: Object.keys(STICKY_CONTENT_POSITION_CLASSNAME)[0],
```
betobe: I would use STICKY_CONTENT_POSITION_CLASSNAME.top directly. Changing the order or adding a new sticky classname could change the default behavior without notice.

components/sticky/content/src/index.js
```diff
@@ -0,0 +1,130 @@
...
+      this._setElementTop()
+    }
+
+    return windowTop >= this._elementTop
+  }
+
+  _toggleFixedStatus = () => {
+    this._isFixed = !this._isFixed
+    this._DOMElement.classList.toggle('sui-StickyContent-fixed')
+    this.props.onDisplayChange({ isFixed: true })
```
betobe: Shouldn't it be `this.props.onDisplayChange({ isFixed: this._isFixed })`?

betobe (approved):

PR #289, opened 2018-01-19
feat(error/appBoundary): make error appBoundary more resilient and prepared

midudev:

- [x] Add tolerance error ratio in order to avoid infinite loops for errors on `render()`s methods.
- [x] Add the possibility to redirect to a different page if you have an error.
- [x] Add comments for propTypes.

### DIFF
--- a/components/error/appBoundary/src/index.js
+++ b/components/error/appBoundary/src/index.js
@@ -3,10 +3,24 @@ import PropTypes from 'prop-types'
 
 class ErrorAppBoundary extends Component {
   AlertBasicComponent = null
-  state = { hasError: false }
+  state = { errorCount: 0, hasError: false }
   componentDidCatch (errorMessage, errorStack) {
-    this.props.onError({ errorMessage, errorStack })
-    // import async the alert basic component
+    const { numberOfToleratedErrors, onError, redirectUrlOnUntolerated } = this.props
+    const { errorCount } = this.state
+
+    onError({ errorMessage, errorStack })
+    this.setState({ errorCount: errorCount + 1 })
+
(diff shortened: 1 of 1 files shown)


davidbarna (approved):

components/error/appBoundary/src/index.js
```diff
@@ -3,10 +3,24 @@ import PropTypes from 'prop-types'
...
-    this.props.onError({ errorMessage, errorStack })
-    // import async the alert basic component
+    const { errorThreshold, onError, redirectUrlOnBreakingThreshold } = this.props
+    const { errorCount } = this.state
+
+    onError({ errorMessage, errorStack })
+    this.setState({ errorCount: errorCount + 1 })
+
+    return errorCount >= errorThreshold
+      ? redirectUrlOnBreakingThreshold && (window.location.href = redirectUrlOnBreakingThreshold)
```
ruben-martin-lozano, line 15: Does the "componentDidCatch" occurs only in client side?

ruben-martin-lozano: If so, then it's ok, but if dont', this window.location.href would fail in server side

midudev: componentDidCatch doesn't catch errors for server side rendering.

https://reactjs.org/docs/error-boundaries.html#introducing-error-boundaries

ruben-martin-lozano: great! thx :)

ruben-martin-lozano (approved):

components/error/appBoundary/src/index.js
midudev: Thanks to you!

PR #290, opened 2018-01-22
feat(collapsible/accordion): added behavior when receive new props

davidmartin84:
added behavior when receive new props

### DIFF
--- a/components/collapsible/accordion/src/index.js
+++ b/components/collapsible/accordion/src/index.js
@@ -14,6 +14,11 @@ class CollapsibleAccordion extends Component {
     this._collapseItems(false, this.props.items.findIndex(item => !item.collapsed))
   }
 
+  componentWillReceiveProps (nextProps) {
+    const id = nextProps.items.findIndex(item => !item.collapsed)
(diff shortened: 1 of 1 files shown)


components/collapsible/accordion/src/index.js
```diff
@@ -14,6 +14,11 @@ class CollapsibleAccordion extends Component {
     this._collapseItems(false, this.props.items.findIndex(item => !item.collapsed))
   }
 
+  componentWillReceiveProps (nextProps) {
+    const id = nextProps.items.findIndex(item => !item.collapsed)
+    this.setState({ openIndex: id === this.state.openIndex ? null : id })
```
davidbarna: this code is repeated on line 26
You should refactor to segregate operations and reuse them.

davidbarna (changes requested):

components/collapsible/accordion/src/index.js
davidmartin84: yes, i'm agree! done.

davidbarna (approved):

PR #292, opened 2018-01-25
SUI Atom Spinner

jcger:
Definition: https://paper.dropbox.com/doc/SUI-Spinner-WIP-IofeUUhDTTzmfm0LK6p5s


components/atom/spinner/src/SUILoader/index.scss
```diff
@@ -0,0 +1,45 @@
...
+    animation: $anim-atom-spinner;
+    background-color: transparent;
+    border-radius: 50%;
+    box-shadow: -$sz-atom-spinner-oval $sz-atom-spinner-oval $c-accent-atom-spinner-loader, 0 $sz-atom-spinner $c-primary-atom-spinner-loader;
+    height: $sz-atom-spinner-oval;
+    left: 50%;
+    margin: 0 auto;
+    position: absolute;
+    top: calc(50% - #{$sz-atom-spinner});
+    width: $sz-atom-spinner-oval;
```
beatrizip: IE don´t loves 'calc'

jcger: We support IE >= 11 and it looks like calc is IE >= 9 https://developer.mozilla.org/es/docs/Web/CSS/calc

Also I've tested it and it seems to be working

components/atom/spinner/README.md
```diff
@@ -0,0 +1,27 @@
...
+```
+
+## Usage
+
+### Basic usage
+```js
+import AtomSpinner, {AtomSpinnerTypes} from '@schibstedspain/sui-atom-spinner'
+
+return (
+  <AtomSpinner on type={AtomSpinnerTypes.FULL}>
```
davidbarna: Can you describe somewhere what the options FULL and SECTION mean ?

components/atom/spinner/src/SUILoader/index.scss
```diff
@@ -0,0 +1,45 @@
+@import '~@schibstedspain/sui-theme/lib/index';
+
+$c-accent-atom-spinner-loader: color-variation($c-accent, 1) !default;
+$c-primary-atom-spinner-loader: color-variation($c-primary, -1) !default;
+$sz-atom-spinner-oval: $sz-icon-s;
+$sz-atom-spinner: $sz-atom-spinner-oval * 2;
+
+@keyframes sui-atom-spinner {
```
davidbarna: For readability, perhaps you could scope the variables here.

```sass
$sz-oval: $sz-atom-spinner-oval;
$c-accent: $c-accent-atom-spinner-loader;

/* etc */
```

components/atom/spinner/src/index.js
```diff
@@ -0,0 +1,74 @@
...
+      this.setState({delayed: false})
+    }, DELAY)
+  }
+
+  componentWillUnmount () {
+    clearTimeout(this.timer)
+  }
+
+  getLayerClassName () {
+    const BASE_CLASS = 'sui-atom-spinner-layer'
```
davidbarna: Why declare this constant here instead of the root of the file, like TYPES ?

components/atom/spinner/src/index.scss
```diff
@@ -0,0 +1,31 @@
...
+  right: 0;
+  top: 0;
+  z-index: $z-atom-spinner;
+}
+
+.sui-atom-spinner {
+  position: relative;
+
+  &-layer {
+    @include spinner-layer;
```
davidbarna: For fixed properties placeholders do the job with less bytes, or 

```sass

&-layer {
   &, &-full {
       @include spinner-layer;
    }
  &-full {
      position: fixed;
   }
}

File: demo/atom/spinner/playground [hunk 6]
```diff
@@ -0,0 +1,79 @@
...
+      this.setState({contentLoaded: true})
+    }, 5000)
+  }
+
+  render () {
+    const {contentLoaded} = this.state
+    return contentLoaded
+      ? <p style={boxStyle}>Content loaded, spinner dissapears</p>
+      : (
+        <AtomSpinner on={!this.state.contentLoaded} delayed>
```
davidbarna: Why the "on" property instead of simple removing the spinner ??

demo/atom/spinner/themes/carfactory.scss
```diff
@@ -0,0 +1,3 @@
+@import '../../../../themes/carfactory';
+@import '../../../../components/atom/spinner/src/index.scss';
```
davidbarna, line 2: Why is SUILoader/index.scss not imported in src/index.scss

If the js is imported implicitly, the scss should be as well IMO

components/atom/spinner/src/index.js
```diff
@@ -0,0 +1,74 @@
...
+      type === TYPES.SECTION ? BASE_CLASS : `${BASE_CLASS}-full`,
+      (!delayed && on) || `${BASE_CLASS}-hide`
+    )
+  }
+
+  render () {
+    const {children, ...props} = this.props
+
+    return (
+      <div className='sui-atom-spinner'>
```
davidbarna: Is it really accurate to wrap the element with a div ?? I think it's pretty intrusive. A div is a block that defaults to width 100%.
If i want to put the spinner to a content that is 300px wide, i would have to create my own container.
To me the spinner is a decorator element... You could even play with befores and afters (I'm wondering)

components/atom/spinner/README.md
jcger: Will add it as prop type description

components/atom/spinner/src/SUILoader/index.scss
jcger: Looks like it's not possible to define vars inside keyframes

demo/atom/spinner/playground
jcger: forgot to remove the prop as not needed

demo/atom/spinner/themes/carfactory.scss
jcger: Would love to have the loader as an optional component, but will integrate it for now

components/atom/spinner/src/index.scss
PablitoGS: It´s true, because there are no arguments

components/atom/spinner/src/SUILoader/index.scss
davidbarna: 😞

components/atom/spinner/src/index.js
PablitoGS: We need a parent relative, with height and with equal to the component, 
with afters or befores I think we would have the same problem, any ideas?

davidbarna: https://codesandbox.io/s/l2v0qm8ynm
It's a 10m proof of concept. But it's at least a beginning of approach to be less intrusive and more adaptative.

PablitoGS: Is a good way, I like it, but talking to UI (is also in the UX definition), the loader spinner can be something else for another vertical. If instead of css they want to put an image, a text or a svg, this will not be very useful as a component, because they will not need the before css

davidbarna: Don't take this PoC as definitive. I'm just giving you some thought of a way to reduce the markup.

components/atom/spinner/src/SUILoader/index.scss
beatrizip: ok :)

beatrizip (approved):

nguasch (approved):
👍

components/atom/spinner/src/index.js
```diff
@@ -0,0 +1,106 @@
...
+AtomSpinner.displayName = 'AtomSpinner'
+
+AtomSpinner.propTypes = {
+  /**
+   * Possible options:
+   * 'FULL': The spinner fits the whole page container
+   * 'SECTION': The spinner fits a specific site component
+   */
+  type: PropTypes.oneOf(Object.values(TYPES)),
+  on: PropTypes.bool,
```
davidbarna: Please describe your props. For instance "on" is not giving me any meaning. I think a "show" or "enable" would be more meaningful, but just an opinion

davidbarna (changes requested):
Please review naming conventions

components/atom/spinner/README.md
```diff
@@ -0,0 +1,27 @@
...
+
+```sh
+$ npm install @schibstedspain/sui-atom-spinner --save
+```
+
+## Usage
+
+### Basic usage
+```js
+import AtomSpinner, {AtomSpinnerTypes} from '@schibstedspain/sui-atom-spinner'
```
midudev: 👏 I like very much the AtomSpinnerTypes. :)

components/atom/spinner/src/index.js
```diff
@@ -0,0 +1,106 @@
...
+
+    return !delayed && on
+  }
+
+  render () {
+    const {loader} = this.props
+    return (
+      this.shouldRenderLoader()
+        ? loader
+        : <noscript />
```
midudev: `null` will do the trick.

jcger: unfortunately it doesn't work with `null` as `ReactDOM.findDOMNode(this)` returns `null` and we are not able to retrieve the parent node, so we cannot set its class, so no spinner is shown

components/atom/spinner/src/SUILoader/index.js
```diff
@@ -0,0 +1,15 @@
+import React from 'react'
+import PropTypes from 'prop-types'
+
+const SUILoader = ({text}) =>
+  <span className='sui-atom-spinner-loader'>
```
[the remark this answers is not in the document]
kikoruiz: In fact should be: `sui-AtomSpinner-loader`

kikoruiz (approved):

PablitoGS (approved):

components/atom/spinner/src/index.js
```diff
@@ -0,0 +1,115 @@
...
+    if (this.props.show === nextProps.show) return
+    if (!this.props.show) return this._addParentClass()
+    if (this.props.show) return this._removeParentClass()
+  }
+
+  shouldRenderLoader () {
+    const {show} = this.props
+    const {delayed} = this.state
+
+    return !delayed && show
```
davidbarna: I'm sorry but I just one last thing.
Why do we even need a show prop ? Does it make sense to set show to true or false instead of simple not rendering the spinner.


I mean, instead of 
````
<AtomSpinner show={this.state.show} type={AtomSpinnerTypes.FULL} />
````


this:
````
this.state.show && <AtomSpinner type={AtomSpinnerTypes.FULL} />
````

I'll approve the PR because I guess there's a reason but I wanna know it :-)

davidbarna (approved):

components/atom/spinner/src/index.js
jcger: No, there is no reason anymore, very well seen

davidbarna (approved):

PR #299, opened 2018-01-30
fix(atom/button): avoid adding type attribute to button tag, since in…

ruben-martin-lozano:
@SUI-Components/developers please review.

By default, HTML5 button tag behaviour is as a "submit", and because of not cleaning the "type" attribute from the props (it can be primary, secondary, ...), it was rendering:

```
<button type="primary">...</button>
```

And this value is not a valid option (in terms of W3C spec, it can be only "button", "submit" or "reset"), si it had a wrong behaviour in browsers (IE basically) where if not a valid option is set, then its behaviour is a simple button (but the rest of browsers consider as a submit).

Maybe we should consider of moving from "type" propType to something different that not collides with native html attributes (buttonType for example).

### DIFF
--- a/components/atom/button/src/index.js
+++ b/components/atom/button/src/index.js
@@ -8,7 +8,7 @@ const SIZES = ['small', 'large']
 const MODIFIERS = ['disabled', 'fullWidth', 'focused', 'negative', 'link']
 const OWN_PROPS = [
   ...TYPES, ...SIZES, ...MODIFIERS,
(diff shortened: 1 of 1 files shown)


aleanez (approved):

danivm (approved):

JoanClaret (approved):
well seen! 👍

jcger (approved):

kikoruiz (approved):

midudev (approved):

PR #301, opened 2018-01-30
feat(layout/breakpointSplit): add className prop to component

davidmartin84:
add className prop to component

### DIFF
--- a/components/layout/breakpointSplit/src/index.js
+++ b/components/layout/breakpointSplit/src/index.js
@@ -47,7 +47,12 @@ class LayoutBreakpointSplit extends Component {
         )}
       />
     )))
-    return (<span className={cx(this.props.className, 'sui-LayoutBreakpointSplit')}>{content}</span>)
+
+    const className = cx(
+      'sui-LayoutBreakpointSplit',
+      this.props.className
+    )
+    return (<span className={className}>{content}</span>)
   }
 }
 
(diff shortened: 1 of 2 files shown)


demo/layout/breakpointSplit/playground
```diff
@@ -1,4 +1,5 @@
 const getStyle = style => Object.assign(style, {width: '33.33%', padding: 20, color: '#fff', float: 'left', minHeight: 400})
+const className = 'sui-LayoutBreakpoint'
```
davidbarna: const className = 'my-custom-class' would be more understandable XD

davidbarna (approved):

components/layout/breakpointSplit/src/index.js
```diff
@@ -47,7 +47,7 @@ class LayoutBreakpointSplit extends Component {
         )}
       />
     )))
-    return (<span className='sui-LayoutBreakpointSplit'>{content}</span>)
+    return (<span className={cx(this.props.className, 'sui-LayoutBreakpointSplit')}>{content}</span>)
```
JoanClaret: i'm more used to see something like:

```js
const className = cx(
  'sui-LayoutBreakpointSplit',
  this.props.className
)
return (
  <span className={className}>{content}</span>
)
```

what do you think? to me looks a bit more clear.

davidmartin84: i'm agree! done!

PR #311, opened 2018-02-07
feat(image/slider): add image counter

betobe:
Demo: http://image-counter-demo.surge.sh/workbench/image/slider/demo

Not ready to merge since the icon needed has not been published to npm registry yet.


components/image/slider/src/index.scss
```diff
@@ -1,9 +1,54 @@
...
+      height: $sz-image-slider-icon;
+      width: $sz-image-slider-icon;
+      margin-right: $m-s;
+    }
+
+    &-text {
+      min-width: $miw-image-slider-counter-text;
+      text-align: right;
+    }
+  }
```
JoanClaret: As far as i know, classnames should be defined like so:

```scss
sui-ImageSlider-counterText
sui-ImageSlider-counterIcon
```

btw, not really sure about how should we define the modifiers.
there is my proposition:

```scss
sui-ImageSlider-counter--bottomLeft
sui-ImageSlider-counter--bottomRight
```

components/image/slider/src/index.scss
```diff
@@ -1,9 +1,54 @@
...
+
+    &--bottom-left {
+      bottom: $m-image-slider-counter;
+      left: $m-image-slider-counter;
+    }
+
+    &-icon {
+      fill: $c-image-slider-icon !important;
+      height: $sz-image-slider-icon;
+      width: $sz-image-slider-icon;
```
JoanClaret: incorrect property sort order

JoanClaret (approved):
apart from two minor comments, and the missing icon,  looks fine to me

components/image/slider/src/index.scss
betobe: 👍

components/image/slider/src/index.scss
betobe: Absolutely. My mistake 👍

ronaldo-mh (approved):

components/image/slider/src/index.js
```diff
@@ -1,45 +1,73 @@
...
-      return image.link ? linkFactory({ href: image.link, target: target, className: '', children: img, key: key }) : img
-    })
-  } else {
-    return []
+  /**
+   * @param {Array} images List given by props.images.
+   * @return {Array} List of img elements.
+   */
+  _getSlides (images, linkFactory) {
+    if (images && images.length) {
```
kikoruiz, line 60: If you perform this check in the line 28 of this component, you can skip this `if else`.

components/image/slider/src/index.js
```diff
@@ -1,45 +1,73 @@
...
+      this.setState(currentSlide)
+      sliderOptions.doAfterSlide && sliderOptions.doAfterSlide(currentSlide)
+    }
+  }
+
+  state = {currentSlide: 0}
+
+  render () {
+    const {images, linkFactory, handleClick, dynamicContent, enableCounter} = this.props
+    const slides = this._getSlides(images, linkFactory)
```
kikoruiz, line 28: This could be: `const slides = images && images.length ? this._getSlides(images, linkFactory) : []`

components/image/slider/src/index.js
kikoruiz: And then take advantage of arrow function syntax ;)

components/image/slider/src/index.js
```diff
@@ -1,45 +1,73 @@
...
+  /**
+   * @param {Array} images List given by props.images.
+   * @return {Array} List of img elements.
+   */
+  _getSlides (images, linkFactory) {
+    if (images && images.length) {
+      return images.map((image, index) => {
+        const key = image.key ? image.key + index : index
+        const img = <img className='sui-ImageSlider-image' key={key} src={image.src} alt={image.alt} />
+        const target = image.target ? image.target : '_blank'
```
kikoruiz: Better save this value `_blank` into a constant.

components/image/slider/src/index.js
```diff
@@ -1,45 +1,73 @@
...
+        return image.link ? linkFactory({ href: image.link, target: target, className: '', children: img, key: key }) : img
```
kikoruiz: Personal taste, but defining objects with more than one property in just one line is toooooo ugly, unreadable...

components/image/slider/src/index.js
```diff
@@ -1,45 +1,73 @@
...
- * @param {Array} images List given by props.images.
- * @return {boolean}
- */
-const hasMoreThanOneImage = (images) => (images && images.length > 1)
+  _buildCounter (totalImages) {
+    const classNames = cx('sui-ImageSlider-counter', `sui-ImageSlider-counter--${this.props.counterPosition}`)
+    const Icon = this.props.counterIcon
+    return <div className={classNames}>
+      <Icon svgClass='sui-ImageSlider-counterIcon' />
+      <span className='sui-ImageSlider-counterText'>{this.state.currentSlide + 1}/{totalImages}</span>
```
kikoruiz: Just one suggestion, but it could be so cool to have here some kind of customizable pattern for this string, for example, having a prop for the counter data named `counterPatternFactory` (with a default value, not mandatory):
```javascript
<ImageSlider counterPatternFactory={
  ({ current, separator = '/', total }) => `${current}${separator}${total}`
} />
```
And then, inside the component:
```javascript
 <span className='sui-ImageSlider-counterText'>
  {counterPatternFactory({
     current: this.state.currentSlide + 1,
     total: totalImages
  })}
</span>
```
So that, separator could be a symbol, text or whatever, and the counter parts order could be also modified ;)

components/image/slider/src/index.js
betobe: Actually I prefer this way because it makes this function returns always an array, which is a simple and clear contract that simplifies the render() code and also avoids writing another if/else or ternary operator whenever _getSlides was called (even when it is only called once  by now). Also, the if/else would not be skipped but written in a different manner in another place, with a ternary operator. I see no profit here.

components/image/slider/src/index.js
betobe: Regarding this comment see https://github.com/SUI-Components/sui-components/pull/311/files#r166914428

components/image/slider/src/index.js
betobe: I do this kind of things just to annoy people 🤣 
👍 I'll change it

components/image/slider/src/index.js
betobe: 👍

components/image/slider/src/index.js
betobe: I love it but should I? I mean, this component is supposed to be sui compliant (definition on WIP) (I could use a troll/lol face here) Should I open that door?

betobe: Aaaaand I did it anyway 👍

PR #312, opened 2018-02-08
feat(molecule/thumbnail): new molecule/thumbnail component

tonipinel:
Thumbnail component using the image-placeholder component to create thumbnails with or without links, different aspect ratios and shapes


davidbarna (approved):

components/molecule/thumbnail/src/index.scss
```diff
@@ -0,0 +1,97 @@
+@import '~@schibstedspain/sui-theme/lib/index';
+
+$sizes-sui-molecule-thumbnail: (
```
JoanClaret: As far as i know, SCSS vars should not have `sui` prefix:

https://github.schibsted.io/scmspain/frontend-all-wiki/wiki/Naming-of-Components%27-Sass-Variables

components/molecule/thumbnail/src/index.scss
```diff
@@ -0,0 +1,97 @@
...
+) !default;
+
+$ratios-sui-molecule-thumbnail: (
+  1: 1,
+  4: 3,
+  16: 9
+) !default;
+
+$bd-outer-sui-molecule-thumbnail: 1px solid color-variation($c-gray, 3) !default;
+$bdc-outer-hover-sui-molecule-thumbnail: $c-text-link !default;
```
JoanClaret: What about using: `$bdc-molecule-thumbnail-outer-hover`

instead of `$bdc-outer-hover-sui-molecule-thumbnail` ?

i think it fits better with our convention

components/molecule/thumbnail/src/index.scss
```diff
@@ -0,0 +1,97 @@
...
+$bd-inner-sui-molecule-thumbnail: 3px solid $c-white !default;
+$bgc-sui-molecule-thumbnail: $c-gray-light !default; 
+$bs-sui-molecule-thumbnail: 0 !default;
+
+a {
```
JoanClaret: we dont' usually style elements. We use to style classes.
can somebody confirm if this has changed?

components/molecule/thumbnail/src/index.scss
davidbarna: Naming should follow the convention, yes

components/molecule/thumbnail/src/index.scss
davidbarna: Agree

components/molecule/thumbnail/src/index.scss
tonipinel: ok

components/molecule/thumbnail/src/index.scss
tonipinel: sure

components/molecule/thumbnail/src/index.scss
tonipinel: Ok. So, i should create an element for each one then?

demo/molecule/thumbnail/playground
```diff
@@ -0,0 +1,392 @@
...
+            }}>
+            <h3>Circled</h3>
+            <MoleculeThumbnail
+              src={imgSrc}
+              alt='Some alt'
+              href='https://someLink'
+              target='_blank'
+              captionText='Show!'
+              placeholder={placeholder}
+              type={moleculeThumbnailShape.CIRCLED}
```
jcger: according to the props, this should be shape instead of type

components/molecule/thumbnail/src/index.js
```diff
@@ -0,0 +1,118 @@
...
+              {this.props.captionText}
+            </figcaption>
+        }
+      </figure>
+    )
+  }
+  render () {
+    return this.props.href
+      ? (
+        <a href={this.props.href} target={this.props.target}>
```
jcger: Please use a [LinkFactory](https://github.com/SUI-Components/sui-components/blob/bd8cd89d9fb67559e627c35a5b3a9adeff804896/components/atom/button/src/index.js#L148) instead of an anchor, we might want to use react router links instead of this

also please add rel noopener like [here](https://github.com/SUI-Components/sui-components/blob/bd8cd89d9fb67559e627c35a5b3a9adeff804896/components/atom/button/src/index.js#L62) in case target = _blank

components/molecule/thumbnail/src/index.scss
```diff
@@ -0,0 +1,101 @@
...
+  16: 9
+) !default;
+
+$bd-outer-molecule-thumbnail: 1px solid color-variation($c-gray, 3) !default;
+$bdc-outer-hover-molecule-thumbnail: $c-text-link !default;
+$bd-inner-molecule-thumbnail: 3px solid $c-white !default;
+$bgc-molecule-thumbnail: $c-gray-light !default; 
+$bs-molecule-thumbnail: 0 !default;
+
+a {
```
jcger: please avoid using element selector, better use classes. You might need to change something here when using the link factory

jcger (changes requested):

davidbarna (changes requested):

components/molecule/thumbnail/package.json
```diff
@@ -0,0 +1,18 @@
+{
+  "name": "@schibstedspain/sui-molecule-thumbnail",
```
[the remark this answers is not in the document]
adriadoz: Which should be?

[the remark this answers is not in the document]
davidbarna: It was hidden by gh parsing, sorry. Comment updated.

components/molecule/thumbnail/src/tablet-mobile.scss
```diff
@@ -0,0 +1,26 @@
...
+
+$sizes-mobile-molecule-thumbnail: (
+  large: 96px,
+  medium: 72px,
+  small: 40px,
+  xsmall:  32px
+) !default;
+
+.sui-MoleculeThumbnail {
+  @include media-breakpoint-down(m) {
```
jcger: this mixin is forbidden for new implementations, please only use `media-breakpoint-up` for now on as decided in this [meeting](https://confluence.schibsted.io/display/ST/Breakpoints+-+Frontend)

jcger (changes requested):

jcger (approved):

components/molecule/thumbnail/src/index.scss
```diff
@@ -0,0 +1,116 @@
...
+
+  .sui-MoleculeThumbnail-link & {
+    &:hover {
+      border-color: $bdc-outer-hover-molecule-thumbnail;
+    }
+  }
+
+  .sui-ImagePlaceholder-image {
+    height: 100%;
+    object-fit: cover;
```
davidbarna, line 48: object-fit is not supported by IE11. Do we have some kind of callback ?
https://caniuse.com/#search=object-fit

davidbarna (approved):

components/molecule/thumbnail/src/index.scss
davidbarna: 👍

PR #314, opened 2018-02-15
feat(script/loader): introduce script loader component

naxhh:
Introduce a generic script loader. Based on the one we are using for Gigya.

The function `waitUntil` is a copy from the one in `optimizelyXExperiment` maybe we can move it somewhere or something :)


components/script/loader/README.md
```diff
@@ -0,0 +1,27 @@
...
+
+### Basic usage
+```js
+import ScriptLoader from '@schibstedspain/sui-script-loader'
+
+return (<ScriptLoader src={src} verifier={() => true} isAsync={false} render={() => 'Ready to render!'} />)
+```
+
+
+> **Find full description and more examples in the [demo page](#).**
```
jcger: You might want to update this line

davidbarna (approved):

jcger (approved):

carlosvillu (approved):

PR #315, opened 2018-02-16
fix(modal/gallery): remove negative header margin for mobile landscap…

betobe:
…e view.

So, the modal header (that includes the close button) was not being displayed in mobile using lansdcape view.

### DIFF
--- a/components/modal/gallery/src/index.scss
+++ b/components/modal/gallery/src/index.scss
@@ -56,12 +56,14 @@ $mt-modal-gallery-header-height-mobile: -($h-modal-basic-icon + ($m-l * 2)) !def
         justify-content: center;
       }
 
-      &-header {
-        margin-top: $mt-modal-gallery-header-height-mobile;
-      }
-
       &-content {
         flex-grow: 0;
+      }
+    }
+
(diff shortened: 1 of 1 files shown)


davidmartin84 (approved):

components/modal/gallery/src/index.scss
```diff
@@ -56,15 +56,17 @@ $mt-modal-gallery-header-height-mobile: -($h-modal-basic-icon + ($m-l * 2)) !def
...
-
       &-content {
         flex-grow: 0;
       }
     }
 
+    @include media-breakpoint-down (xs) {
+      &-header {
+        margin-top: $mt-modal-gallery-header-height-mobile;
+      }
```
JoanClaret, line 67: i've been checking the demo, and i must say that i don't really understand why does we need a negative margin, but there must be some reason!
🔦could you bring some light?

betobe: Header and content are both displayed vertically aligned. And following design specs we wanted the content to be perfectly centered so this negative margin does the trick by moving up everything a bit.

davidbarna (approved):

PR #316, opened 2018-02-19
feat(card/collapsible): add X close button on card's header when its …

ronaldo-mh:
…expanded.

screenshot


JoanClaret:
can you provide an screenshot? 🙏 
i don't really understand what does contain `sui-CardCollapsible-infoItemIcon`
Thank you!

ronaldo-mh:
@JoanClaret I have added the screenshot in README and in this PR description!
The `sui-CardCollapsible-info` is the region where we can add some info labels. `sui-CardCollapsible-infoItem` is each element that we add on the info region. And the new `sui-CardCollapsible-infoItemIcon` is intended to include an icon on the left side of the label info text (recently added in our UX designs, is the phone icon next to the phone number in screenshot).

betobe (commented):
Margin between the X ant top-right doesn't seem right. Shouldn't it be 16px? (I'm just watching the image you attached)

components/card/collapsible/src/index.scss
```diff
@@ -45,25 +46,70 @@
...
     }
   }
 
+  &.is-expanded {
+    &.sui-CardCollapsible--boxShadow {
+      box-shadow: $bxsh-card-collapsible-expanded;
+    }
+  }
+
+  &-expandButton {
```
davidbarna: Don't override internal properties by css. Use one the the provided sizes

davidbarna (changes requested):

components/card/collapsible/src/index.scss
ronaldo-mh: OK, I am removing this class. It was just because in our designs the font-size of the card's button is bigger than the defined in atom button by 2px, but I think we can survive without it :D

ronaldo-mh: @betobe The spacing is correct (16px):
```
&Close {
      display: flex;
      justify-content: flex-end;
      padding: $p-l;
    }
```
It just seems bigger because a copied a sample in the playground which had an extra padding around the collapsible card: `<div style={{backgroundColor:'white', padding: 16}}>`.

davidbarna (approved):

betobe (approved):

PR #317, opened 2018-02-19
fix(map/basic): prevent crash on IE

ruben-martin-lozano:
@SUI-Components/developers please review. This prevents IE to crash

### DIFF
--- a/components/map/basic/src/leaflet/map.js
+++ b/components/map/basic/src/leaflet/map.js
@@ -33,7 +33,10 @@ export default class LeafletMap {
   }
 
   dispatchFirstLoad () {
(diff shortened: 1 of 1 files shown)


davecarter (approved):

danivm (approved):

components/map/basic/src/leaflet/map.js
```diff
@@ -33,7 +33,10 @@ export default class LeafletMap {
   }
 
   dispatchFirstLoad () {
-    this.mapDOM.dispatchEvent(new window.CustomEvent('leaflet_map_loaded', { detail: this.getParamsForRequest() }))
+    this.dispatchCustomEvent({
```
ruben-martin-lozano, line 36: This method alreadu checks if can use window.CustomEvent or initCustomEvent and so on

PR #318, opened 2018-02-19
feat(collapsible/basic): allow animation speed customization in colla…

ronaldo-mh:
…psible basic.

### DIFF
--- a/components/collapsible/basic/src/index.scss
+++ b/components/collapsible/basic/src/index.scss
@@ -23,9 +23,9 @@
       transition: $trs-collapsible-basic-trigger-icon;
 
       &-icon {
-        fill: $c-collapsible-basic-trigger-icon !important;
+        fill: $c-collapsible-basic-trigger-icon !important; // sass-lint:disable-line no-important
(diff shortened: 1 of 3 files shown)


components/collapsible/basic/src/index.scss
```diff
@@ -63,7 +72,8 @@
     }
 
     .sui-CollapsibleBasic-collapsibleContent {
-      max-height: 100%;
+      max-height: 100vh;
```
betobe, line 75: Do you see a visible change when using vh? Just being curious.

betobe (approved):

components/collapsible/basic/src/index.scss
ronaldo-mh: It does not change anything visible at all, but using `100vh` allows that the animations work when using the component in FC page, with `max-height: 100%` the transition does not work.

betobe: 👍

davidmartin84 (approved):

PR #319, opened 2018-02-19
feat(cover/basic): use atom button to create cover basic buttons.

ronaldo-mh:


betobe (approved):

components/cover/basic/src/index.scss
```diff
@@ -71,6 +79,27 @@ $op-cover-basic-gradient-after: .35 !default;
...
+  &-button {
+    background-color: $bgc-cover-basic-button;
+    border-radius: $bdrs-cover-basic-button;
+    margin-right: $m-m;
+
+    &:last-child {
+      margin-right: 0;
+    }
+
+    &Item {
```
davidbarna: you must not style internal properties of a component with css.
Use variables instead.

davidbarna (changes requested):

components/cover/basic/src/index.scss
ronaldo-mh: This is similar to the problem you pointed in PR https://github.com/SUI-Components/sui-components/pull/316
But here, our button is even more different and cannot be displayed with the available options of the atom button (it just don't fit with any of the button types defined in sui).

Anyway, I will remove the class from here, and depending if our UX sees that we must follow the designs strictly, I will have to overwrite the atom button in our corresponding domain NC component.

davidmartin84 (approved):

davidbarna (approved):

PR #326, opened 2018-03-02
fix(atom/panel): update panel from component definition

PablitoGS:


davidbarna (approved):

demo/atom/panel/playground
```diff
@@ -61,14 +61,24 @@ return (
...
+              <div style={{height: '150px', width: '200px', color: '#000', fontSize: '24px', fontWeight: 600, margin: 'auto', padding: '16px'}}>
                 <p>Lorem ipsum dolor sit amet</p>
               </div>
             </AtomPanel>
           </div>
           <span>Dark text on light background</span>
         </div>
+        <div style={flexItem}>
+          <div style={{width: '200px'}}>
+            <AtomPanel src={'https://picsum.photos/250/200'}>
```
jcger: I think it would be better if we don't use random picture here in order to ensure the accessibility in the demo, what about greyscale as you are using white font color? (https://picsum.photos/g/250/200?image=10 for example)

demo/atom/panel/playground
```diff
@@ -61,14 +61,24 @@ return (
...
+              <div style={{height: '150px', width: '200px', color: '#fff', fontSize: '24px', fontWeight: 600, margin: 'auto', padding: '16px'}}>
+                <p>Lorem ipsum dolor sit amet</p>
+              </div>
+            </AtomPanel>
+          </div>
+          <span>Dark text on Image</span>
```
jcger: You are setting white as font color

jcger (approved):

PR #330, opened 2018-03-06
feat(cover/basic): remove object-fit fix, use always background-image…

ronaldo-mh:
… with bg-size cover.

As the validation if 'object-fit' style is supported in current page is not compatible with SSR (because it is checking the `document`), we have decided to remove the use of image and only support a div with its `background-image` with the style `background-size: cover`.

This solution works well in the different supported browsers, and with Client and Server Side Rendering.

### DIFF
--- a/components/cover/basic/src/index.js
+++ b/components/cover/basic/src/index.js
@@ -13,19 +13,18 @@ const CoverBasic = (props) => {
     {'is-clickable': isClickable(props)})
 
   const imageContainerClassNames = cx(
-    {'sui-CoverBasic-gradient': props.gradient},
-    {'sui-CoverBasic-objectFitFix': !isObjectFitSupported()})
+    'sui-CoverBasic-image',
+    {'sui-CoverBasic-gradient': props.gradient})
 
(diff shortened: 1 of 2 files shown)


components/cover/basic/src/index.js
```diff
@@ -13,19 +13,18 @@ const CoverBasic = (props) => {
...
+    'sui-CoverBasic-image',
+    {'sui-CoverBasic-gradient': props.gradient})
 
-  // Sadly, object-fit compatibility support cannot be done with just adding one class.
-  // The following line is the second part of the solution which is an inline style.
-  const ieObjectFitInlineStyleFix =
-    isObjectFitSupported() ? '' : { style: {backgroundImage: 'url(' + props.src + ')'} }
+  const coverBackgroundImageInlineStyle = {
+    style: {
+      backgroundImage: 'url(' + props.src + ')'
```
jcger, line 21: you could use template literals, just a suggestion though

jcger (approved):

kikoruiz (approved):

ronaldo-mh:
Resolved in https://github.com/SUI-Components/sui-components/pull/333

PR #333, opened 2018-03-06
adds background image as a backgroundimage property

davecarter:
## PR Summary
First of all. Sorry for colliding with @frmedina . We have been working to solve the same issue separately. My mistake. **we have agreed to have a look to both PRs and get the one that better fits our needs**

- [x] Refactored to Statefull component to allow SSR.
- [x] Removed image element and use a background-image instead.
- [x] Added a Sass var `$c-cover-basic-bgsize` to setup cover or contain bg size.
- [x] Adds better background images in example 😄 
- [x] Removed references to DOM elements and pass an Style object instead

Please review!

### DIFF
--- a/components/cover/basic/README.md
+++ b/components/cover/basic/README.md
@@ -2,3 +2,30 @@
 ### CoverBasic
 
 It shows an image an also accepts a list of buttons and children.
+
+
+### Usage
+CoverBasic must have an `src` prop and might have a `height` prop too to define container height.
+- If `height` prop is not provided the container height will be set via Sass var `$mah-cover-basic`.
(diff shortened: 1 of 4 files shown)


midudev (approved):

ronaldo-mh (approved):

ronaldo-mh:
As I think you are covering the case I was trying to solve in https://github.com/SUI-Components/sui-components/pull/330, we can just merge this one and close the other one.
Thanks!

davecarter:
Thanks a lot @frmedina 👍 
And sorry for the hassle 😊

PR #335, opened 2018-03-07
feat(card/collapsible): add callback parameter on info item

betobe:


rackso (approved):

danivm (approved):

components/card/collapsible/src/index.js
```diff
@@ -179,7 +179,8 @@ CardCollapsible.propTypes = {
         link: PropTypes.string,
         highlighted: PropTypes.bool,
         target: PropTypes.string,
-        iconLeft: PropTypes.func
+        iconLeft: PropTypes.func,
+        onClick: PropTypes.func
```
kikoruiz, line 183: Is it optional? What happens if you don't set a default value? You checked it?

components/card/collapsible/src/index.js
```diff
@@ -31,9 +31,9 @@ class CardCollapsible extends Component {
    * @param {boolean} itemProps.highlighted Highlight the item (bold style).
    * @return {Element} Item to display on card's header info area.
    */
-  _renderInfoItem ({label, link, highlighted, target = '_blank', iconLeft: IconLeft}, index) {
+  _renderInfoItem ({label, link, highlighted, target = '_blank', iconLeft: IconLeft, onClick}, index) {
     const InfoItemElement = link ? 'a' : 'span'
-    const props = link ? { href: link, target } : {}
+    const props = link ? { href: link, target, onClick } : {}
```
kikoruiz, line 36: Maybe you have to add a default value: `const props = link ? { href: link, target, onClick = () => {} } : {}`

betobe: It works properly even without a default value.

PR #340, opened 2018-03-08
feat(cover/basic): button list must be displayed over gradient bg image.

ronaldo-mh:
When displaying the cover with a gradient, the `sui-CoverBasic-buttonList` container displayed on top-left of the cover is shown below of the gradient.
We are just adding a z-index to display it on top of it (same case as the `sui-CoverBasic-children` container).


betobe (approved):

davecarter (approved):
👍

PR #347, opened 2018-03-14
chore(META): new prefix and scope

jcger:
New component name should be like [at]s-ui/react-[category]-[name]


midudev (dismissed):

carlosvillu (commented):
If we do that. I prefer deprecate this repo. And create a new one only for the compliant components.

jcger:
We will move the compliant components to a new studio

PR #356, opened 2018-03-27
feat(list/image): simplify image gutter customisation, improve gradient…

ronaldo-mh:
…over last item.

- Calculation of the gutter between images was a bit messy and did not work correctly when changing its container's size. So, in this PR we are simplifying the required vars to set up the gutter and the sizes of each image item.

- The last item that should display a gradient with a text label (only when there are more images available than the prop `maxItems` provided) was not styled according to designs, so we are fixing this issue creating the gradient now in the pseudo element after.

### DIFF
--- a/components/list/image/src/index.scss
+++ b/components/list/image/src/index.scss
@@ -1,58 +1,62 @@
 @import '~@schibstedspain/sui-theme/lib/index';
 @import '~@schibstedspain/sui-image-lazy-load/lib/index';
-@import 'settings';
+@import './settings';
 
 .sui-ListImage {
   @include sui-list('custom');
-  @include media-breakpoint-only(xl) {
-    margin: -($m-v-list-image-item-half-gutter);
+  @include media-breakpoint-up(l) {
+    justify-content: flex-start;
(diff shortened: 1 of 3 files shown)


davidbarna (approved):

components/list/image/src/index.scss
```diff
@@ -1,58 +1,62 @@
...
-  @include media-breakpoint-only(xl) {
-    margin: -($m-v-list-image-item-half-gutter);
-  }
-  @include media-breakpoint-only(l) {
-    margin: (-($m-v-list-image-item-half-gutter)) (-($m-h-list-image-item-breakpoint-l-gutter));
-  }
-  @include media-breakpoint-down(m) {
-    justify-content: space-between;
+  @include media-breakpoint-up(l) {
+    justify-content: flex-start;
```
davecarter, line 8: Setting `justify-content` to its default value of `flex-start` is redundant.
I mean... Items inside a Flex container are using a justify-content flex-start alignment by default so you don't have to assign that value.

davecarter (approved):

midudev (approved):

PR #364, opened 2018-04-06
feat(list/image): allow customising more items box text styles.

ronaldo-mh:
Additionally, let last item container grows its height to 100% to avoid that the :after content is not displayed when an image url is broken.

### DIFF
--- a/components/list/image/src/index.scss
+++ b/components/list/image/src/index.scss
@@ -29,6 +29,7 @@
   }
 
   &-lastItemContainer {
+    height: 100%;
     position: relative;
 
     &::after {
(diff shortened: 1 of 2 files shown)


davecarter (approved):

davidbarna (approved):

components/list/image/src/index.scss
```diff
@@ -56,6 +57,10 @@
 
     &Label {
       color: $c-list-image-more-items-box-label;
+      font-size: $fz-list-image-more-items-box-label;
+      font-wieght: $fw-list-image-more-items-box-label;
```
davidbarna: won't work :-)

davidbarna (changes requested):

components/list/image/src/index.scss
ronaldo-mh: XD I will fix it now

jcger (approved):

PR #375, opened 2018-04-13
Add Switch component

varavan:
paper [https://paper.dropbox.com/doc/SUI-Switch-DqiRbahVL1CbbNdYSLljA](https://paper.dropbox.com/doc/SUI-Switch-DqiRbahVL1CbbNdYSLljA)


components/atom/switch/src/index.js
```diff
@@ -0,0 +1,179 @@
...
+
+  toggleSwitch = function () {
+    if (this.state.disabled) {
+      return
+    }
+    if (this.onToggle) {
+      this.onToggle(!this.state.toggle)
+    }
+    this.setState({toggle: !this.state.toggle})
+  }.bind(this)
```
davidbarna: You don't have to do this. You can create the function as an instance one.
https://github.com/tc39/proposal-class-fields#field-declarations

Right now you are creating 2 instances of Function each time.

components/atom/switch/src/index.js
```diff
@@ -0,0 +1,179 @@
...
+        break
+      case ' ':
+        this.toggleSwitch()
+        event.preventDefault()
+        break
+    }
+  }
+
+  toggleSwitch = function () {
+    if (this.state.disabled) {
```
davidbarna: To improve readability, we usually declare props as varaibles on the top of the scope:
```
const {disabled, onToggle,toggle} = this.state
```

components/atom/switch/src/index.js
```diff
@@ -0,0 +1,179 @@
...
+  focusSwitch = function () {
+    this.setState({isFocus: true})
+  }.bind(this)
+
+  focusOutSwitch = function () {
+    this.setState({isFocus: false})
+  }.bind(this)
+
+  render () {
+    let divClass = (this.state.toggle || this.type === TYPES.SELECT) ? cx(BASE_CLASS, 'active') : cx(BASE_CLASS)
```
davidbarna: You really can reduce that code and make it more readable and avoid unnecessary executions.
Check classNames docs

components/atom/switch/src/index.js
```diff
@@ -0,0 +1,179 @@
...
+    divClass = (this.size === SIZES.LARGE) ? cx(divClass, 'large') : divClass
```
davidbarna: Please make use of our class naming convention

components/atom/switch/src/index.scss
```diff
@@ -0,0 +1,116 @@
+@import '~@schibstedspain/sui-theme/lib/index';
+
+$sz-atom_switch-input-height: $sz-icon-l !default;
```
davidbarna: Use our variables naming convention.

components/atom/switch/src/index.scss
```diff
@@ -0,0 +1,116 @@
...
+
+$c-atom-switch-active: #60acd0 !default;
+$c-atom-switch-default: #dcdcdc !default;
+
+.sui-AtomSwitch {
+  display: inline-block;
+  cursor: pointer;
+
+  &.toggle-type {
+    & > div.container {
```
davidbarna: Don't use tag selectors.

components/atom/switch/src/index.scss
```diff
@@ -0,0 +1,116 @@
...
+    }
+  }
+
+  & > .container {
+    display: flex;
+    justify-content: center;
+    height: $sz-atom_switch-input-height;
+    padding-top: ($sz-atom_switch-input-height - $fz-m) / 4;
+
+    & > span.text {
```
davidbarna: You should try to avoid descendant selectors. Moreover, them seem to be unnecessary most of the time.

davidbarna (changes requested):

components/atom/switch/src/index.js
varavan: make sense, thanks

components/atom/switch/src/index.js
```diff
@@ -0,0 +1,179 @@
...
+
+class AtomSwitch extends Component {
+  constructor (props) {
+    super(props)
+
+    const {size, disabled, labelLeft, labelRight, type, name, label, labelOptionalText, onToggle} = props
+    this.onToggle = onToggle
+    this.size = size
+    this.type = type
+    this.state = {
```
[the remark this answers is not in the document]
varavan: do you mean toggle and isFocus ? they are not static I need them to start on this state. Do you have any proposal to initiate it? 

I am really not sure about name and labels, they look like quiete static for me but if they are not part of the satate we would not be able to change labels on runtime. What do you think?

components/atom/switch/src/index.js
varavan: do you mean to do this? https://github.com/SUI-Components/sui-components/blob/master/components/collapsible/basic/src/index.js#L15 

otherwise, I am not sure how to use it on a JS event

davidbarna: I meant this:

````
class AtomSwitch extends Component {
   myMethod = () => {
      /* YOUR CODE */
   }
}

File: components/atom/switch/src/index.js (hunk 8 above)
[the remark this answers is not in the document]
Comment (davidbarna) [reply components/atom/switch/src/index.js]: I mean that labelRight, labelOptionalText, etc are static.
You put in the state only mutable properties

File: components/atom/switch/README.md [hunk 9]
```diff
@@ -0,0 +1,38 @@
+# AtomSwitch
+
+> Description
+
+Switch component [paper definition](https://paper.dropbox.com/doc/SUI-Switch-DqiRbahVL1CbbNdYSLljA).
```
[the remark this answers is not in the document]
jcger: Yep, just add a little description (you could copy from the paper) and remove the link to the paper please

components/atom/switch/src/index.js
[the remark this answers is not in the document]
varavan: cool, thanks!

varavan:
@davidbarna  everything should be ready

components/atom/switch/package.json
```diff
@@ -0,0 +1,18 @@
...
+  "description": "",
+  "main": "lib/index.js",
+  "scripts": {
+    "build": "rm -Rf ./lib && mkdir -p ./lib && npm run build:js && npm run build:styles",
+    "build:js": "../../../node_modules/.bin/babel --presets sui ./src --out-dir ./lib",
+    "build:styles": "../../../node_modules/.bin/cpx './src/**/*.scss' ./lib"
+  },
+  "dependencies": {
+    "@s-ui/component-dependencies": "1",
+    "@s-ui/react-atom-label": "^1.1.0"
```
midudev: Just use the major, as we're folling Compatible Version:

`"@s-ui/react-atom-label": "1"`

components/atom/switch/src/index.js
```diff
@@ -0,0 +1,193 @@
...
+  SELECT: 'select',
+  SINGLE: 'single'
+}
+
+class AtomSwitch extends Component {
+  constructor (props) {
+    super(props)
+
+    const {size, disabled, labelLeft, labelRight, type, name, label, labelOptionalText, onToggle} = props
+    this.onToggle = onToggle
```
midudev: Avoid this. It's fine to just access `this.props` in your component and even deconstruct it, instead of copying every prop in the class as a property.

components/atom/switch/src/index.js
```diff
@@ -0,0 +1,193 @@
...
+        break
+    }
+  }
+
+  toggleSwitch = () => {
+    const {disabled, onToggle, toggle} = this.state
+    if (disabled) {
+      return
+    }
+    (onToggle) && onToggle(!toggle)
```
midudev: I would say to put as a defaultProp an empty method rather than checking if the `onToggle` is a function.

components/atom/switch/src/index.js
```diff
@@ -0,0 +1,193 @@
...
+    const {isFocus} = this.state
+    if (!isFocus) {
+      return
+    }
+
+    if (event.defaultPrevented) {
+      return
+    }
+
+    switch (event.key) {
```
midudev: Even as I don't recommend using switch for this case, it could be written as this:

```
switch (event.key) {
  case 'Enter':
  case ' ':
    this.toggleSwitch()
    event.preventDefault()
    break
}

File: components/atom/switch/src/index.js [hunk 14]
```diff
@@ -0,0 +1,193 @@
...
+        break
+      case ' ':
+        this.toggleSwitch()
+        event.preventDefault()
+        break
+    }
+  }
+
+  toggleSwitch = () => {
+    const {disabled, onToggle, toggle} = this.state
```
midudev: `onToggle` is not in the state, is in the props.

components/atom/switch/src/index.js
varavan: okay, I am gonna refactor to `if` statement

components/atom/switch/src/index.js
```diff
@@ -0,0 +1,181 @@
...
+
+    const {disabled} = props
+    this.state = {
+      toggle: false,
+      isFocus: false,
+      disabled
+    }
+
+    if (typeof window !== 'undefined' && window.addEventListener) {
+      window.addEventListener('keydown', this.keyBindings.bind(this), true)
```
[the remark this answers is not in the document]
varavan: okay, thanks

davidbarna (changes requested):
Some of my comments are only proposals. Others require changes.

components/atom/switch/src/SwitchType/single.js
```diff
@@ -0,0 +1,27 @@
+import React from 'react'
+import cx from 'classnames'
+import {BASE_CLASS, TYPES} from '../index'
+import AtomLabel from '@s-ui/react-atom-label'
+
+export const SingleSwitchTypeRender = (opts, isToggle, isFocus, isDisabled, focusSwitchCallback, blurSwitchCallback, toggleSwitchCallback) => {
```
jcger: please treat this as a React component `({...your props...}) => (<div>....</div>)` and use it as one in your AtomSwitch.render function `<SingleSwitchTypeRender {...this.props} newProp1 newProp2... />`

Also define the propTypes

Same for the `ToggleSwitchTypeRender` fn

components/atom/switch/src/index.js
```diff
@@ -0,0 +1,143 @@
...
+  onToggle: PropTypes.func.isRequired
+}
+
+AtomSwitch.defaultProps = {
+  size: SIZES.DEFAULT,
+  disabled: false,
+  labelLeft: 'Off',
+  labelRight: 'On',
+  type: TYPES.TOGGLE,
+  onToggle: (toggle) => {}
```
jcger: if it's required you should remove this

components/atom/switch/src/index.scss
```diff
@@ -0,0 +1,121 @@
...
+      background-color: $c-white;
+      border-radius: ($h-atom-switch-container - 2) / 2;
+      border: solid 1px $c-white;
+      box-shadow: 2px 0 2px 0 $c-gray-light;
+      height: $h-atom-switch-container - 2;
+      transition: margin 0.4s ease-in-out;
+      width: $h-atom-switch-container - 2;
+
+      &.sui-AtomSwitch--toggle {
+        margin-left: 12px;
```
jcger: please use margin var

components/atom/switch/src/index.scss
```diff
@@ -0,0 +1,121 @@
...
+  }
+
+  &.sui-AtomSwitch--disabled {
+    color: $c-gray;
+  }
+
+  &.sui-AtomSwitch--active {
+    & .sui-AtomSwitch--container {
+      .sui-AtomSwitch--inputContainer {
+        background-color: $c-AtomSwitch-active;
```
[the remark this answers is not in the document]
jcger: I think you'r right, modifiers should have two hyphens

demo/atom/switch/playground
```diff
@@ -0,0 +1,39 @@
+
+const flexInlineContainer = {display: 'flex', justifyContent: 'space-between', padding: '2vh 2vw'}
+const itemInlineFlex = {padding: '10px'}
+
+return (
+    <div>
```
jcger: I know this is not a js file, but please use 2 spaces everywhere

components/atom/switch/src/index.scss
varavan: I created some new vars, as this margin is used to place the circle within the component

components/atom/switch/src/SwitchType/single.js
```diff
@@ -0,0 +1,50 @@
...
+    labelOptionalText,
+    size,
+    type,
+    disabled,
+    isToggle,
+    isFocus,
+    focusSwitchCallback,
+    blurSwitchCallback,
+    toggleSwitchCallback
+  } = opts
```
jcger: you could directly deconstruct in the params

= ({name, label, ...}) =>

components/atom/switch/src/SwitchType/single.js
```diff
@@ -0,0 +1,49 @@
...
+  )
+}
+
+SingleSwitchTypeRender.displayName = 'SingleSwitchTypeRender'
+
+SingleSwitchTypeRender.propTypes = {
+  name: PropTypes.string,
+  label: PropTypes.string,
+  labelOptionalText: PropTypes.string,
+  size: PropTypes.string,
```
davidbarna: Usually we use PropTypes.oneOf when the spec enumerates limited options.
Have you considered it ?

components/atom/switch/src/SwitchType/helpers.js
```diff
@@ -0,0 +1,18 @@
+import {BASE_CLASS, TYPES} from '../index'
+import cx from 'classnames'
+
+export function workClassNames (size, type, classType, isToggle, isFocus, isDisabled) {
+  return cx(
+    BASE_CLASS,
+    prefixClass(classType),
+    prefixClass(size),
+    {
+      'sui-AtomSwitch--active': (isToggle || type === TYPES.SELECT),
```
davidbarna: You could use prefixClass here to (just a comment)

components/atom/switch/src/index.js
```diff
@@ -0,0 +1,149 @@
...
+}
+
+class AtomSwitch extends Component {
+  state = {
+    toggle: false,
+    isFocus: false
+  }
+
+  componentDidMount () {
+    window.addEventListener('keydown', this.keyBindings.bind(this), true)
```
davidbarna: listeners MUST ALWAYS be removed on componentWillUnmount

components/atom/switch/src/index.js
```diff
@@ -0,0 +1,149 @@
...
+  }
+
+  keyBindings = (event) => {
+    const {isFocus} = this.state
+    if (!isFocus) {
```
davidbarna:
```js
 if (!isFocus || event.defaultPrevented) {
  return
}
```

components/atom/switch/src/index.js
```diff
@@ -0,0 +1,149 @@
...
+
+  focusOutSwitch = () => {
+    this.setState({isFocus: false})
+  }
+
+  render () {
+    const {toggle, isFocus} = this.state
+
+    return (this.props.type === TYPES.SINGLE)
+      ? (
```
davidbarna, line 86: Consider creating your bind functions out from the render for performance.
```js
this.focusSwitchIfEnabled = this.executeIfEnabled(this.focusSwitch)
```

components/atom/switch/README.md
```diff
@@ -0,0 +1,37 @@
...
+
+## Usage
+
+### Basic usage
+```js
+import AtomSwitch from '@s-ui/react-atom-switch'
+
+return (
+       <AtomSwitch
+          size=''
```
davidbarna: This example would throw error as type and size are wrong

components/atom/switch/src/index.scss
```diff
@@ -0,0 +1,121 @@
+@import '~@schibstedspain/sui-theme/lib/index';
+
+$h-atom-switch--m: $sz-icon-l !default;
+$w-atom-switch--m: 34px !default;
```
davidbarna: are youtr sure all these values must be hardcoded ?
We have to base components on sui-theme variables as much as possible

components/atom/switch/src/index.scss
```diff
@@ -0,0 +1,121 @@
...
+$m-atom-switch-circle--m: 12px !default;
+$h-atom-switch-container--m: 22px !default;
+
+$h-atom-switch--l: $sz-icon-xl !default;
+$w-atom-switch--l: 49px !default;
+$m-atom-switch-circle--l: 18px !default;
+$h-atom-switch-container--l: 32px !default;
+
+$c-AtomSwitch--active: color-variation($c-primary, 1) !default;
```
davidbarna: wrong naming:
https://github.schibsted.io/scmspain/frontend-all-wiki/wiki/Naming-of-Components%27-Sass-Variables

davidbarna (changes requested):

components/atom/switch/src/index.js
```diff
@@ -0,0 +1,158 @@
...
+    this.executeIfEnabledToggleSwitch = this.executeIfEnabled(this.toggleSwitch)
+    this.executeIfEnabledActivateSwitch = this.executeIfEnabled(this.activateToggle)
+    this.executeIfEnabledDeactivateSwitch = this.executeIfEnabled(this.deactivateToggle)
+  }
+
+  componentDidMount () {
+    window.addEventListener('keydown', this.keyBindings.bind(this), true)
+  }
+
+  componentWillUnmount () {
```
davidbarna: This won't work. When you do bind, you are creating a new instances of Function. You need to provide the same listener instance to actually remove it.

components/atom/switch/src/index.js
```diff
@@ -0,0 +1,160 @@
...
+    super(props)
+
+    this.executeIfEnabledFocusSwitch = this.executeIfEnabled(this.focusSwitch)
+    this.executeIfEnabledFocusOutSwitch = this.executeIfEnabled(this.focusOutSwitch)
+    this.executeIfEnabledToggleSwitch = this.executeIfEnabled(this.toggleSwitch)
+    this.executeIfEnabledActivateSwitch = this.executeIfEnabled(this.activateToggle)
+    this.executeIfEnabledDeactivateSwitch = this.executeIfEnabled(this.deactivateToggle)
+  }
+
+  componentDidMount () {
```
jcger, line 36:
```
  componentDidMount () {
    window.addEventListener('keydown', this.keyBindings, true)
  }

  componentWillUnmount () {
    window.removeEventListener('keydown', this.keyBindings, true)
  }
```

components/atom/switch/src/index.js
```diff
@@ -0,0 +1,160 @@
...
+    window.removeEventListener('keydown', this.keyBindingsInstance, true)
+    this.keyBindingsInstance = null
+  }
+
+  keyBindings = (event) => {
+    const {isFocus} = this.state
+    if (!isFocus || event.defaultPrevented) {
+      return
+    }
+    if (event.key === 'Enter' || event.key === ' ') {
```
jcger: please add `event.key === 'Spacebar'` for older browsers

components/atom/switch/src/index.scss
jcger: I asked the design guys, the numbers are wrong. Keep it with these numbers, I'll take care of it

components/atom/switch/src/index.js
```diff
@@ -0,0 +1,160 @@
...
+      this.toggleSwitch()
+      event.preventDefault()
+    }
+  }
+
+  toggleSwitch = () => {
+    const {toggle} = this.state
+    const {onToggle} = this.props
+    onToggle(!toggle)
+    this.setState({toggle: !toggle})
```
jcger: I think it could be dangerous to call the callback before changing the state. In case you agree `this.setState({toggle: !toggle}, () => onToggle(!toggle))` should do the job

jcger (changes requested):

jcger:
Focus seems to not be showing up in IE11 as expected, could you check it please?

varavan:
Checked on IE 11, focus color was not okay .

components/atom/switch/src/index.scss
jcger: Hi! This has been updated to 40px, I'm not sure if we should use the `$sz-icon-xl` var or create new ones for component widths and heights. What do you think @davidbarna ?

components/atom/switch/src/index.scss
```diff
@@ -0,0 +1,127 @@
+@import '~@schibstedspain/sui-theme/lib/index';
+
+$h-atom-switch--m: $sz-icon-l !default;
+$w-atom-switch--m: 34px !default;
+$m-atom-switch-circle--m: 12px !default;
+$h-atom-switch-container--m: 22px !default;
+
+$h-atom-switch--l: $sz-icon-xl !default;
+$w-atom-switch--l: 49px !default;
```
jcger: This has been updated to 56px

varavan: Cool Thanks

jcger (approved):

davidbarna (approved):

beatrizip (approved):

PR #379, opened 2018-04-26
feat(tab/content): add handler method to execute when active tab is c…

ronaldo-mh:
…hanged.


midudev (approved):

davidbarna (approved):

betobe (approved):
Technically, it is "onTabClick"

PR #392, opened 2018-05-17
fix(molecule/notification): write correct name for the sui-component

midudev:
Use correct name for the package of this component.

Before:
`@s-ui/sui-molecule-notification`

After:
`npm install @s-ui/react-molecule-notification`

### DIFF
--- a/components/molecule/notification/README.md
+++ b/components/molecule/notification/README.md
@@ -5,14 +5,14 @@
 ## Installation
 
 ```sh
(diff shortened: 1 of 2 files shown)


jcger:
maybe you should release a deprecation warning before changing the components name

midudev:
@jcger is already done. :)

image

jcger (approved):

JoanClaret (approved):

PR #397, opened 2018-05-23
feat(script/loader): add timeout callback and custom detection delay

kikoruiz:


naxhh (approved):
Good job taking in care compatibility.

kikoruiz:
Also added one prop naming change and return `null` value by default if no case is accomplished.

siscu (approved):

jcger (approved):

PR #399, opened 2018-05-25
[WIP] Format js

davidbarna:
Format code with new sui-lint rules.
This is used to validate new rules.
Build will fail until new version of sui-lint is published.
The new rules can be reviewed here: https://github.com/SUI-Components/sui/pull/278

### DIFF
--- a/components/ad/smartbanner/src/index.js
+++ b/components/ad/smartbanner/src/index.js
@@ -1,42 +1,66 @@
 import PropTypes from 'prop-types'
-import React, { Component } from 'react'
+import React, {Component} from 'react'
 import IconCloseDefault from '@schibstedspain/sui-svgiconset/lib/X'
 import RatingStar from './rating-star'
 import cx from 'classnames'
 
 class AdSmartbanner extends Component {
-  _handleClick = (event) => {
+  _handleClick = event => {
     const {onClick} = this.props
     event.preventDefault()
     onClick()
   }
 
-  _handleClose = (event) => {
+  _handleClose = event => {
     const {onClose} = this.props
     event.preventDefault()
     onClose()
   }
 
-  render () {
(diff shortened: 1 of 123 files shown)


components/ad/smartbanner/src/index.js
```diff
@@ -1,42 +1,66 @@
...
+    const {
+      imageUrl,
+      title,
+      text,
+      buttonText,
+      staticPosition,
+      icon: IconClose,
+      ratingValue,
+      ratingMax,
+      customRatingIcons
```
jcger, line 30: could we add the trailing comma?

davidbarna: I maintain the already approved prettier config:
https://github.com/SUI-Components/sui/blob/master/packages/sui-lint/.prettierrc.js#L5

But let's see:
we had `only-multiline` https://github.com/SUI-Components/sui/compare/0566e646292bb7e1d351e81e46ee237d43dc5373...5a839108405b5bcaa4f3d337d23de5d2241ac6b9#diff-d1bf7adb61d3977be2d4e6cc1b7f9fabL141

We can change it, but we have to choose between `es5` and `all`
https://prettier.io/docs/en/options.html#trailing-commas

I vote all.
What do you think @carlosvillu @miduga @jcger ?

jcger: all

carlosvillu: I don't know, which are the differences between es6 and ALL

carlosvillu: But +1 for trailing commas !!!

davidbarna: We already decided not to use it with Julian.

davidbarna: Anyways it's late. It's been 6 days and we agreed to merge and repos were reformatted. 😩

PR #400, opened 2018-05-29
Feature/sui pta component

zecafa:


components/form/pta/src/index.scss
```diff
@@ -0,0 +1,29 @@
...
+    width: 100%;
+  }
+
+  &-topbar {
+    width: 100%;
+  }
+
+  &-pta {
+    color: $c-white;
+    padding: 50px;
```
davidbarna: You should use a $p-* var of the theme

components/form/pta/src/index.js
```diff
@@ -0,0 +1,74 @@
...
+
+    contentWindow.postMessage(formSettings, '*')
+  }
+
+  render () {
+    const { props: { formUrl }, sendSettingsToForm } = this
+
+    return (
+      <div>
+        <h1>PTA Form builder</h1>
```
davidbarna: are you really displaying this title in ALL sites ? No translation ?
I thought it was going to be only an iframe.

zecafa: Upss, you are right!

components/form/pta/src/index.js
```diff
@@ -0,0 +1,74 @@
+import React, { Component } from 'react'
+import PropTypes from 'prop-types'
+
+export const BASE_CLASS = 'sui-FormPta'
```
jcger: why do you need to export this?

components/form/pta/src/index.js
```diff
@@ -0,0 +1,74 @@
...
+    const formSettings = {
+      enableDraft,
+      formUrl,
+      exitURLs: {
+        success: redirectOnSuccessUrl,
+        error: redirectOnErrorUrl
+      }
+    }
+
+    contentWindow.postMessage(formSettings, '*')
```
jcger: should the targetOrigin be a prop for security reasons?

components/form/pta/src/index.js
midudev: You're not using this const. You may want to use it to build the classnames in the `render`. Besides you're exporting it for some reason?

zecafa: I don't need to export it ;-)

zecafa: I don't need to export it ;-)

components/form/pta/src/index.js
zecafa: 👍

components/form/pta/src/index.scss
zecafa: 👍

components/form/pta/src/index.js
carlesnunez: As David said. Is it going to be an iframe wrapper? or are you going to add the title too?

carlosvillu (approved):

PR #401, opened 2018-05-29
fix(atom/button): disallow overwrite generic hover effect for button link

midudev:
Overwrite the text-decoration of the hovered button of type link so a generic styling doesn't affect it and change it behaviour as a lot of websites has the `a:hover { text-decoration: underline }`.

Example of wrong behaviour without this style:
image

### DIFF
--- a/components/atom/button/src/index.scss
+++ b/components/atom/button/src/index.scss
@@ -188,5 +188,9 @@ $p-atom-button-large: $p-l !default;
 
   &--link {
     text-decoration: none;
+
+    &:hover {
+      text-decoration: none;
+    }
   }
 }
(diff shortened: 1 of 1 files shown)


ruben-martin-lozano (approved):

components/atom/button/src/index.scss
```diff
@@ -188,5 +188,9 @@ $p-atom-button-large: $p-l !default;
 
   &--link {
     text-decoration: none;
+
+    &:hover {
```
kikoruiz: Without duplicating code:
```
  &--link {
    &,
    &:hover {
      text-decoration: none;
    }
  }
```

JoanClaret (approved):

components/atom/button/src/index.scss
midudev: I did that at the beggining but I though it was a bit confusing. :D

carlesnunez (approved):

PR #402, opened 2018-05-30
Feature/enable target property access link fc

devtoni:


midudev (approved):

naxhh (approved):

components/dropdown/basic/src/index.js
```diff
@@ -175,8 +178,8 @@ DropdownBasic.propTypes = {
 
 DropdownBasic.defaultProps = {
   expandOnMouseOver: false,
-  linkFactory: ({ href, className, children, title }) =>
-    <a href={href} className={className} title={title}>{children}</a>
+  linkFactory: ({ href, className, children, title, target }) =>
+    <a href={href} className={className} target={target} title={title}>{children}</a>
```
kikoruiz, line 182: What about using the `...rest` operator?

devtoni: I think it is better to restrict props we passed through but it would be ok too

kikoruiz:
```
linkFactory: ({ children, ...props }) =>
    <a {...props}>{children}</a>
```

davidmartin84 (approved):

siscu (approved):

components/dropdown/basic/src/index.js
kikoruiz: In this case, we restrict the props we pass as per defined shape of links: https://github.com/SUI-Components/sui-components/blob/21b2a70705f790207b71677d4f66f20aafddc183/components/dropdown/basic/src/index.js#L154

kikoruiz: The thing is to write as less code as we can if it's possible ;)

kikoruiz (approved):

JoanClaret (approved):

davidbarna (approved):

PR #408, opened 2018-06-11
feat(services/markdown): first commit

carlosvillu:
Simplify the way to embed larges texts in the pages.


components/services/markdown/README.md
```diff
@@ -0,0 +1,23 @@
+# ServiceMarkdown
+
+> Description
```
midudev: Missing description.

components/services/markdown/README.md
```diff
@@ -0,0 +1,23 @@
...
+
+<!-- ![](./assets/preview.png) -->
+
+## Installation
+
+```sh
+$ npm install @schibstedspain/sui-services-markdown --save
```
midudev: `@s-ui/react-services-markdown`

components/services/markdown/src/index.js
```diff
@@ -0,0 +1,39 @@
...
+    )
+  }
+}
+
+ServiceMarkdown.displayName = 'ServiceMarkdown'
+
+// Remove these comments if you need
+// ServiceMarkdown.contextTypes = {i18n: PropTypes.object}
+ServiceMarkdown.propTypes = {
+  src: PropTypes.string
```
midudev: Missing comment for the description.

components/services/markdown/src/index.js
```diff
@@ -0,0 +1,39 @@
...
+          dangerouslySetInnerHTML={{__html: html}}
+        />
+      </div>
+    )
+  }
+}
+
+ServiceMarkdown.displayName = 'ServiceMarkdown'
+
+// Remove these comments if you need
```
midudev: Follow the advice. XD

midudev (approved):

components/services/markdown/src/index.js
```diff
@@ -0,0 +1,39 @@
...
+    if (req.status === STATUS_OK) {
+      this.setState({html: marked(req.responseText)})
+    }
+  }
+  render() {
+    const {html} = this.state
+    return (
+      <div className="sui-ServiceMarkdown">
+        <div
+          className="sui-ServiceMarkdown-body"
```
jcger: maybe you could remove the classnames as you're not using them.

components/services/markdown/src/index.js
jcger: this prop should be required

components/services/markdown/README.md
```diff
@@ -0,0 +1,23 @@
...
+
+### Basic usage
+```js
+import ServiceMarkdown from '@schibstedspain/sui-services-markdown'
+
+return (<ServiceMarkdown src="https://mycdn.com/myfile.md"/>)
+```
+
+
+> **Find full description and more examples in the [demo page](#).**
```
jcger: You should fulfil the link to the demo

components/services/markdown/src/index.js
```diff
@@ -0,0 +1,39 @@
+import React, {Component} from 'react'
+import PropTypes from 'prop-types'
+
+const STATUS_OK = 200
+
+class ServiceMarkdown extends Component {
+  state = {html: ''}
+  async componentDidMount() {
+    const marked = await import('marked')
+    const req = new window.XMLHttpRequest()
```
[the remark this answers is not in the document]
midudev: If we use fetch, then we need to polyfill it. I would like to use it as well, for the Fetcher too, but if we do that now we're gonna to inflate our bundles.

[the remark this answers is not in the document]
davidbarna: Get it. Just weird to me to assume that a component will be standalone.
Also, we are creating a global component with a basic feature in sui that is not isomorphic.

components/services/markdown/src/index.js
carlesnunez: How is supposed that we are going to handle with the MD styling on our pages? I mean, for example in MA the static pages have kind of this aspect.

image

Who will be the responsible to define the different styles of the MD format. The own web-app? a wrapper component on the site studio?

components/services/markdown/src/index.js
```diff
@@ -0,0 +1,39 @@
...
+    req.open('GET', this.props.src, false)
```
carlesnunez: If we do an XMLHttpRequest to get the statics we should handle the possibility of a exception throw due to CORS incompatibility, for example (this happened to me testing it on surge).

On the other hand I feel that the MD load is a bit slow in some cases and we are not adding any feedback to the user. Is this a right behavior

components/services/markdown/src/index.js
```diff
@@ -0,0 +1,39 @@
...
+    req.send(null)
+    if (req.status === STATUS_OK) {
+      this.setState({html: marked(req.responseText)})
```
carlesnunez, line 15: seems that use marked as we use here is causing problems on their build form. Maybe we must change it for marked.default

midudev: What version of sui-bundler are you using in order to get that error? There're some differences between @s-ui/bundler 2 and 3 regargind dynamic imports, and it might be the root of the problem.

components/services/markdown/package.json
```diff
@@ -0,0 +1,18 @@
+{
+  "name": "@schibstedspain/sui-services-markdown",
+  "version": "1.0.0",
+  "description": "",
+  "main": "lib/index.js",
+  "scripts": {
+    "build": "rm -Rf ./lib && mkdir -p ./lib && npm run build:js && npm run build:styles",
```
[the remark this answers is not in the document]
midudev: In order to do the change on once, I've let this as it is.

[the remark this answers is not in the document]
davidbarna: https://github.com/SUI-Components/sui/issues/297

davidbarna (approved):

danivm (approved):

PR #415, opened 2018-06-13
Feature/atom tooltip

juanmaguitar:
`AtomTooltip` component according to the specifications provided by @jcger and Christian Jimenez


components/atom/tooltip/src/index.scss
```diff
@@ -0,0 +1,147 @@
...
+
+$fz-atom-tooltip: $fz-m !default;
+$lh-atom-tooltip: $lh-m !default;
+$wlc-atom-tooltip: 4 !default; // -webkit-line-clamp → lines to show
+
+$maw-atom-tooltip: 256px !default;
+$mah-atom-tooltip: 88px !default;
+$c-atom-tooltip: $c-white !default;
+$bgc-atom-tooltip: $c-black !default;
+$bdrs-atom-tooltip: 4px !default;
```
jcger: please create a new [sui-theme](https://github.com/SUI-Components/sui-theme/blob/master/src/settings/_box-style.scss) border radius var for this 4px

Don't worry in case the current border radius should be m sized or not, we can do naming refactor later on

components/atom/tooltip/src/index.scss
```diff
@@ -0,0 +1,149 @@
...
+$w-atom-tooltip-arrow: .8rem !default;
+$h-atom-tooltip-arrow: .4rem !default;
+$c-atom-tooltip-arrow: $bgc-atom-tooltip !default;
+
+// Base class
+.sui-AtomTooltip {
+  display: block;
+  font-size: $fz-atom-tooltip;
+  margin: $m-atom-tooltip;
+  opacity: 0;
```
davidbarna: Are we rendering all instances of tooltip in the DOM even when not displayed ?

components/atom/tooltip/README.md
```diff
@@ -0,0 +1,77 @@
+# AtomTooltip
+
+AtomTooltip is a component that attaches a tooltip over a targeted element. It has some properties like being shown always on the screen changing sides if necessary, delays can be set, options to select text on tooltip and more. This component is based on the [Tooltip component](https://reactstrap.github.io/components/tooltips/) of [reactstrap](https://reactstrap.github.io/) which is built using [popper.js](https://popper.js.org/), so:
+- all the props that are accepted by reacstraps's [`Tooltip`]([Tooltip component](https://reactstrap.github.io/components/tooltips/)) will also be accepted by `AtomTooltip`
+- all the [modifiers](https://popper.js.org/popper-documentation.html#modifiers) that are accepted by [popper.js](https://popper.js.org/) will also be accepted by `AtomTooltip` through the `modifiers` prop
```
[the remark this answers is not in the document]
jcger: I agree, the component shouldn't be able to behave different in any aspect as defined by design, it shouldn't have neither less nor more functionality.

components/atom/tooltip/src/index.scss
```diff
@@ -0,0 +1,149 @@
...
+
+$fz-atom-tooltip: $fz-m !default;
+$lh-atom-tooltip: $lh-m !default;
+$wlc-atom-tooltip: 4 !default; // -webkit-line-clamp → lines to show
+
+$maw-atom-tooltip: 250px !default;
+$mah-atom-tooltip: 100px !default;
+$c-atom-tooltip: $c-white !default;
+$bgc-atom-tooltip: $c-black !default;
+$bdrs-atom-tooltip: 4px !default;
```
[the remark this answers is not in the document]
jcger: https://github.com/SUI-Components/sui-theme/pull/153/files

components/atom/tooltip/src/index.scss
juanmaguitar: The tooltip markup is added and removed from the DOM every time the `isOpen` state change, so yes, the `opacity` values can be removed

components/atom/tooltip/src/index.js
```diff
@@ -0,0 +1,108 @@
...
+  }
+
+  componentWillUnmount() {
+    document.removeEventListener('click', this.handleClickOutsideElement)
+  }
+
+  handleClickOutsideElement = event => {
+    const {isOpen} = this.state
+    if (isOpen) {
+      const tooltipDom = document.querySelector(`.${BASE_CLASS}`)
```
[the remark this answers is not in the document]
juanmaguitar: The thing is this element is added and removed from the DOM every time the value of `isOpen` change, so i don't think caching this value would work

components/atom/tooltip/src/index.js
```diff
@@ -0,0 +1,108 @@
...
+  }
+
+  render() {
+    return (
+      <Tooltip
+        {...this.props}
+        isOpen={this.state.isOpen}
+        toggle={this.toggle}
+        className={BASE_CLASS}
+        innerClassName={`${BASE_CLASS}-inner`}
```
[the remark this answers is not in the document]
juanmaguitar: Right! Fixed

components/atom/tooltip/src/index.js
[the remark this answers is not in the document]
davidbarna: ref means react will store this value when he creates the DOM element. So no search step needed.
Plus, if you identify by class you should search by class too: https://jsperf.com/getelementsbyclassname-vs-queryselectorall/67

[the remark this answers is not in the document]
juanmaguitar: As discussed in person w/ @davidbarna we don't have access to the DOM element on the inside because this component is internally using the `reactstrap` Tooltip component. So we cannot access to this DOM elements via `ref`

There's an issue opened related to this (https://github.com/reactstrap/reactstrap/issues/1082) that maybe someday will give us access to the DOM elements inside and will allow to perform this kind of operations in a more elegant way

Meanwhile there's no other way than selecting the DOM element with some of available methods. In this case, using the `getElementsByClassName` seems to be the most easy and performant way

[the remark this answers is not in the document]
juanmaguitar: As discussed w/ @davidbarna there's no way of using `ref` to access the inner DOM element with the current implementation of `Tooltip`'s reactstrap
I've created this PR (https://github.com/reactstrap/reactstrap/pull/1090) to be able to use `ref` to access the DOM element containing the Tooltip _from the outside_
Meanwhile using `getElementsByClassName` to get this element seems to be the most straightforward way.

juanmaguitar:
Added some changes to the PR related to:
- Improved the DOM selection w/ innerRef handler
- Added IntersectionObserver to hide tooltip when target element is not in the viewport
- Tooltip displayed when target element is selected via keyboard (using tab, keyIndex)
- Hide Contextual Menú on Chrome when longpress
- Updated `reactstrap` to version `6.2.0` which includes some custom changes requested for this `AtomTooltip`
- Added longpress handling to allow coexistence of _click_ (trigger custom event) & _long-click_ (will display Tooltip) on mobile devices

> PR: I'm still working on some use cases related to the _click (trigger custom event) & long-click (will display Tooltip) on mobile devices_ feature

components/atom/tooltip/package.json
```diff
@@ -0,0 +1,20 @@
...
+  "description": "",
+  "main": "lib/index.js",
+  "scripts": {
+    "build": "rm -Rf ./lib && mkdir -p ./lib && npm run build:js && npm run build:styles",
+    "build:js": "../../../node_modules/.bin/babel --presets sui ./src --out-dir ./lib",
+    "build:styles": "../../../node_modules/.bin/cpx './src/**/*.scss' ./lib"
+  },
+  "dependencies": {
+    "@s-ui/component-dependencies": "1",
+    "@researchgate/react-intersection-observer": "0.6.1",
```
midudev: I think this dependency is not used in the component at all.

juanmaguitar: You're right! Removed

components/atom/tooltip/src/hoc/withIntersectionObserver.js
```diff
@@ -0,0 +1,39 @@
+/* global IntersectionObserver */
```
midudev, line 1: This component is soooo interesting. :) We could adapt it as a standalone component in order to be reused in other components. For now, I would say is fine to be here, but let's keep it in mind.

components/atom/tooltip/src/hoc/withIntersectionObserver.js
```diff
@@ -0,0 +1,39 @@
...
+import React, {Component} from 'react'
+import 'intersection-observer' // polyfill
```
midudev: I think this polyfill is not isomorphic, thus this component will break server side rendering. We could do something like this: https://github.com/miduga/react-slidy/blob/master/src/index.js#L7

juanmaguitar: Is there a way I can try the tooltip demo in the playground using server side rendering? Or any other documented way i can check this possible issue?

components/atom/tooltip/src/index.scss
```diff
@@ -0,0 +1,166 @@
...
+      border-width: $h-atom-tooltip-arrow ($w-atom-tooltip-arrow / 2) 0;
+      top: 0;
+    }
+  }
+}
+
+.sui-AtomTooltip--right {
+  padding: 0 $h-atom-tooltip-arrow;
+
+  .arrow {
```
jcger: if you use this, someone would be able to change this styling, we want to avoid it, it has to be encapsulated.

Is arrow a class inside the wrapped component?

components/atom/tooltip/src/index.js
```diff
@@ -0,0 +1,234 @@
...
+    clearTimeout(this.touchTimer)
+    this.touchTimer = setTimeout(() => {
+      if (!this.hasTouchEnded) {
+        this.setState({
+          isOpen: !this.state.isOpen
+        })
+      }
+      this.preventNonTouchEvents = false
+      this.hasTouchEnded = false
+    }, 1000)
```
jcger: should this 1000 be a prop?

components/atom/tooltip/src/index.js
```diff
@@ -0,0 +1,234 @@
...
+
+class AtomTooltip extends Component {
+  state = {isOpen: false}
+  preventNonTouchEvents = false
+  hasTouchEnded = false
+  touchTimer = null
+  refTooltip = React.createRef()
+  refTarget = React.createRef()
+
+  static defaultProps = {
```
jcger: Could you please add this at the end together with the prop types? I'm afraid this gets overseen as we always add the prop types outside the class itself

Why are you setting it to true? Isn't it set by the intersection observer?

components/atom/tooltip/src/index.js
```diff
@@ -0,0 +1,234 @@
...
+  handleStopPropagation = e => {
+    e.preventDefault()
+    e.stopPropagation()
+    return false
+  }
+
+  toggle = e => {
+    const {type} = e
+    if (this.onClickTarget) {
+      if (type === 'touchstart') this.handleTouchStart(e)
```
jcger: maybe you could separate it in descriptive functions/strategies so its easier to understand why you do this

components/atom/tooltip/src/index.js
```diff
@@ -0,0 +1,234 @@
...
+        this.setState({
+          isOpen: !this.state.isOpen
+        })
+      }
+    }
+  }
+
+  render() {
+    const {hideArrow, delay, autohide, placement} = this.props // eslint-disable-line react/prop-types
+    const target = this.refTarget.current
```
jcger: If I don't remember wrong, you aren't able to cache this as it changes? Maybe you could add a comment explaining why

components/atom/tooltip/src/index.js
juanmaguitar: Sure! I'll move it outside the class
If this `isVisible` is `false` or `undefined`, it cancels any other behavior of the tooltip. I've detected some cases where this was provoking an undesired behavior of the tooltip (wasn't working when it should) so setting this initial value fixed this. I can do more thorough tests to validate the removal of this `defaultProps`

components/atom/tooltip/src/index.js
juanmaguitar: Don't think so. This exact value is defined in the specification.
Anyway, it can be modified to be a prop without too much effort

components/atom/tooltip/src/index.js
juanmaguitar: I agree! I'll do that

components/atom/tooltip/src/index.scss
juanmaguitar: Yes! this `.arrow` class definition will only be applied when it is found under a `.sui-AtomTooltip` element. The base component used  (`Tooltip` of [reactstrap](https://github.com/reactstrap/reactstrap)) doesn't allow in its current version to set a custom className for this _arrow_ element.

For now it can be kept in this way (very few chances for this to cause a problem). 

Meanwhile, an [issue](https://github.com/reactstrap/reactstrap/issues/1117) with its related [PR](https://github.com/reactstrap/reactstrap/pull/1119) has been submitted to the source project.

If this change is accepted we can proceed to set a custom className for this element.

juanmaguitar:
Latest version can be tested at: https://sui-components-feature-atom-tooltip.now.sh/workbench/atom/tooltip/demo

jcger (approved):

components/atom/tooltip/src/index.js
juanmaguitar: Added as prop

components/atom/tooltip/src/hoc/withIntersectionObserver.js
juanmaguitar: Added SSR support in this HOC. Tested using this quick-boilerplate-for-SSR → https://github.com/juanmaguitar/quick-ssr-test-sui-components

davidbarna (approved):

PR #416, opened 2018-06-13
feat(atom/image): performance optimization w/ memoization 

juanmaguitar:
Performance optimization after feedback from @davidbarna using memoization for classNames generation
Also prettify format update


components/atom/image/package.json
```diff
@@ -9,7 +9,8 @@
     "build:styles": "../../../node_modules/.bin/cpx './src/**/*.scss' ./lib"
   },
   "dependencies": {
-    "@s-ui/component-dependencies": "1"
+    "@s-ui/component-dependencies": "1",
+    "memoize-one": "^3.1.1"
```
davidbarna: Use exact versions

components/atom/image/src/index.js
```diff
@@ -23,44 +24,42 @@ class AtomImage extends Component {
...
       className,
       `is-${loading ? 'loading' : 'loaded'}`,
       error && `is-error`
     )
-  }
+  })
 
-  get classNamesFigure () {
-    const { placeholder, skeleton } = this.props
+  getClassNamesFigure = memoize((placeholder, skeleton) => {
```
davidbarna: Are you sure memoize is doing less work to check if it should return cache than what the function does ?
https://github.com/alexreardon/memoize-one/blob/master/src/index.js
I'm really not sure if we should add a package (even small) for this purpose.

juanmaguitar: Yes, because it cache the result of `cx` function so when invoked, if parameters hasn't changed it will return the cached value (it will not execute the `cx` function again to get the same result)

The main optimization here (and this is a typical case) is not doing additional processing (execution of `cx` in this case) to get the same results 

This library is less than than 1/2 Kb gzipped and is the one recommended by react in their blog as a good solution to solve this use case → https://reactjs.org/blog/2018/06/07/you-probably-dont-need-derived-state.html#what-about-memoization

davidbarna: I know memoize and what's it for. It's recommended to avoid operations repetition but id doesn't mean it's recommended for all cases. 

Memoize has a cost of comparison. I was just wondering if what does cx is more or less expensive than this comparison. Or if it would be better to check diff on your own. That's with componentWillUpdate is for.

Plus, classNamesFigure depends only on props. You could use willReceiveProps and it could be even more performant IMO. Right now you execute memoize twice for each render, no matters if the render is consequence of props change or state change. If state changes but not props, getClassNamesFigure doesn't even need to be executed (nor memoize).

It's micro performance and we're not strict about it. But as you were doing a PR with a micro performance improvement, I was willing to go in depth. But what you did is valid.

davidbarna (approved):

components/atom/image/package.json
davidbarna: Approved but you still need to change this :-)

jcger (approved):

components/atom/image/src/index.js
juanmaguitar: It seems the methods `componentWillMount`, `componentWillReceiveProps`, and `componentWillUpdate` will be deprecated from next version of React (17.x) as they are considered _unsafe_  (https://diigo.com/0cf0e1) in favor of a new lifecycle method called [`getDerivedStateFromProps`](https://reactjs.org/docs/react-component.html#static-getderivedstatefromprops) [available from version 16.3.x](https://reactjs.org/blog/2018/03/29/react-v-16-3.html) (the current version in Schibsted projects I understand is 16.4.0)

The method `getDerivedStateFromProps` [is invoked right before calling the render method]( https://diigo.com/0cf0vm), both on the initial mount and on subsequent updates of props

So, in this context, they suggest the memoization solution as the way to go when [we need to re-compute data only when a prop changes](https://diigo.com/0cf0v0)

juanmaguitar:
I finally think in this case the current approach is the most flexible one and only in the case there are performance issues, we could apply some of the ideas discussed in this PR
So i'm closing this PR without merging anything

PR #417, opened 2018-06-13
fix(services/markdown): make it compatible with dev and built version

midudev:
Make it work in dev and where the built version of the component is used.

It seems that while Webpack is fine using this import, babel is transforming it to a code where it needs access the property default. Maybe is because how is the third party library made (UMD) but this fixes the problem.

### DIFF
--- a/components/services/markdown/src/index.js
+++ b/components/services/markdown/src/index.js
@@ -7,7 +7,8 @@ class ServiceMarkdown extends Component {
   state = {html: ''}
 
   async componentDidMount() {
-    const marked = await import('marked')
+    const markedLibrary = await import('marked')
+    const marked = markedLibrary.default || markedLibrary
     const req = new window.XMLHttpRequest()
     req.open('GET', this.props.src, false)
     req.send(null)


davidbarna (approved):

components/services/markdown/src/index.js
```diff
@@ -7,7 +7,8 @@ class ServiceMarkdown extends Component {
   state = {html: ''}
 
   async componentDidMount() {
-    const marked = await import('marked')
+    const markedLibrary = await import('marked')
+    const marked = markedLibrary.default || markedLibrary
```
davidbarna, line 11: Are you sure it's not related to https://github.com/SUI-Components/sui/pull/298/files ?
Maybe it has to do with the babel preset ?

danivm (approved):

components/services/markdown/src/index.js
davidbarna: No. We tried with phoenix with both my computer and @devtoni 's and then we edited the node_modules and it worked. And it didn't happen with sui-studio@4. Only 5.

PR #419, opened 2018-06-15
Fix/make the atom switch great again

midudev:
- [x] Add the possibility of initialize the component with a value.
- [x] Stop using global event listener and use React synthetic event instead.
- [x] Avoid any interaction when the component is disabled.
- [x] Fix circle animation to stop before the end of the width.
- [x] Simplify the callbacks.
- [x] Fix weird jump when toggling.
- [x] Improve playground.
- [x] Refactor to use same props for the two different component types.

### DIFF
--- a/components/atom/switch/src/index.js
+++ b/components/atom/switch/src/index.js
@@ -114,11 +114,11 @@ AtomSwitch.propTypes = {
 }
 
 AtomSwitch.defaultProps = {
-  initialValue: false,
-  size: SIZES.DEFAULT,
   disabled: false,
+  initialValue: false,
   labelLeft: 'Off',
   labelRight: 'On',
+  size: SIZES.DEFAULT,
   type: TYPES.TOGGLE
 }
 
--- a/components/atom/switch/src/index.js
+++ b/components/atom/switch/src/index.js
@@ -16,101 +16,54 @@ export const TYPES = {
   SINGLE: 'single'
 }
 
+const SUPPORTED_KEYS = [' ', 'Enter', 'Spacebar']
+
 class AtomSwitch extends Component {
   state = {
-    toggle: false,
+    isToggle: this.props.initialValue,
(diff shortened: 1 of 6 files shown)


rackso (approved):

components/atom/switch/src/index.js
```diff
@@ -16,108 +16,65 @@ export const TYPES = {
...
-    const {toggle} = this.state
+  _onToggle = forceValue => {
+    const {isToggle: stateToggle} = this.state
     const {onToggle} = this.props
-    this.setState({toggle: !toggle}, () => onToggle(!toggle))
-  }
-
-  activateToggle = () => {
-    this.setState({toggle: true})
+    const isToggle = forceValue !== undefined ? forceValue : !stateToggle
```
danivm, line 39: why not `const isToggle = forceValue || !stateToggle`?

midudev: Because `forceValue` could be `false` or `true`. If I use your code, when I try to force a value of `false` it will use always the `!stateToggle`

components/atom/switch/src/index.js
```diff
@@ -16,108 +16,65 @@ export const TYPES = {
   SINGLE: 'single'
 }
 
+const SUPPORTED_KEYS = [' ', 'Enter', 'Spacebar']
+
 class AtomSwitch extends Component {
   state = {
-    toggle: false,
+    isToggle: this.props.initialValue,
```
danivm, line 23: `isToggle` or `isToggled` ?

danivm (approved):

components/atom/switch/src/index.js
midudev: From my perspective, `isToggled` is the correct one. But I followed the `isFocus` structure. I felt I was changing too much. :D

jcger (approved):

PR #422, opened 2018-06-19
[services/markdown] Use async Ajax request

carlosvillu:


carlesnunez (approved):

components/services/markdown/src/index.js
```diff
@@ -13,11 +14,15 @@ class ServiceMarkdown extends Component {
         const markedLibrary = require('marked')
         const marked = markedLibrary.default || markedLibrary
         const req = new window.XMLHttpRequest()
-        req.open('GET', this.props.src, false)
-        req.send(null)
-        if (req.status === STATUS_OK) {
-          this.setState({html: marked(req.responseText)})
+        req.open('GET', this.props.src, true)
+        req.onload = () => {
```
midudev, line 18: If you're using `onload` I think you don't need to check the `readyState`. If you use `onreadystatechange` you could check the `readyState` then.

midudev (approved):

components/services/markdown/src/index.js
midudev: Feeling ignored. 🤣

carlesnunez: I thought about the same but I didn't comment it 😆 

you are not alone @miduga

PR #427, opened 2018-06-21
feat(services/markdown): add scroll to element

danivm:
Add scroll to an element after markdown is rendered.

[SCMI-21452](https://jira.scmspain.com/browse/SCMI-21452)

### DIFF
--- a/components/services/markdown/src/index.js
+++ b/components/services/markdown/src/index.js
@@ -30,8 +30,10 @@ class ServiceMarkdown extends Component {
 
   componentDidUpdate() {
     const id = document.location.hash.substring(1)
-    const element = document.getElementById(id)
-    element && element.scrollIntoView({block: 'start', behavior: 'smooth'})
+    if (id) {
+      const element = document.getElementById(id)
+      element && element.scrollIntoView({block: 'start', behavior: 'smooth'})
+    }
   }
 
   render() {
--- a/components/services/markdown/src/index.js
+++ b/components/services/markdown/src/index.js
@@ -5,7 +5,7 @@ const STATUS_OK = 200
 const COMPLETED = 4
 
 class ServiceMarkdown extends Component {
-  state = {html: ''}
+  state = {html: '', loaded: false}
 
   async componentDidMount() {
     require.ensure(
@@ -18,7 +18,7 @@ class ServiceMarkdown extends Component {
         req.onload = () => {
           if (req.readyState === COMPLETED) {
             if (req.status === STATUS_OK) {
-              this.setState({html: marked(req.responseText)})
+              this.setState({html: marked(req.responseText), loaded: true})
             }
           }
         }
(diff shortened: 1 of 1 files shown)


davecarter (approved):

midudev (approved):

components/services/markdown/src/index.js
```diff
@@ -28,6 +28,14 @@ class ServiceMarkdown extends Component {
     )
   }
 
+  componentDidUpdate() {
+    const id = document.location.hash.substring(1)
```
jcger, line 32: I think this makes the component less reusable, "id" should be a prop passed by to the dumb UI component and the logic behind loading this id should be done in the wrapping component from each vertical

jcger (changes requested):

jcger (approved):

components/services/markdown/src/index.js
```diff
@@ -28,6 +28,14 @@ class ServiceMarkdown extends Component {
...
+    if (id) {
+      const element = document.getElementById(id)
+      element && element.scrollIntoView({block: 'start', behavior: 'smooth'})
```
davidbarna, line 35: FYI: smooth not supported by IE11

components/services/markdown/src/index.js
```diff
@@ -28,6 +28,14 @@ class ServiceMarkdown extends Component {
     )
   }
 
+  componentDidUpdate() {
```
davidbarna, line 31: componentDidUpdate will be called every time the parent component is rendered, even if your props are unchanged. So, you will rescroll over and over on each update. You must control that.
The scroll should be done only html changes.
So you can fix this with a shouldComponentUpdate.

components/services/markdown/src/index.js
midudev: We've been talking about this and we're keeping as it is to keep the native spec of HTML with hash and id.

davidbarna: We are generating HTML anyways in a web environment in a cycle method that only executes in browser. So it's ok to me to do it here.

components/services/markdown/src/index.js
midudev: I think he's aware about that. The good thing is that the same code works in IE11 but without behaviour.

components/services/markdown/src/index.js
```diff
@@ -28,6 +28,18 @@ class ServiceMarkdown extends Component {
...
+    const id = document.location.hash.substring(1)
+    if (id) {
+      const element = document.getElementById(id)
+      element && element.scrollIntoView({block: 'start', behavior: 'smooth'})
+    }
+  }
+
+  shouldComponentUpdate(nextProps, nextState) {
+    return nextState.html !== this.state.html
```
midudev: I think we could simplify this kind of comparison. Instead comparing the huge html, creating a new state and using it in the shouldComponentUpdate to only do the action once.

carlosvillu (approved):

PR #431, opened 2018-06-25
fix(atom/tag): added right and bottom margin, removed left margin

beatrizip:
In document definition we need 8px of margin right and 8px of margin bottom.


kikoruiz (approved):

components/atom/tag/src/index.scss
```diff
@@ -27,13 +27,10 @@ $p-atom-tag: 0 $p-l !default;
   font-size: $fz-s;
   height: $h-atom-tag;
   line-height: $h-atom-tag;
+  margin: 0 $m-m $m-m 0;
```
PablitoGS, line 30: if don't break line, we would not need margin bottom.

PablitoGS (approved):

PR #432, opened 2018-06-25
fix(atom/tag): added line heigth

beatrizip:
The icon took the height of his father's line inside his own container


PablitoGS (approved):

davidbarna (approved):

davidbarna:
It's ok. But check is there is a line-height defined by design that should be assigned by varaible.

PR #433, opened 2018-06-27
chore(META): use the correct script for generating new components

midudev:
Fix the generate script in order to use the new one category and prefix.

### DIFF
--- a/package.json
+++ b/package.json
@@ -9,7 +9,7 @@
     "co": "sui-studio commit",
     "commitmsg": "validate-commit-msg",
     "deploy": "sui-deploy spa sui-components --now -p",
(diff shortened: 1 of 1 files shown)


jcger:
https://github.com/SUI-Components/sui-components/pull/347 🤷‍♂️

package.json
```diff
@@ -9,7 +9,7 @@
     "co": "sui-studio commit",
     "commitmsg": "validate-commit-msg",
     "deploy": "sui-deploy spa sui-components --now -p",
-    "generate": "sui-studio generate -S schibstedspain",
```
davidbarna, line 12: The convention of "@s-ui" was decided only for compliant components. I would not merge this yet.

PR #435, opened 2018-06-27
fix(abtesting/optimizelyXExperiment): fix: wait for sdk of optimizely…

davidbarna:
… to be loaded.

When optimizely v1 and v2 are loaded in a page,
window.optimizely can be an array for a moment.
We have to wait for the actual sdk

### DIFF
--- a/components/abtesting/optimizelyXExperiment/src/optimizely-x.js
+++ b/components/abtesting/optimizelyXExperiment/src/optimizely-x.js
@@ -19,7 +19,8 @@ const waitUntil = (truthyFn, callback, delay = 100, interval = 100) => {
   }, interval)
 }
 
-const getOptmizely = () => window && window.optimizely
+const getOptmizely = () =>
+  window && window.optimizely && window.optimizely.get && window.optimizely
 
 let optimizelyPromise
 


siscu (approved):

kikoruiz (approved):

davidbarna:
Build is failing temporarily but tests were executed locally.

components/abtesting/optimizelyXExperiment/src/optimizely-x.js
```diff
@@ -19,7 +19,8 @@ const waitUntil = (truthyFn, callback, delay = 100, interval = 100) => {
   }, interval)
 }
 
-const getOptmizely = () => window && window.optimizely
+const getOptmizely = () =>
+  window && window.optimizely && window.optimizely.get && window.optimizely
```
naxhh, line 23: is intentional to evalueate window.optimizely twice?

carlesnunez: Seems to be done to force return on the left to right conditional evaluation.

davidbarna: I need to return window.optimizely only if window.optimizely.get exists.

PR #436, opened 2018-06-28
feat(atom/tag): added new small tag size

beatrizip:
Added 'small' and 'medium' classes and style

### DIFF
--- a/components/atom/tag/src/index.js
+++ b/components/atom/tag/src/index.js
@@ -17,7 +17,7 @@ class AtomTag extends Component {
    * @return {string}
    */
   get _classNames() {
-    const {className, icon, size} = this.props
+    const {icon, size} = this.props
     return classnames(
       'sui-AtomTag',
       `sui-AtomTag-${size}`,
--- a/components/atom/tag/src/index.js
+++ b/components/atom/tag/src/index.js
@@ -21,7 +21,6 @@ class AtomTag extends Component {
     return classnames(
       'sui-AtomTag',
       `sui-AtomTag-${size}`,
-      className,
       icon && 'sui-AtomTag-hasIcon'
     )
   }
--- a/components/atom/tag/src/index.js
+++ b/components/atom/tag/src/index.js
@@ -5,12 +5,25 @@ import classnames from 'classnames'
 import ActionableTag from './Actionable'
 import StandardTag from './Standard'
 
+const SIZES = {
+  MEDIUM: 'medium',
+  SMALL: 'small'
+}
+
 class AtomTag extends Component {
   static MAX_LABEL_LENGTH = 100
 
+  /**
+   * @return {string}
+   */
   get _classNames() {
-    const {className, icon} = this.props
-    return classnames('sui-AtomTag', className, icon && 'sui-AtomTag-hasIcon')
+    const {className, icon, size} = this.props
+    return classnames(
+      'sui-AtomTag',
+      `sui-AtomTag-${size}`,
+      className,
+      icon && 'sui-AtomTag-hasIcon'
+    )
   }
 
   get _isActionable() {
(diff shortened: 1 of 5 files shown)


components/atom/tag/src/index.js
```diff
@@ -5,12 +5,25 @@ import classnames from 'classnames'
...
+   * @return {string}
+   */
   get _classNames() {
-    const {className, icon} = this.props
-    return classnames('sui-AtomTag', className, icon && 'sui-AtomTag-hasIcon')
+    const {className, icon, size} = this.props
+    return classnames(
+      'sui-AtomTag',
+      `sui-AtomTag-${size}`,
+      className,
```
jcger: As you are working on this component, could you please remove the className prop completely as we don't want to make components "styleable" from outside?

jcger (approved):

components/atom/tag/src/index.scss
```diff
@@ -83,4 +84,18 @@ $p-atom-tag: 0 $p-l !default;
       background-color: $c-primary-dark;
     }
   }
+
+  &-small {
+    height: $p-xl;
```
midudev: I'm not sure if we should use here `$p-xl` when the default size is using a `$h-atom-tag`. So, while the default one is customizable via local variables to the component, the small size is customizable using global ones. We might want to use a multiplier here in order to create the small using `$h-atom-tag * 0.8` for example.

The same happen with the line-height and the padding.

Other solution could be having also component variables for the small size.

components/atom/tag/src/index.js
davidbarna: Aren't we making a breaking change removing className? @jcger

jcger: Being strict we should create a major, but it looks like there is nobody (except @beatrizip) using it

https://www.npmjs.com/package/@schibstedspain/sui-atom-tag
https://github.schibsted.io/search?q=sui-atom-tag&type=Code&utf8=✓

components/atom/tag/src/index.scss
beatrizip: ok! tx!

midudev (approved):

components/atom/tag/src/index.scss
```diff
@@ -22,14 +24,14 @@ $p-atom-tag: 0 $p-l !default;
 
 .sui-AtomTag {
   background-color: $bgc-atom-tag;
-  border-radius: ceil($h-atom-tag / 2);
+  border-radius: ceil($h-atom-tag-medium / 2);
```
davidbarna: are convention for "medium" is "m"

components/atom/tag/src/index.scss
```diff
@@ -2,15 +2,17 @@
 
 $bgc-atom-tag: color-variation($c-gray, 3) !default;
 $c-atom-tag-actionable: $c-white !default;
-$h-atom-tag: $p-xxl !default;
+$h-atom-tag-medium: 32px !default;
+$h-atom-tag-small: 24px !default;
```
davidbarna: our convention for "small" is "s"

components/atom/tag/src/index.scss
```diff
@@ -2,15 +2,17 @@
 
 $bgc-atom-tag: color-variation($c-gray, 3) !default;
 $c-atom-tag-actionable: $c-white !default;
-$h-atom-tag: $p-xxl !default;
+$h-atom-tag-medium: 32px !default;
```
davidbarna: I'm not sure if it should be the case here but for atom-button we expressed sizes with paddings. https://github.com/SUI-Components/sui-components/blob/master/components/atom/button/src/index.scss#L14

The question underneath is "if the paddings grow, should tags vertical padding grow as well?". I'd say  "yes". Sizes of tags and buttons are used to mark space between text and border: a padding. So the height should be expressed in paddings and the text vertically centered. It's my opinion.

jcger: I agree with you, I'm gonna talk to the UI guys

components/atom/tag/src/index.scss
beatrizip: 👍

components/atom/tag/src/index.scss
beatrizip: 👍

components/atom/tag/src/index.scss
beatrizip: After the conversation the agreement is to use the height of the component

PablitoGS (approved):

nguasch (approved):

PR #437, opened 2018-07-03
Feat/react form input

jcger:
this PR implements only the input part of the SUI-Input component defined in https://paper.dropbox.com/doc/SUI-Input-03mHJFkOCjviSZevsaTwm. It includes the native and the password input fields. Mask input for postal codes or bank accounts will be done in a new PR. Later one we will implement a new componente which includes the Label, helpText, validationText and this input in order to acomplish with all defined features


beatrizip (approved):

components/form/input/README.md
```diff
@@ -0,0 +1,102 @@
+# FormInput
+
+> Inputs are the text fields that users fill in with different types of information. These include dates, passwords or even short answers. It’s a field where users can write alphanumeric texts.
+
+## Installation
+
+```sh
+ƛ npm install @s-ui/react-form-input --save
```
davidbarna, line 8: What's this symbol `ƛ` ?

components/form/input/README.md
```diff
@@ -0,0 +1,102 @@
...
+
+```js
+import FormInput from '@s-ui/react-form-input'
+
+return <FormInput type='sui-password' />
+```
+
+## Addons
+
+>[What are addons?](https://paper.dropbox.com/doc/SUI-Input-03mHJFkOCjviSZevsaTwm#:uid=125362683844628624581838&h2=Icons-and-addons-inside-the-in)
```
davidbarna, line 35: Should we mention a private document in a public README file ?

components/form/input/README.md
```diff
@@ -0,0 +1,102 @@
...
+
+## Form Usage
+
+Each field returns its value on every onChange event so you can save it inside your form state.
+
+```js
+import React from 'react'
+import ReactDOM from 'react-dom'
+import Input from '@s-ui/react-form-input'
+import Button from '@schibstedspain/sui-atom-button'
```
davidbarna, line 55: At some point we should address the renaming of compliant components packages.
There's a way to do it seemlessly.

components/form/input/README.md
```diff
@@ -0,0 +1,102 @@
...
+  }
+
+  render() {
+    const {login, password} = this.state
+    return (
+      <form>
+        <Input
+          type="text"
+          value={login}
+          onChange={({value}) => this.onChange({value, field: 'login'})}
```
davidbarna: It's weird to me to have an onChange callback that does not return the event...
Something to be discussed.

components/form/input/src/Input/Component/index.js
```diff
@@ -0,0 +1,72 @@
+import React, {Component} from 'react'
+import PropTypes from 'prop-types'
+import cx from 'classnames'
+
+import './index.scss'
```
davidbarna: I don't think this is allowed to do. scss should always be added manually. 
To me we can decide to use this way but I think it may imply problems with themes

components/form/input/src/Input/Component/index.js
```diff
@@ -0,0 +1,72 @@
...
+  /* The DOM id global attribute. */
+  id: PropTypes.string,
+  /* sets the name property of an element in the DOM */
+  name: PropTypes.string,
+  /* onChange callback */
+  onChange: PropTypes.func,
+  /* A hint to the user of what can be entered in the control. The placeholder text must not contain carriage returns or line-feeds. */
+  placeholder: PropTypes.string,
+  /* 's' or 'm', default: 'm' */
+  size: PropTypes.string,
```
davidbarna: `PropTypes.oneOf`

components/form/input/src/Input/Component/index.js
```diff
@@ -0,0 +1,72 @@
...
+  /* text, password, date or number */
+  type: PropTypes.string,
```
davidbarna: `PropTypes.oneOf`

components/form/input/src/Password/index.scss
```diff
@@ -0,0 +1,18 @@
+@import '~@schibstedspain/sui-theme/lib/index';
+
+$r-form-input-password-toggle-button: $gutter !default;
+$pr-form-input-password-input: 60px !default;
```
davidbarna: This should be a padding or margin variable.

davidbarna (commented):
So well structured and easy to understand IMO

components/form/input/README.md
jcger: it's a console prompt, a saturday night console prompt

components/form/input/README.md
jcger: it shouldn't be private

components/form/input/README.md
jcger: It returns it, I'm just not loading it in this example

components/form/input/README.md
jcger: It returns it, I'm just not loading it in this example

Edit: or maybe not, will make it work with ev

components/form/input/src/Input/Component/index.js
jcger: We should check alternatives, loading all styles from the main styles files might make us load css that might not be used

components/form/input/src/Password/index.scss
jcger: I'll remove it as it's not used

davidbarna (approved):

PR #442, opened 2018-07-12
Feat/input mask

jcger:


components/form/input/src/Mask/index.js
```diff
@@ -0,0 +1,47 @@
...
+
+  render() {
+    const {name} = this.props
+    return (
+      <Input
+        id={name}
+        reference={input => {
+          this.field = input
+        }}
+        onChange={this.onChange}
```
davidbarna, line 29: if someone passes onChange prop you onChange handler will not be executed, right?

jcger: which onChange handler? I'm not sure what you mean

davidbarna (approved):

midudev (approved):

PR #447, opened 2018-07-17
Badge contrast

beatrizip:
Created a new variable that allows you to combine different backgrounds and color font.

### DIFF
--- a/components/atom/badge/src/index.scss
+++ b/components/atom/badge/src/index.scss
@@ -1,6 +1,5 @@
 @import '~@schibstedspain/sui-theme/lib/index';
 
-$c-contrast-atom-badge: $c-white !default;
 $bdrs-atom-badge: $bdrs-m !default;
 $fz-atom-badge-l: $fz-m !default;
 $fz-atom-badge-s: $fz-xs !default;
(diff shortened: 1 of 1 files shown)


components/atom/badge/src/index.scss
```diff
@@ -10,7 +9,24 @@ $m-atom-badge-l: $m-m !default;
 $m-atom-badge-s: $m-s !default;
 $p-atom-badge: 0 $p-m !default;
 
-$tags: success $c-success, error $c-error, alert $c-alert, info $c-gray;
+$badges: (
+  success: (
+    background: $c-success,
```
jcger: what about `bgc` instead of `background` and `c` instead of `font`?

jcger (approved):

components/atom/badge/src/index.scss
```diff
@@ -50,22 +66,21 @@ $tags: success $c-success, error $c-error, alert $c-alert, info $c-gray;
...
+      fill: $c;
     }
 
     &-#{$type}--transparent {
       background-color: transparent;
-      color: $color;
-      fill: $color;
+      color: $bgc;
+      fill: $bgc;
+      padding: 0;
```
davidbarna, line 83: can you tell us why you need padding 0 ?

davidbarna (approved):

components/atom/badge/src/index.scss
beatrizip: @davidbarna When it is not framed with a background, padding visually adds space. If you look at the dropbox paper there is no defined padding for transparent badges.

beatrizip: image

PR #452, opened 2018-07-23
Feature/form input radio

juanmaguitar:
Some modifications on `FormInput` component to adapt it to `checkbox` and `radio` types according to → https://paper.dropbox.com/doc/SUI-Checkbox-Radio-button--AIUdrknlqSnTP5T57B8OaDJCAg-tmVSbtoe8nZTaZk9mkdsE


juanmaguitar:
The demo can be seen online at https://sui-components-feature-form-input-radio.now.sh/workbench/form/input/demo

kikoruiz (approved):
It looks great!! 😃

davidbarna (approved):

components/form/input/src/Input/Component/index.scss
```diff
@@ -7,12 +7,23 @@ $sizes-form-input: medium $h-form-input--m, small $h-form-input--s !default;
   border-color: $c-gray-light;
   border-style: solid;
   border-width: 1px;
+  box-sizing: border-box;
   font-size: $fz-base;
   height: $h-form-input--m;
+  width: 100%;
```
jcger: all css props should be sorted alphabetically, it's a warning though

demo/form/input/playground
```diff
@@ -1,3 +1,168 @@
+class SimpleLoginForm extends React.Component {
```
jcger: this is no simple login form, do we really need this here? I think this form shouldn't be part of this playground but in fieldset's playground

juanmaguitar: To understand how to use this component i think a real use case in a real form like this is needed. In the `Fieldset` we can have another form to show the integration of that component in a form.
Also, it shows how to implement some uses of radio buttons like the example with the images as buttons to select a "category"
Besides this, the name of the class it's not right so i'll change it to reflect the purpose of the form

PR #456, opened 2018-07-26
Feature/molecule fieldset

juanmaguitar:


juanmaguitar:
Demo online can be seen at → https://sui-components-feature-molecule-fieldset.now.sh/workbench/molecule/fieldset/demo

Also, an additional demo (to show the use case of a textArea) can be found here → https://poc-field-textarea.stackblitz.io | https://stackblitz.com/edit/poc-field-textarea

components/molecule/fieldset/src/index.js
```diff
@@ -0,0 +1,97 @@
+import React, {Component} from 'react'
+import PropTypes from 'prop-types' // eslint-disable-line no-unused-vars
+import cx from 'classnames'
+
+import AtomValidationText, {
+  AtomValidationTextTypes
+} from '@s-ui/react-atom-validation-text'
+import AtomLabel, {AtomLabelTypes} from '@s-ui/react-atom-label'
+
+import '@s-ui/react-atom-validation-text/lib/index.scss'
```
midudev: You should import this from the sass file instead.

demo/molecule/fieldset/playground
```diff
@@ -0,0 +1,36 @@
...
+  marginTop: '10px',
+  padding: '10px'
+}
+
+return (
+  <div>
+    <h1>MoleculeFieldset</h1>
+    <h2>Basic Usage</h2>
+    <div style={stylesSection}>
+      <MoleculeFieldset label="name">
```
jcger: I miss an example with an helptext

components/molecule/fieldset/src/index.js
```diff
@@ -0,0 +1,97 @@
...
+
+const BASE_CLASS = 'sui-MoleculeFieldset'
+const CLASS_INPUT = `${BASE_CLASS}-input`
+
+class MoleculeFieldset extends Component {
+  state = {
+    value: this.props.children // eslint-disable-line react/prop-types
+  }
+
+  idLabelFor = `sui-MoleculeFieldset-${+new Date()}`
```
jcger: This might be not really accessible, but I'm not sure if we need to be careful with the `name` and `for` attributes, could you please check it?

JoanClaret:
inputs are hidden in mobile devices. is this something that we expect?

image

juanmaguitar:
@JoanClaret solved! This is an issue more related to the playground than to the component itself. It's fixed anyway

components/molecule/fieldset/README.md
```diff
@@ -0,0 +1,61 @@
+# MoleculeFieldset
```
davidbarna: The name is confusing to me. To me, the label and validation doesn't make it a fieldset.
I mean, a fieldset is a set of fields to me, with more that one input spot.

components/molecule/fieldset/src/index.js
```diff
@@ -0,0 +1,88 @@
...
+MoleculeFieldset.displayName = 'MoleculeFieldset'
+
+MoleculeFieldset.propTypes = {
+  /** Text to be displayed as label of the textarea */
+  label: PropTypes.string.isRequired,
+
+  /** used as for attribute. Must be the same as the input element id */
+  name: PropTypes.string.isRequired,
+
+  /** Sucess message to display when sucess state  */
```
davidbarna: typo "sucess" => "success"

davidbarna (approved):

davidbarna (approved):

components/molecule/field/src/index.js
```diff
@@ -0,0 +1,88 @@
...
+  /** Text to be displayed as label of the textarea */
+  label: PropTypes.string.isRequired,
+
+  /** used as for attribute. Must be the same as the input element id */
+  name: PropTypes.string.isRequired,
+
+  /** Success message to display when success state  */
+  successText: PropTypes.string,
+
+  /** Error message to display when error state  */
```
jcger, line 78: it's kind of strange as I do think about an error/success as a state, as well as you do in your comments, but it's resolved via props. Should there really be no state at all?

components/molecule/field/src/index.js
```diff
@@ -0,0 +1,88 @@
...
+    }
+    if (this.props.errorText) {
+      if (element === 'label') return AtomLabelTypes.ERROR
+      if (element === 'validationText') return AtomValidationTextTypes.ERROR
+    }
+  }
+
+  get statusValidationText() {
+    if (this.props.successText) return this.props.successText
+    if (this.props.errorText) return this.props.errorText
```
jcger: what if the developer passed through a success but also an error message? is it just a corner case that shouldn't happen? I intuitively think that the error message should be shown as it is more restrictive

jcger (approved):

components/molecule/field/src/index.js
juanmaguitar: Shouldn't the form including all these `Field`'s be in charge of managing when there is a  sucess/error state for each `Field`?

PR #458, opened 2018-07-27
feat(cmp/services): add support to new cmp contracts and some refactor

pablovegau:


midudev (approved):

carlosvillu (approved):

components/cmp/services/test/getConsentStatusSpec.js
```diff
@@ -13,12 +14,16 @@ describe('getConsentStatus', () => {
...
+    getConsentStatus
+      .execute()
+      .then(response => {
+        expect(window.__cmp.calledOnce).to.be.true
+        const command = window.__cmp.getCall(0).args[0]
+        expect(command).to.equal('getConsentStatus')
+        expect(response).to.equals(CONSENT_STATUS_NOT_ACCEPTED)
+        done()
+      })
+      .catch(error => done(new Error(error)))
```
jcger, line 27: Wouldn't it do the same if you just do `.catch(done)`?. I'm not sure, but I thought that if you call done with an error (which will happen when ending up in the catch) the test will automatically fail

jcger (approved):

components/cmp/services/src/useCases/GetPurposesAndVendorsUseCase.js
```diff
@@ -3,6 +3,31 @@ export class GetPurposesAndVendorsUseCase {
     this._repository = repository
   }
 
+  execute({retrieveConsentsFromCmp} = {}) {
```
davidbarna, line 6: You're returning promises anyways, so this should be enoough:
```js
execute({retrieveConsentsFromCmp} = {}) {
   return retrieve ? this._loadStoredConsents() : this._loadDefaultConsents()
}
```

davidbarna (approved):

components/cmp/services/test/getConsentStatusSpec.js
victuxbb: Nop, the last time that I checked this, if value from catch is false for example, calling function "done(false)" will do the test green.
Seems that done function must get an Error object to fail.

components/cmp/services/src/useCases/GetPurposesAndVendorsUseCase.js
victuxbb: execute is the method were all starts in this use case, so it starts with a Promise flow.
Also, I don't like the idea of depending on returning values of private methods to know what's the type of the returning value for this method.
Anyway, I understand your point but if this code grows, when you will start to think in using a Promise flow?
;)

davidbarna: I don't see what you mean by "promise flow". Only if you want to catch errors wrapping everything into a promise.
Either are fine. async funcs too.

PR #463, opened 2018-07-31
Feature/layout media query

juanmaguitar:


juanmaguitar:
Demo online can be seen at → https://sui-components-feature-layout-media-query.now.sh/workbench/layout/mediaQuery/demo

adriadoz (approved):

carlesnunez (approved):
ILGTM you've done a sick job here! I have a question just to know... why do you created a factory to pass custom breakpoints instead of using a props and react hooks to build it?

juanmaguitar:
@carlesnunez , the main reason is that breakpoints shouldn't be changed for this component unless is strictly necessary. The component is prepared for that in case there's a specific need for this, but most of the times default component with default breakpoints agreed for all schibsted products, should be used

components/layout/mediaQuery/README.md
```diff
@@ -0,0 +1,67 @@
...
+## Usage
+
+### Basic usage (default breakpoints)
+
+```javascript
+import LayoutMediaQuery from '@s-ui/react-layout-media-query'
+
+<LayoutMediaQuery>
+  {({M, L, XL}) => {
+    if (XL) return <p>LargeDesktop</p>
```
davecarter, line 20: Just wondering how this approach might be implemented in our current layouts..
Do you mean that we should return one component or another depending the matching breakpoint??
If so.. this is not the way responsive is.. if you replace components

ddmarin94 (approved):

components/layout/mediaQuery/README.md
jcger: Hi @davecarter, this is one of our biggest concerns. We want to avoid having two apps, but we are reaching our limits because of designs that we need to be compliant with, like the MA home page or some IJ headers. Mostly this component will be used to set one or another prop, component replacement should be the last approach we think of.

carlesnunez:
Hello! Is the release of this sui-component going to take for so long? The MA dev team is excited to use it on our brand new home!

PR #465, opened 2018-08-01
Feature/use new react molecule notification

midudev:
Drop @schibstedspain/sui-alert-basic dependency and use the new and shiny **@s-ui/react-molecule-notification** in order to use the component approved by UX and, also, save a few bytes by avoiding adding stuff that we for sure already have in our web.

image


jcger (commented):
wow, nice!

ruben-martin-lozano (approved):

danivm (approved):

jcger (approved):

PR #466, opened 2018-08-02
Feature/collapsible basic is clickable

ddmarin94:
This feature allows us to manage when should be clickable the collapsible basic. For example, when has been opened for the first time we can set isCickable prop to false and the user won't be able to close the component.

### DIFF
--- a/components/collapsible/basic/src/index.js
+++ b/components/collapsible/basic/src/index.js
@@ -36,7 +36,8 @@ class CollapsibleBasic extends Component {
       label,
       animationSpeed,
       hideTriggerIcon,
-      children
+      children,
+      isClickable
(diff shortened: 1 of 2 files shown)


midudev (approved):

components/collapsible/basic/src/index.js
```diff
@@ -100,6 +101,10 @@ CollapsibleBasic.propTypes = {
   /**
    * Customise the speed of the transition animation: normal 0.3s, fast: 0.15s
    */
+  isClickable: PropTypes.bool,
+  /**
```
jcger: This prop description isn't at the right place

jcger (approved):

PR #471, opened 2018-08-08
feat(atom/textarea): added border to textarea

juanmaguitar:
Fix to solve the issue of textarea without border on mobile devices


jcger (approved):

midudev (approved):
I think you could sort the properties accordingly or it will give you a warning in the linter but :+1:

davidbarna (approved):

PR #474, opened 2018-08-16
Textarea value prop

juanmaguitar:
Modified textArea to only display value received via `value` prop


juanmaguitar:
Online demo can be checked at → https://sui-components-textarea-value-prop.now.sh/workbench/atom/textarea/demo

davidbarna (approved):

jcger (approved):

PR #475, opened 2018-08-17
Header actions

jcger:
image


components/header/actions/src/index.js
```diff
@@ -0,0 +1,42 @@
+import React from 'react'
+import PropTypes from 'prop-types'
+import Button from '@schibstedspain/sui-atom-button'
+
+const getClass = type => `sui-HeaderActions${type ? `-${type}` : ''}`
```
davidbarna, line 5: ❤️

components/header/actions/src/index.scss
```diff
@@ -0,0 +1,28 @@
+@import '~@schibstedspain/sui-theme/lib/index';
+@import '~@schibstedspain/sui-atom-button/lib/index';
+
+$bgc-header-actions: $c-primary !default;
+$c-header-actions: $c-white !default;
+$w-header-actions-icon: 42px !default;
```
davidbarna: we have variables for icon size: https://github.com/SUI-Components/sui-theme/blob/master/src/settings/_size.scss

davidbarna (approved):

components/header/actions/src/index.scss
jcger: I wasn't sure about using them because this isn't a sui-compliant/related component. Do you think I should use it anyway?

davidbarna: if it fits, yes. Even non compliant components rely on sui-theme variables

davidbarna (approved):

PablitoGS (approved):

PR #476, opened 2018-08-17
feat(molecule/fieldTextarea): component moleculeFieldTextarea

juanmaguitar:


juanmaguitar:
Online demo can be checked at → https://sui-components-feature-molecule-field-textarea.now.sh/workbench/molecule/fieldTextarea/demo

jcger (approved):

demo/molecule/fieldTextarea/playground
```diff
@@ -0,0 +1,62 @@
+const styleList = {
```
jcger, line 1: you could add one example with less `maxCharacters` in order to test it easier

components/molecule/fieldTextarea/README.md
```diff
@@ -0,0 +1,93 @@
...
+```
+
+### With maxCharacters
+    
+```
+  <MoleculeFieldTextarea
+    id="description"
+    label="Description"
+    value="In some place of La Mancha which name..."
+    maxCharacters={75}
```
davidbarna: the shorter the API the better as long as it's easily understandable. IMO `maxChars` would be a better name.

davidbarna (approved):

components/molecule/fieldTextarea/README.md
juanmaguitar: 👍 changed to `maxChars`

demo/molecule/fieldTextarea/playground
juanmaguitar: 👍 `maxChars={30}`

PR #490, opened 2018-08-29
Feature/molecule button group

juanmaguitar:
Component `MoleculeButtonGroup` according to specifications provided at → https://jira.scmspain.com/browse/EFE-179
Also modifications in `AtomButton` to provide some props needed for `MoleculeButtonGroup`


components/molecule/buttonGroup/src/index.scss
```diff
@@ -0,0 +1,7 @@
+@import "~@schibstedspain/sui-atom-button/lib/index";
```
jcger: There shouldn't be any atom button dependency. The styles should be loaded in the component which uses the button group using atom buttons as child. It would be top if the button group component would work flawless even for html buttons as childs

components/molecule/buttonGroup/package.json
```diff
@@ -0,0 +1,18 @@
...
+  "description": "",
+  "main": "lib/index.js",
+  "scripts": {
+    "build": "npx rimraf ./lib && npx mkdirp ./lib && npm run build:js && npm run build:styles",
+    "build:js": "../../../node_modules/.bin/babel --presets sui ./src --out-dir ./lib",
+    "build:styles": "../../../node_modules/.bin/cpx './src/**/*.scss' ./lib"
+  },
+  "dependencies": {
+    "@s-ui/component-dependencies": "1",
+    "@schibstedspain/sui-atom-button": "1"
```
jcger: this should be removed

jcger (changes requested):

demo/molecule/buttonGroup/playground
```diff
@@ -0,0 +1 @@
+return (<MoleculeButtonGroup />)
```
davidbarna, line 1: ?

components/molecule/buttonGroup/package.json
juanmaguitar: I need it to be able to do this → `import {atomButtonGroupPositions} from '@schibstedspain/sui-atom-button'` (added after @davidbarna review comment)

components/molecule/buttonGroup/src/index.js
```diff
@@ -0,0 +1,27 @@
...
+
+import React from 'react'
+import cx from 'classnames'
+
+const BASE_CLASS = 'sui-MoleculeButtonGroup'
+
+const MoleculeButtonGroup = ({children, fullWidth, ...props}) => {
+  const numChildren = children.length
+  const extendedChildren = React.Children.map(children, (child, index) => {
+    const groupPosition = (() => {
```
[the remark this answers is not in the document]
juanmaguitar: Why not? Any suggestion?

davidbarna (dismissed):

components/molecule/buttonGroup/package.json
jcger: It would be cleaner having the positions as props instead of this hard dependency

components/molecule/buttonGroup/src/index.js
```diff
@@ -0,0 +1,46 @@
...
+}
+
+MoleculeButtonGroup.displayName = 'MoleculeButtonGroup'
+
+MoleculeButtonGroup.propTypes = {
+  /** If buttons should stretch to fit the width of container */
+  fullWidth: PropTypes.bool,
+
+  /** groupPositions constants (FIRST, MIDDLE, LAST) */
+  groupPositions: PropTypes.object.isRequired
```
jcger: if it is required you should update the README

jcger: I would suggest setting as default values the values that would work with the atomButton

juanmaguitar: You mean, adding these values directly like ...
```
MoleculeButtonGroup.defaultProps = {
  groupPositions: {
    FIRST: 'first',
    MIDDLE: 'middle',
    LAST: 'last'
  }
}
```

jcger: yes

jcger (approved):

davidbarna (approved):

PR #496, opened 2018-08-31
feat(card/basic): add a Call To Action to the BasicCard

devictoribero:
Allow the basic card to have a CTA.

### DIFF
--- a/components/card/basic/src/index.js
+++ b/components/card/basic/src/index.js
@@ -91,7 +91,7 @@ CardBasic.propTypes = {
   /**
    * Button for actions
    */
-  button: PropTypes.object
+  button: PropTypes.node
 }
 
 CardBasic.defaultProps = {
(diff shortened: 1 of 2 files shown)


components/card/basic/src/index.js
```diff
@@ -84,7 +87,11 @@ CardBasic.propTypes = {
   /**
    * Lazy load flag / config.
    */
-  lazyLoad: PropTypes.oneOfType([PropTypes.bool, PropTypes.object])
+  lazyLoad: PropTypes.oneOfType([PropTypes.bool, PropTypes.object]),
+  /**
+   * Button for actions
+   */
+  button: PropTypes.object
```
kikoruiz: It would rather be a `PropTypes.node`, isn't it?

kikoruiz (dismissed):

components/card/basic/src/index.js
devictoribero: Yes, that's true.

JoanClaret (approved):

davidbarna (approved):

jcger (approved):

PR #497, opened 2018-08-31
No new button on every render

jcger:
I also added the `isSubmit` prop which adds `type="submit"` to the `<button />` tag if set


davidbarna (approved):
👏

midudev (approved):

PR #502, opened 2018-08-31
fix(META): fix wrong sass import

davidbarna:
image


midudev:
I think this is already fixed: https://github.com/SUI-Components/sui-components/pull/498

davidbarna:
LOL

PR #504, opened 2018-09-03
Revert "No new button on every render"

midudev:
- [x] Reverts SUI-Components/sui-components#497, link buttons are not shown correctly.
- [x] Keep isSubmit feature still.


davidbarna (dismissed):
Can you tell us what happened ?

davidbarna (approved):

ruben-martin-lozano:
Basically with the 9ef2312 refactor, the props link/href were not used in the new Button subcomponent (since they were cleaned in "cleanProps" method), so it never render a Link, and it is needed, at least, for PTA button in FC.

ruben-martin-lozano (approved):

PR #512, opened 2018-09-04
feat(form/rangeDatepicker): add control to check date entries on change

devtoni:
Until now, user could set a start date after than the end date, with this change we check this situation and prevent it assigning to the endDate the startDate.
Also I made some refactor.


kikoruiz (approved):

davidbarna (approved):
IMO this is logic that is part of the domain...
But as it's just MY opinion. So approved.

PR #520, opened 2018-09-12
Bugfix/molecule notification auto close fix

devictoribero:
Because of the `autoClose` behaviour was not working, I propose this solution.

Any thoughts?

### DIFF
--- a/components/molecule/notification/src/index.js
+++ b/components/molecule/notification/src/index.js
@@ -38,6 +38,18 @@ class MoleculeNotification extends Component {
     this.state.show !== nextProps.show && this.toggleShow()
   }
 
+  componentDidMount() {
+    const {show, autoClose: autoCloseTiming} = this.props
+
+    if (
+      show &&
+      AUTO_CLOSE_TIME[autoCloseTiming] &&
+      autoCloseTiming !== AUTO_CLOSE_TIME[autoCloseTiming]
+    ) {
+      this.autoClose(AUTO_CLOSE_TIME[autoCloseTiming])
+    }
+  }
+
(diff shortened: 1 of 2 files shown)


components/molecule/notification/src/index.js
```diff
@@ -38,6 +38,18 @@ class MoleculeNotification extends Component {
     this.state.show !== nextProps.show && this.toggleShow()
   }
 
+  componentDidMount() {
+    const {show, autoClose: autoCloseTiming} = this.props
```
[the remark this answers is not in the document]
devictoribero: if encapsulated the checked in a function so its not duplicated. Anyway... the name for sure could be more accurated. I accept proposals.

Since autoClose is a function in the component,  I would do the check inside the function following the `tell, dont ask principle`

components/molecule/notification/src/index.js
```diff
@@ -38,6 +38,10 @@ class MoleculeNotification extends Component {
     this.state.show !== nextProps.show && this.toggleShow()
   }
 
+  componentDidMount() {
+    this.autoCloseIfRequired()
```
davidbarna: What if show is `false` here. You're setting an unneeded timeout, right ?

devictoribero: True. The fact that has a different behaviour when `show=true` or `show=false` with the `autoclose prop` has confused me again 🤣

kikoruiz (dismissed):

kikoruiz (dismissed):

kikoruiz (approved):

jcger (approved):

PR #521, opened 2018-09-12
[layout/mediaQuery] initial breakpoints

carlosvillu:


jcger (dismissed):

components/layout/mediaQuery/src/index.js
```diff
@@ -10,9 +10,19 @@ import {BREAKPOINTS} from './breakpoints'
 
 const LayoutMediaQueryFactory = BREAKPOINTS =>
   class extends Component {
+    static defaultProps = {
+      initialMediaQueries: {}
+    }
+
+    static propTypes = {
+      /** MediaQueries to be used in the first render. Until of be called componentDidMount and setup the proper MQ´s Usefull for SSR */
```
midudev:
```
  /**
   * MediaQueries to be used in the first render until componentDidMount is called and it setups the proper media query. Useful for SSR for avoiding re-renders.
   */
```

midudev: There're some typos fixed and used the correct format for the comment.

demo/layout/mediaQuery/playground
```diff
@@ -53,7 +53,7 @@ return (
     <p>Resize the width of the viewport or the container with dashed blue border to see the behavior</p>
     <h2>Basic Usage (default values)</h2>
     <div style={stylesSection}>
-      <LayoutMediaQuery>
+      <LayoutMediaQuery initialMediaQueries={{XS: true, S: true}}>
```
midudev, line 56: How is that working with two initialMediaQueries as true?

midudev (commented):
I miss further documentation in order to understand how it works and how to use it.

components/layout/mediaQuery/src/index.js
```diff
@@ -57,10 +68,19 @@ const LayoutMediaQueryFactory = BREAKPOINTS =>
       if (!shallowEqual(result, params)) this.setState({params: result})
     }
 
+    shouldComponentUpdate(_, nextState) {
+      return !shallowEqual(this.state.params, nextState.params)
```
carlesnunez, line 74: This hook is not required as long while you are using a PureComponent.
https://reactjs.org/docs/shallow-compare.html

In any case.. im not pretty sure about that we need a pure component on this piece. Why did you changed the stateless version for a pure component?

carlosvillu: I had remove the pure component. And left the shouldComponentUpdate. The PureComponent only didnt works.

PR #532, opened 2018-09-24
feat(atom/input): icon Left

juanmaguitar:
https://jira.scmspain.com/browse/EFE-212
Demo online can be checked at → https://sui-components-feature-input-icons.now.sh/workbench/atom/input/demo


components/atom/input/src/Input/Features/WithIcon/index.scss
```diff
@@ -0,0 +1,31 @@
...
+  .sui-AtomInput-input {
+    padding-left: 50px;
+  }
+
+  &--icon {
+    align-items: center;
+    color: $c-primary;
+    cursor: pointer;
+    display: flex;
+    height: 25px;
```
JoanClaret: what about using sui-theme icon sizes?
https://github.com/SUI-Components/sui-theme/blob/master/src/components/_icon.scss

btw, @jcger  should we use icon sizes related to the input size?

components/atom/input/src/Input/Features/WithIcon/index.scss
```diff
@@ -0,0 +1,31 @@
...
+    justify-content: center;
+    left: $r-atom-input-password-toggle-button;
```
JoanClaret: i guess `r` is a right abbreviature, but it's assigned to the `left` property.

components/atom/input/src/Input/Features/WithIcon/index.js
```diff
@@ -0,0 +1,31 @@
+/* eslint-disable react/prop-types, no-unused-vars, no-console */
```
davidbarna: Don't use unnecessary disablers.

components/atom/input/src/Input/Features/WithIcon/index.js
```diff
@@ -0,0 +1,31 @@
...
+
+import React from 'react'
+import PropTypes from 'prop-types'
+
+const CLASS_ICON = 'sui-AtomInput-withIcon'
+const CLASS_ICON_COMPONENT = `${CLASS_ICON}--icon`
+
+const WithIconHoC = WrappedInput =>
+  class Addon extends React.Component {
+    static propTypes = {
```
davidbarna: I like to define props this way better. But I thought we had a differente convention that implies to put it in bottom.

components/atom/input/src/Input/Features/WithIcon/index.scss
```diff
@@ -0,0 +1,31 @@
...
+    position: absolute;
+    top: 50%;
+    transform: translateY(-50%);
+    width: 25px;
```
davidbarna: 25px and 50px are values related to the icon size. Please use variables.

components/atom/input/src/Input/index.js
```diff
@@ -1,16 +1,16 @@
...
 
 export default props => {
-  const {leftAddon, rightAddon} = props // eslint-disable-line react/prop-types
-  return leftAddon || rightAddon ? (
-    <WithAddonComponent {...props} />
-  ) : (
-    <Component {...props} />
-  )
+  const {leftAddon, leftIcon, rightAddon} = props // eslint-disable-line react/prop-types
+  if (leftAddon || rightAddon) return <WithAddonComponent {...props} />
```
davidbarna, line 9: Si I can't have an input with icon AND right addon ?
Does it make sense to treat icons as addons as well ?

davidbarna (changes requested):

components/atom/input/src/Input/Features/Addon/index.scss
```diff
@@ -4,6 +4,13 @@ $bgc-atom-input-addon: $c-gray-lightest !default;
 $bdc-atom-input-addon: $c-gray-light !default;
 $c-atom-input-addon: $c-gray-light !default;
 
+$l-atom-input-icon: $gutter !default;
+$w-atom-input-icon: $sz-icon-m !default;
+$h-atom-input-icon: $sz-icon-m !default;
+$t-atom-input-icon: 50% !default;
```
davidbarna: 50% and -50% is the same value.
But they don't need to have variables anyway.

components/atom/input/src/Input/Features/Addon/index.scss
```diff
@@ -4,6 +4,13 @@ $bgc-atom-input-addon: $c-gray-lightest !default;
...
+$trf-ty-atom-input-icon: -50% !default;
+$pl-atom-input-icon: 50px !default;
```
davidbarna: Are you sure this is not `$sz-icon-m*2`?
What may happen to your component if $sz-icon-m is 30px ?

davidbarna (dismissed):

components/atom/input/src/Input/Features/Addon/index.scss
juanmaguitar: This value uses a `sui-theme` variable and is configurable so I don't see the issue here. What's your point/suggestion?

components/atom/input/src/Input/Features/Addon/index.scss
juanmaguitar: It is the same vale but they refer to different properties so I think it's better to have different variables for each value. Also these values refer to Icon positioning that may be related to Icon dimensiones that are configurable, so I think these value should also be configurable

ddmarin94 (dismissed):

components/atom/input/src/Input/Features/WithIcon/index.scss
jcger: we are using https://github.com/SUI-Components/sui-theme/blob/master/src/settings/_size.scss for icon sizes. (hope the sui-theme refactor happens soon). If related or not should be defined by UX, looks light they only need one icon size at the moment

components/atom/input/src/Input/Features/Addon/index.scss
```diff
@@ -27,4 +34,33 @@ $c-atom-input-addon: $c-gray-light !default;
   &--left {
     border-width: $bdw-atom-input-addon-left;
   }
+
+  &--withIcon {
+    position: relative;
+    width: 100%;
+  
+    .sui-AtomInput-input {
```
jcger: couldn't this be done via props in order to maintain it encapsulated?

components/atom/input/src/Input/index.js
```diff
@@ -5,12 +5,10 @@ import Component, {inputSizes} from './Component'
...
 export default props => {
-  const {leftAddon, rightAddon} = props // eslint-disable-line react/prop-types
-  return leftAddon || rightAddon ? (
-    <WithAddonComponent {...props} />
-  ) : (
-    <Component {...props} />
-  )
+  const {leftAddon, leftIcon, rightAddon} = props // eslint-disable-line react/prop-types
+  if (leftAddon || rightAddon || leftIcon)
+    return <WithAddonComponent {...props} />
```
jcger: not sure if leftIcon be treated as an Addon

components/atom/input/src/Input/Features/Addon/index.scss
juanmaguitar: what do you suggest for this?

demo/atom/input/playground
```diff
@@ -215,6 +216,21 @@ return (
...
+        name="second"
+        placeholder="Medium Input"
+        leftIcon={LeftAddonComponent}
+      />
+    </div>
+    <div style={field}>
+      <AtomInput
+        name="second"
+        placeholder="Medium Input"
+        leftIcon={LeftAddonComponent}
```
jcger: there should be no left addon if a left icon is set, from the paper ```If it is combined with an icon, the addon must be placed to the right side, because the icon must be placed at the left side.```

jcger: but only remove the example

jcger (dismissed):

jcger (approved):

carlosvillu (approved):

ddmarin94 (approved):

PR #533, opened 2018-09-25
feat(atom/input): sizeChars

juanmaguitar:
https://jira.scmspain.com/browse/EFE-211
Online demo can be checked at → https://sui-components-feature-input-fixed-width.now.sh/workbench/atom/input/demo

### DIFF
--- a/components/atom/input/src/Input/Component/index.js
+++ b/components/atom/input/src/Input/Component/index.js
@@ -24,11 +24,11 @@ class Input extends Component {
     return ''
   }
 
-  getClassNames({size, sizeChars, hideInput, errorState}) {
+  getClassNames({size, charsSize, hideInput, errorState}) {
     return cx(
       CLASS,
       `${CLASS}-${size}`,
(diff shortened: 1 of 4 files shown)


components/atom/input/src/Input/Component/index.js
```diff
@@ -85,6 +88,8 @@ Input.propTypes = {
   placeholder: PropTypes.string,
   /* 's' or 'm', default: 'm' */
   size: PropTypes.oneOf(Object.values(SIZES)),
+  /* width of input based in number of characters (native "size" attribute) */
+  sizeChars: PropTypes.number,
```
davidbarna: I'd say charsSize in proper english.
(Also, as its the same feature as `cols` in textarea, we could use it as alternative name of prop.)

davidbarna (dismissed):

davidbarna (approved):

JoanClaret (approved):

PR #541, opened 2018-10-09
Feature/atom back to top

juanmaguitar:
https://jira.scmspain.com/browse/SUIC-41
Demo online can be checked at → https://sui-components-feature-atom-back-to-top.now.sh/workbench/atom/backToTop/demo


demo/atom/backToTop/playground
```diff
@@ -0,0 +1,247 @@
...
+          d="M9.4,109.3 L114.4,9 C120.2,3 128.1,0 136.8,0 C145.5,0 153.3,3 159.2,9 L264.6,109.3 C277.1,121.2 277.1,140.6 264.6,152.5 C252.1,164.4 231.9,164.4 219.4,152.5 L169,104.4 L169,321.4 C169,338.3 154.7,352 137,352 C119.3,352 105,338.3 105,321.4 L105,104.4 L54.6,152.6 C42.1,164.5 21.9,164.5 9.4,152.6 C-3.1,140.6 -3.1,121.3 9.4,109.3 Z"
+          id="Shape"
+        />
+      </g>
+    </g>
+  </svg>
+)
+
+return (
+  <div>
```
jcger, line 139: the example is to short for big screens, you should make it longer

components/atom/backToTop/src/index.js
```diff
@@ -0,0 +1,144 @@
...
+
+const STYLES = {
+  DARK: 'dark',
+  LIGHT: 'light'
+}
+
+class AtomBackToTop extends PureComponent {
+  state = {
+    show: null,
+    hover: false
```
jcger, line 22: Couldn't this be done via css directly with :hover? Looks like you are just adding a css class

components/atom/backToTop/src/utils.js
```diff
@@ -0,0 +1,34 @@
+const isReactRefObj = target => {
```
jcger: If you gonna reuse this, it's ok to move this to a `react-components` https://github.com/SUI-Components/sui/tree/master/packages/sui-js/src folder or something like that

components/atom/backToTop/src/index.js
juanmaguitar: I found 2 problems when using the native css `hover`:
- The `AtomBackToTop` doesn't disappear when the scroll crosses the half height if the pointer is still over the button (a common behaviour)
- On mobile The `AtomBackToTop` doesn't disappear even when the scroll reaches the top (it needs a click outisde the button to get the `unhover` event )

By Managing this through JS I got more control over this and I could solve these issues

You can see these problems when implementing this w/ the native CSS `hover` here → https://sui-components-feature-atom-back-to-top-hover-css.now.sh/workbench/atom/backToTop/demo

components/atom/backToTop/src/utils.js
juanmaguitar: https://github.com/SUI-Components/sui/pull/411

demo/atom/backToTop/playground
juanmaguitar: Solved w/ a button in the Demo that adds paragraphs on demand

demo/atom/backToTop/playground
```diff
@@ -0,0 +1,108 @@
...
+          </button>
+        </div>
+        <div style={ dark ? { marginTop: '10px', padding: '10px', background: '#4d4d4d', color: 'white'} : {padding: '10px'}}>
+          {new Array(paragraphs)
+            .fill(0)
+            .map((elem, index) => <Paragraph key={index} />)}
+        </div>
+        <AtomBackToTop
+          iconTop={iconTop}
+          refContainer=".sui-Studio-main"
```
jcger: the demo doesn't work in dev mode as this class doesn't exists there

dev mode  → `npm run dev atom/backToTop`

juanmaguitar: solved!

components/atom/backToTop/src/index.scss
```diff
@@ -0,0 +1,82 @@
+@import '~@schibstedspain/sui-theme/lib/index';
+
+$z-back-to-top: 10000;
```
jcger: maybe you could use on the new z-index vars

juanmaguitar: Yes, but which one?
```
$z-spinner: 1500 !default;
$z-cmp: 1400 !default;
$z-banner: 1300 !default;
$z-notifications: 1200 !default;
$z-modal: 1100 !default;
$z-sidebar: 1000 !default;
$z-tooltips: 900 !default;
$z-navigation: 800 !default;
```

Maybe `$z-tooltips` ?

danivm (approved):

jcger (approved):

PR #542, opened 2018-10-10
Feature/atom progress bar

juanmaguitar:
https://jira.scmspain.com/browse/SUIC-32
Demo online can be checked at → https://sui-components-feature-atom-progress-bar.now.sh/workbench/atom/progressBar/demo


components/atom/progressBar/src/index.js
```diff
@@ -0,0 +1,53 @@
...
+        </div>
+        {indicatorBottom && this.renderIndicator(indicatorBottom)}
+      </div>
+    )
+  }
+}
+
+AtomProgressBar.displayName = 'AtomProgressBar'
+
+// Remove these comments if you need
```
midudev: Remove the comments not needed. :)

components/atom/progressBar/src/index.js
```diff
@@ -0,0 +1,53 @@
...
+import PropTypes from 'prop-types'
+import cx from 'classnames'
+
+const BASE_CLASS = 'sui-AtomProgressBar'
+const CLASS_INDICATOR = 'sui-AtomProgressBar-indicator'
+const CLASS_CONTAINER_BAR = 'sui-AtomProgressBar-container'
+const CLASS_BAR = 'sui-AtomProgressBar-bar'
+
+class AtomProgressBar extends PureComponent {
+  renderIndicator(indicatorBottom) {
```
midudev: Usually, I would like more a component itself than a render method.

midudev: Also, IMHO it's a bit weird the implementation of this method. As the indicatorBottom is retrieved from param but still is accessing from props the rest of props. I would or use only named params or only use props.

components/atom/progressBar/src/index.scss
```diff
@@ -0,0 +1,42 @@
+@import '~@schibstedspain/sui-theme/lib/index';
+
+$bg-progress-bar: $c-gray-lightest;
```
midudev: Should these be default values in order to be custom or are not customizable any of them?

components/atom/progressBar/src/index.scss
```diff
@@ -0,0 +1,42 @@
...
+    &--bottom {
+      margin-top: $m-base;
+    }
+    
+  }
+
+  &-container {
+    background: $bg-progress-bar;
+    border: 1px solid $bc-progress-bar;
+    border-radius: 25px;
```
midudev: As the 25px is used twice, might be is a good idea to extract to a variable in order to change once and get the value. Even if not customizable.

midudev (dismissed):
I rather like to see the usage of the progress HTML element but I assume it has some styling considerations that could make it harder to work with.

components/atom/progressBar/src/index.js
juanmaguitar: 👍

components/atom/progressBar/src/index.scss
juanmaguitar: 👍

juanmaguitar:
Regarding this → https://github.com/SUI-Components/sui-components/pull/542#pullrequestreview-163751283
Yes, with the `progress` tag the styling is VERY limited and every browser will display its own version.

components/atom/progressBar/src/index.scss
juanmaguitar: I don't see anything in the specification regarding this but I understand that these colors should be customizable

midudev: Ok! We could ask UX guys in order to certify this, but if the colors are customizable then we should use `!default` in these variables.

midudev (dismissed):

carlosvillu (dismissed):

midudev (dismissed):

ddmarin94 (dismissed):

devictoribero (dismissed):

components/atom/progressBar/src/index.scss
```diff
@@ -0,0 +1,49 @@
+@import '~@schibstedspain/sui-theme/lib/index';
+
+$bg-progress-bar: $c-gray-lightest !default;
+$bc-progress-bar: $c-gray-light !default;
+$c-progress-bar-fill: $c-primary !default;
+
+$trsdu-progress-bar: 500ms !default;
+$trstf-progress-bar: linear !default;
+
+$br-progress-bar: 25px;
```
jcger: this was 4px, right?

ddmarin94 (dismissed):

components/atom/progressBar/README.md
```diff
@@ -0,0 +1,38 @@
+# AtomProgressBar
+
+AtomProgressBar is a component that displays a progress bar
```
jcger: what about `A Progress bar is a visual representation of a specified wait time or the current status of a process.`?

jcger (dismissed):

ddmarin94 (approved):

jcger (approved):

midudev (approved):

components/atom/progressBar/src/index.js
```diff
@@ -0,0 +1,62 @@
...
+import cx from 'classnames'
+
+const BASE_CLASS = 'sui-AtomProgressBar'
+const CLASS_INDICATOR = 'sui-AtomProgressBar-indicator'
+const CLASS_CONTAINER_BAR = 'sui-AtomProgressBar-container'
+const CLASS_BAR = 'sui-AtomProgressBar-bar'
+const CLASS_BAR_ANIMATED = 'sui-AtomProgressBar-bar--animated'
+
+const Indicator = props => {
+  const {indicatorBottom, percentage, indicatorTotal} = props // eslint-disable-line react/prop-types
```
carlosvillu, line 12: Why are avoid add the prop types definition for this component ?!

carlosvillu (approved):

PR #544, opened 2018-10-16
Molecule/Accordion

adriadoz:
accordion_480


components/atom/accordion/README.md
```diff
@@ -0,0 +1,74 @@
...
+```
+
+### Basic usage
+```js
+<AtomAccordion 
+    icon={svgIcon} 
+    showText='Show' 
+    hideText='Hide'
+>
+    <MoleculeCollapsible
```
juanmaguitar: Shouldn't `AtomAccordion` be `MoleculeAccordion`?
Is it ok having _atoms_ (`AtomAccordion`) using _molecules_ (`MoleculeCollapsible`) ? Shouldn't this be the opposite?

components/atom/accordion/src/index.js
```diff
@@ -0,0 +1,94 @@
...
+        this.state.childrenStateList.map((child, index) => {
+          return index !== id
+        })
+      )
+    })
+    childrenNewStatePromise.then(childrenNewStateList => {
+      this.setState({
+        childrenStateList: [...childrenNewStateList]
+      })
+      this.forceUpdate()
```
jcger: You shouldn't need to create a promise nor to force the update. I've seen that you are using componentWillReceiveProps within de collapsible. I think this is breaking things/not working as expected (it's also deprecated). Try getDerivedStateFromProps and/or componentDidUpdate and I'm pretty sure you won't need this promise/force hack

components/molecule/accordion/src/index.scss
```diff
@@ -0,0 +1,41 @@
+@import '~@schibstedspain/sui-theme/lib/index';
+@import '../../collapsible/src/index';
+@import './settings';
+.sui-MoleculeAccordion {
+  .sui-MoleculeCollapsible {
```
JoanClaret: Not sure if this a good practice. IMHO we shouldn't overwrite other component styles.
Is `MoleculeCollapsible` definition providing a borders specification? if so, maybe we can send a prop to the `MoleculeCollapsible` in order to handle the style changes there. Otherwise UX could  rethink `MoleculeCollapsible` specification. 
This discussion starts to be recurrent. More opinions about that would be apreciated.

devictoribero: I do think as @JoanClaret, if we overwrite styles, the components are less consistent. I think that's an anti-pattern we could avoid. IMHO situations like this one should make us send a re-definition to UX or think about the limitations of the component

adriadoz: It's not only the _border_ property that has to be changed,  all the design from `MoleculeCollapsible` doesn't fit with `MoleculeAccordion`'s design. The initial idea was to reuse the behaviour from `MoleculeCollapsible` as children for `MoleculeAccordion` but not the styles. So, would you see it better if this styles were in `MoleculeCollapsible`'s css as a special case when nested in `MoleculeAccordion`?

JoanClaret: ok, i see. 
Considering that `MoleculeCollapsible` has totally different styling, and its functionality does not need a large development (just toggle a layer), for me, the risks of overwriting classes and break the component isolation are bigger than the benefit of importing this component , so I would simply bet for develop `MoleculeAccordion` from scratch.

jcger: @adriadoz please let's talk about this

components/molecule/accordion/src/index.js
```diff
@@ -0,0 +1,90 @@
+import React, {Component} from 'react'
+import PropTypes from 'prop-types'
+import MoleculeCollapsible from '../../collapsible/src'
+
+const BASE_CLASS = 'sui-MoleculeAccordion'
+
+class MoleculeAccordion extends Component {
+  constructor(props) {
```
midudev, line 8:
```
state = {
  childrenStateList: this.getChildrenInitialStateList()
}
```

PR #548, opened 2018-10-19
Atom input icon right click

juanmaguitar:
https://jira.scmspain.com/browse/SUIC-78
Online demo can be checked at → https://sui-components-atom-input-icon-right-click.now.sh/workbench/atom/input/demo

### DIFF
--- a/components/atom/input/src/Input/Features/Icon/index.scss
+++ b/components/atom/input/src/Input/Features/Icon/index.scss
@@ -1,18 +1,27 @@
 $l-atom-input-icon: $gutter !default;
+$r-atom-input-icon: $gutter !default;
 $w-atom-input-icon: $sz-icon-m !default;
 $h-atom-input-icon: $sz-icon-m !default;
 $t-atom-input-icon: 50% !default;
 $trf-ty-atom-input-icon: -50% !default;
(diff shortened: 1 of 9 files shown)


components/atom/input/src/Input/Features/Icon/index.scss
```diff
@@ -0,0 +1,53 @@
...
+    }
+
+    &--right #{$class-input} {
+      padding-right: $pr-atom-input-input;
+    }
+
+    &-icon {
+      align-items: center;
+      color: $c-primary;
+      cursor: pointer;
```
jcger: maybe this should only happen if there is an onClick action, should be easy adding an extra css class in case the onClick is defined

jcger: maybe this should only happen if there is an onClick action, should be easy adding an extra css class in case the onClick is defined

jcger: maybe this should only happen if there is an onClick action, should be easy adding an extra css class in case the onClick is defined

jcger (approved):

danivm (approved):

PR #553, opened 2018-10-30
feat(molecule/tabs): add new prop: allowHandleDisabledTabClick

davidmartin84:


jcger (dismissed):

devictoribero (approved):

JoanClaret (dismissed):

components/molecule/tabs/src/index.js
```diff
@@ -21,7 +21,9 @@ class MoleculeTabs extends Component {
   _createHandleChange(index) {
     return event => {
       event.preventDefault()
-      if (this.props.items[index].enabled !== false) {
+      const {items, allowHandleDisabledTabClick} = this.props
+      const {enabled} = items[index]
+      if (enabled !== false || allowHandleDisabledTabClick) {
```
midudev: I think this should be a && instead ||.

tureey: Instead of `enabled !== false` we could: `enabled === true` that is the same than just `enabled`, right? Or am I missing something?  And yes, it should be `&&` good catch @miduga

devictoribero: Instead of `enabled !== false` we could: `enabled === true` that is the same than just `enabled`, right? Or am I missing something?  And yes, it should be `&&` good catch @miduga

devictoribero (changes requested):

components/molecule/tabs/src/index.js
davidmartin84: isn't the same. enabled can be any value, "" 1 "hello" etc. with enabled === true we don't cover those cases

JoanClaret: `!!enabled === true` ?
:trollface:

jcger: we should avoid blocking PR because this kind of stuff, it works as it is, it has no performance boost and everyone can read it even if it is the negative way around

jcger:
please add an example of this to the playground

JoanClaret (dismissed):
After taking a look with David, i think the best solution is the one that was commited by david some minutes ago.

```javascript
      if (enabled !== false) {
        this.setState({activeTab: index})
      }
      if (allowHandleDisabledTabClick || enabled !== false) {
        this.props.handleChange(index, this.props.items[index])
      }
```
The other solutions would trigger unexpected behaviours

Additionally in my opinion `handleClickInDisabledTabs` would be a bit easier to understand than `allowHandleDisabledTabClick`.

just add some demo playground as @jcger remark and it would be amzing.
good job mate!

JoanClaret (approved):

devictoribero (approved):

devictoribero:
Well I blocked because of the comment that @miduga did. I wanted to make sure that @davidmartin84 read miguel angel's comment

PR #554, opened 2018-10-31
feat(atom/input): component prepared for MoleculeInputTags

juanmaguitar:
- Added `onFocus` & `onEnter`props
- Added mixins ready to be (re)used in other components

These additions are needed for the `MoleculeInputTags` component

Online version of these changes can be checked at → https://sui-components-atom-input-ready-tags.now.sh/workbench/atom/input/demo

It depends on the following PR's (merge this PR **only after** these ones have been merged):
- https://github.com/SUI-Components/sui-theme/pull/165
- https://github.com/SUI-Components/sui-theme/pull/164

### DIFF
--- a/components/atom/input/src/Input/Component/_mixins.scss
+++ b/components/atom/input/src/Input/Component/_mixins.scss
@@ -1,0 +1,16 @@
+@mixin sui-atom-input-input {
+  background: $c-white;
+  border: 1px solid $c-gray-light;
+  box-sizing: border-box;
+  font-size: $fz-base;
+  height: $h-atom-input--m;
+  padding-left: $p-l;
+  padding-right: $p-l;
+  width: 100%;
+}
+
+@mixin sui-atom-input-input-focus {
+  border: solid 1px $c-primary;
+  box-shadow: 0 0 3px 0 $c-primary;
+  outline: 0 none;
+}
--- a/components/atom/input/src/Input/Component/index.js
+++ b/components/atom/input/src/Input/Component/index.js
@@ -23,7 +23,7 @@ class Input extends Component {
     onChange && onChange({value, ev})
   }
 
-  hendleKeyDown = ev => {
+  handleKeyDown = ev => {
     const {onEnter, onEnterKey} = this.props
     const {key} = ev
     if (key === onEnterKey && onEnter) onEnter(ev)
@@ -78,7 +78,7 @@ class Input extends Component {
         onChange={this.changeHandler}
         onFocus={onFocus}
         onBlur={onBlur}
-        onKeyDown={this.hendleKeyDown}
+        onKeyDown={this.handleKeyDown}
         placeholder={placeholder}
         ref={reference}
         type={type}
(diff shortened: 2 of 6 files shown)


components/atom/input/src/Input/Component/_mixins.scss
```diff
@@ -0,0 +1,16 @@
+@mixin sui-atom-input-input {
+  background: $c-white;
+  border: 1px solid $c-gray-light;
```
jcger: we should have a sass var for the 1px

components/atom/input/src/Input/Component/_mixins.scss
```diff
@@ -0,0 +1,16 @@
...
+  font-size: $fz-base;
+  height: $h-atom-input--m;
+  padding-left: $p-l;
+  padding-right: $p-l;
+  width: 100%;
+}
+
+@mixin sui-atom-input-input-focus {
+  border: solid 1px $c-primary;
+  box-shadow: 0 0 3px 0 $c-primary;
```
jcger: this should also be a sui-theme var

components/atom/input/src/Input/Component/index.js
```diff
@@ -15,8 +15,18 @@ const ERROR_STATES = {
...
+    const {
+      target: {value}
+    } = ev
+    onChange && onChange({value, ev})
+  }
+
+  onEnterHandler = ev => {
+    const {onEnter} = this.props
+    const {key} = ev
+    if (key === 'Enter') onEnter && onEnter(ev)
```
jcger: you could do `if (onEnter && key === 'Enter')`

I know that sometimes the Tag is added pressing on comma, maybe the key should be a prop with Enter as default value or something like that

components/atom/input/src/Input/Component/_mixins.scss
```diff
@@ -0,0 +1,16 @@
...
+  outline: 0 none;
+}
```
ddmarin94: In my humble opinion, I think that this could do it with an inheritance of styles instead with a mixin. To me, a mixin is like a function and in this case you are not passing any parameters to this function for this reason I think that you could do inheritance. Like %sui-atom-input-input-focus  {} and declare your sass inside. Where you need to apply this styles you would use @extends instead of @includes. Just my opinion, nice job! 👍

ddmarin94 (dismissed):

danivm (dismissed):

carlosvillu (dismissed):

components/atom/input/src/Input/Component/index.js
```diff
@@ -15,8 +15,18 @@ const ERROR_STATES = {
...
-    onChange && onChange({value: ev.target.value, ev})
+  changeHandler = ev => {
+    const {onChange} = this.props
+    const {
+      target: {value}
+    } = ev
+    onChange && onChange({value, ev})
+  }
+
+  hendleKeyDown = ev => {
```
adriadoz: Should be handleKeyDown?

carlosvillu (approved):

adriadoz (approved):

PR #556, opened 2018-11-01
feat(molecule/buttonGroup): fix use case when false values are passed as childs

juanmaguitar:
So we can deal w/ uses like...

```
 <MoleculeButtonGroup>
          {isTherePrev && (<AtomButtom >Prev</AtomButtom>)}
          {range.map((page, index) => (<AtomButtom key={index}>{page}</AtomButtom>))}
          {isThereNext && (<AtomButtom>Next</AtomButtom>}
</MoleculeButtonGroup>
```

we need to preprare `MoleculeButtonGroup` to deal w/ `false` values as childs so `React.cloneElement` doesn't fail.
This PR fix this.

### DIFF
--- a/components/molecule/buttonGroup/src/index.js
+++ b/components/molecule/buttonGroup/src/index.js
@@ -25,10 +25,17 @@ const MoleculeButtonGroup = ({
     groupPositions,
     numChildren
   )
-  const extendedChildren = React.Children.map(children, (child, index) => {
-    const groupPosition = getGroupPositionByIndex(index)
-    return React.cloneElement(child, {...props, type, groupPosition, fullWidth})
-  })
+  const extendedChildren = React.Children.toArray(children)
+    .filter(child => !!child)
+    .map((child, index) => {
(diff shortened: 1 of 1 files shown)


danivm (dismissed):

components/molecule/buttonGroup/src/index.js
```diff
@@ -25,10 +25,17 @@ const MoleculeButtonGroup = ({
     groupPositions,
     numChildren
   )
-  const extendedChildren = React.Children.map(children, (child, index) => {
-    const groupPosition = getGroupPositionByIndex(index)
-    return React.cloneElement(child, {...props, type, groupPosition, fullWidth})
-  })
+  const extendedChildren = React.Children.toArray(children)
+    .filter(child => !!child)
```
carlosvillu: You could write this like `.filter(Boolean)` is the same but look like a little cleaner

carlosvillu (dismissed):

carlosvillu (approved):

danivm (approved):

PR #557, opened 2018-11-02
Button group improved alignment

juanmaguitar:


components/molecule/buttonGroup/src/index.scss
```diff
@@ -1,4 +1,6 @@
 .sui-MoleculeButtonGroup {
+  display: flex;
+
   &--fullWidth {
     display: flex;
```
midudev, line 5: I guess that if it's always display: flex, then maybe the property for the fullWidth is not needed anymore.

juanmaguitar: this is needed for its children with the class `&--fullWidth` (`AtomButton`'s mostly)

jcger (approved):

danivm (approved):

PR #558, opened 2018-11-06
Input no border prop

juanmaguitar:
Demo online (_no border_ section) at → https://sui-components-input-no-border-prop.now.sh/workbench/atom/input/demo

### DIFF
--- a/components/atom/input/src/Input/Component/index.js
+++ b/components/atom/input/src/Input/Component/index.js
@@ -35,12 +35,13 @@ class Input extends Component {
     return ''
   }
 
-  getClassNames({size, charsSize, hideInput, errorState}) {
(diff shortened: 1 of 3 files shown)


jcger (approved):

jcger:
I'm not sure about this at all as it is no design decision but ours. It should not be used as it is, just in justified cases as the tag input.

It would be nice to have this written down somewhere

juanmaguitar:
> I'm not sure about this at all as it is no design decision but ours. It should not be used as it is, just in justified cases as the tag input.
> 
> It would be nice to have this written down somewhere

Maybe we can just add a comment in the code w/ a link to your comment

carlosvillu (approved):

PR #561, opened 2018-11-09
feat(molecule/notification): add children prop, so we can print HTML

JoanClaret:
### Objective
Insert links in the notification content.

### Why
We need to reduce space used by this notification, so we decided to insert a link in the notification content, and do not use its default CTA buttons.

### How
Printing children if text prop does not exists, so this will not cause a major change.

### Example with CTA buttons
captura de pantalla 2018-11-08 a las 15 06 42

### Example without CTA buttons and using children
image

### DIFF
--- a/components/molecule/notification/src/index.js
+++ b/components/molecule/notification/src/index.js
@@ -99,13 +99,6 @@ class MoleculeNotification extends Component {
       .slice(0, BUTTONS_MAX)
       .map((button, i) => <Button key={i} {...button} />)
   }
-  printContent = () => {
-    const {text, children} = this.props
-    if (children) {
-      return children
-    }
-    return `<span>${text}</span>`
-  }
 
   render() {
     const {show, delay} = this.state
--- a/components/molecule/notification/src/index.js
+++ b/components/molecule/notification/src/index.js
@@ -107,10 +107,6 @@ class MoleculeNotification extends Component {
     return `<span>${text}</span>`
   }
 
-  createMarkup(markup) {
-    return {__html: markup}
-  }
-
   render() {
     const {show, delay} = this.state
     const {
--- a/components/molecule/notification/src/index.js
+++ b/components/molecule/notification/src/index.js
@@ -119,6 +119,10 @@ class MoleculeNotification extends Component {
         [`${CLASS}-effect--hide`]: effect && delay
       }
     )
+    const innerWrapperClassName = cx({
+      [`${CLASS}-children`]: children,
+      [`${CLASS}-text`]: text
+    })
 
     if (!show && !delay) {
       return null
@@ -130,7 +134,7 @@ class MoleculeNotification extends Component {
           <div className={`${CLASS}-iconLeft`}>
             <span className={`${CLASS}-icon`}>{icon || ICONS[type]}</span>
           </div>
-          <div className={`${CLASS}-text`}>{text || children}</div>
+          <div className={innerWrapperClassName}>{children || text}</div>
           {showCloseButton && (
             <div className={`${CLASS}-iconClose`} onClick={this.toggleShow}>
               <span className={`${CLASS}-icon`}>
@@ -159,9 +163,9 @@ MoleculeNotification.propTypes = {
    */
   buttons: PropTypes.array,
   /**
-   * Content to be printed. Alternatively you can use "text" prop
+   * Notification content
    */
-  children: PropTypes.node,
+  children: PropTypes.node.isRequired,
   /**
    * Transition enabled
    */
@@ -183,7 +187,7 @@ MoleculeNotification.propTypes = {
    */
   showCloseButton: PropTypes.bool,
   /**
-   * Content text. Alternatively you can use "children" prop
+   * Content text. Deprecated, use children instead.
    */
   text: PropTypes.string,
   /**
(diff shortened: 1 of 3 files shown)


davidmartin84 (dismissed):

components/molecule/notification/src/index.js
```diff
@@ -109,7 +109,8 @@ class MoleculeNotification extends Component {
       position,
       showCloseButton,
       effect,
-      text
+      text,
```
kikoruiz, line 112: `children` can be anything, also text, so... does it still make sense?

kikoruiz: I mean, prop `text` could be removed, so that you should create a major version.

components/molecule/notification/src/index.js
```diff
@@ -129,9 +130,7 @@ class MoleculeNotification extends Component {
           <div className={`${CLASS}-iconLeft`}>
             <span className={`${CLASS}-icon`}>{icon || ICONS[type]}</span>
           </div>
-          <div className={`${CLASS}-text`}>
-            <span>{text}</span>
-          </div>
+          <div className={`${CLASS}-text`}>{text || children}</div>
```
kikoruiz: Submodule classname `-content` would make more sense.

devictoribero: I agree with @kikoruiz

components/molecule/notification/src/index.js
devictoribero: Totally agree

components/molecule/notification/src/index.js
```diff
@@ -129,9 +130,7 @@ class MoleculeNotification extends Component {
           <div className={`${CLASS}-iconLeft`}>
             <span className={`${CLASS}-icon`}>{icon || ICONS[type]}</span>
           </div>
-          <div className={`${CLASS}-text`}>
-            <span>{text}</span>
```
midudev, line 133: Removing the `<span>` could break styling (or not). Just to be sure this doesn't affect anything.

components/molecule/notification/src/index.js
midudev: While I understand the comment, and it make sense, I would advise trying to not use a major for this. Keeping compatibility with previous versions and adding the `children` is enough for adding the feature and it will let projects to adopt this change seamlessly without having to move to a major.

I rather would prefer to use children as default and add a deprecated notice for the usage of `text` but do it as a minor still.

devictoribero: The idea of doing a major... is to clean up code. Of course the major is not needed and we could make it this changes as a minor. But IMHO, we are creating a _"technical debt"_ that we could solve doing the major now.

In addition, then we should add a `wrapper` for this `content` that renders the `children` because we should apply some spacing between `text` and the `content`. _"Too much"_ for evading to create a major in my opinion. What do you think guys? @miduga @kikoruiz @JoanClaret

midudev: The code to maintain and the technical debt is negligible against the lack of support for actual consumers of the component and the problems and work that this could originate. Creating a major version of a component without even offering a deprecation period is not great and it goes against the idea behind SUI Components for creating a stable set of components. Even that, this decision can't be based on two or three approves in a PR that only wanted to add a feature.

**If we create a major version now:**
- Projects that are already using this component will need to move to the latest version to get fixes and new features.
- If they want to not impact performance, they will need to move all components at once in order to avoid loading two versions of the same component.

IMHO we should be very careful with majors and iterate over actual versions as much as possible. **Does it mean are we not going to release a major never? No.** But we have to be responsible with this and only create majors when they're strictly necessary or after having a bag of deprecations after some time of working on them. The problem never should be creating technical debt, the problem would be not addressing never them.

**What we could do instead?**
- Deprecate the usage of the prop but maintaining compatibility with actual usages as this PR is proposing.
- Require the `children` prop in PropTypes and make `text` optional. Add in the description of `text` prop that the prop is deprecated.
- Share in our channels this deprecation notice and give info in order to move to the new version.
- Give a period of grace in order to users of the component to move to the new prop without hassle.
- If technical debt is produced, comment it over and create issues if necessary, explaining it in order to address in the next major.

devictoribero: Mhh.. You sold it right. I agree then that this could be the best solution for everyone

JoanClaret:
- Added `isRequired` in the `children` prop.
- Marked `text` prop as deprecated 
- Added `innerWrapperClassName` to manage main content wrapper class, so we use `.sui-MoleculeNotification-children` when a children in passed.

Next steps will be open an issue so we can delete the technical debt in the next major release.

Would you check it again? 🙏
@miduga @tureey @kikoruiz @davidmartin84   

Anyways, i would appreciate an agreement about when to release a major. maybe we should specify under which circumstances are we willing to create a major version.

midudev (approved):

jcger (approved):

devictoribero (approved):
I think it's OK! Good job! 👍 

I don't don't know how much time would be okay for the `text` prop deprecation. I have no experience in these things.

JoanClaret:
issue opened, merging PR.
https://github.com/SUI-Components/sui-components/issues/562

kikoruiz (approved):

PR #563, opened 2018-11-13
feat(molecule/notification): add color variables for fonts

beatrizip:
Added color variables for the fonts of the different types of notifications

### DIFF
--- a/components/molecule/notification/src/_settings.scss
+++ b/components/molecule/notification/src/_settings.scss
@@ -6,11 +6,26 @@ $sz-notification-icon: $sz-icon-m !default;
 // Colors
 
 $notification-type-colors: (
-  info: $c-primary,
-  success: $c-success,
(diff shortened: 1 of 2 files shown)


kikoruiz:
I have 2 notes regarding this request:

1. Is this something previously defined by UX?
2. Take care with this kind of changes because you're driving into a new major change, and it would be good to know first what is planned to discuss tomorrow in a global meeting: https://sch-chat.slack.com/archives/C06QTL3C1/p1542021880037100.

beatrizip:
@kikoruiz tx!

1) It has not been validated with UX because it does not add functionality or new use, in the case of ePreselec the primary color does not contrast with white fonts, it is necessary that we can configure the color of the font.

2) About the new version, I cancel the PR until after today's meeting, we will talk about it.

captura de pantalla 2018-11-14 a las 9 46 49

PR #572, opened 2018-11-22
feat(molecule/inputField): update AtomInput version to v2

juanmaguitar:
feat(molecule/inputField): updated version AtomInput to v2 & updated demo

⚠ This change contain `BREAKING_CHANGES` so it will generate a `MAJOR` version as it provokes a change in the API of the event handlers  (`onChange` and so) caused by this update of the `AtomInput` version to v2


jcger:
isn't this a major too?

jcger (approved):

midudev (approved):

PR #576, opened 2018-12-03
Molecule dropdown option enter focus

juanmaguitar:
Focus Styles and Select on Enter

### DIFF
--- a/components/molecule/dropdownOption/src/index.js
+++ b/components/molecule/dropdownOption/src/index.js
@@ -38,8 +38,17 @@ const MoleculeDropdownOption = ({
     )
   }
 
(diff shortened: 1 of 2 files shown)


components/molecule/dropdownOption/src/index.js
```diff
@@ -38,8 +38,17 @@ const MoleculeDropdownOption = ({
     )
   }
 
+  const handleKeyDown = e => {
+    if (e.key === 'Enter' && !disabled) onClick(e, {value})
```
jcger: where does the value come from?

jcger (approved):

adriadoz (approved):

PR #579, opened 2018-12-03
feat(molecule/notification): add color variables for fonts in negativ…

beatrizip:
…e cases

To fulfill text accessibility with our corporate colors we need different font colors

notifications

### DIFF
--- a/components/molecule/notification/src/_settings.scss
+++ b/components/molecule/notification/src/_settings.scss
@@ -13,7 +13,7 @@ $notification-type-colors: (
   system: $c-system,
 ) !default;
 
-$notification-type-color-font: (
+$notification-type-font-colors: (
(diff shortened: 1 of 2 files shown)


nguasch (dismissed):

AitorRodriguez990 (dismissed):

components/molecule/notification/src/_settings.scss
```diff
@@ -13,6 +13,14 @@ $notification-type-colors: (
   system: $c-system,
 ) !default;
 
+$notification-type-color-font: (
```
jcger: what about `$notification-type-font-colors`?

nguasch (approved):

jcger (approved):

PR #589, opened 2018-12-14
Molecule dropdown list inner ref

juanmaguitar:
Added innerRef so we can control this element and its children from parent components (`MoleculeSelect` for example)

### DIFF
--- a/demo/molecule/dropdownList/demo/hoc/withStateSingle.js
+++ b/demo/molecule/dropdownList/demo/hoc/withStateSingle.js
@@ -6,7 +6,9 @@ const BASE_CLASS_DEMO = 'DemoMoleculeDropdownList'
 const CLASS_DEMO_LIST = `${BASE_CLASS_DEMO}-list`
 
 const withStateSingle = BaseComponent => {
+  const displayName = BaseComponent.displayName
   return class BaseComponentWithState extends Component {
+    static displayName = `withStateSingle(${displayName})`
     state = {
       value: this.props.value
     }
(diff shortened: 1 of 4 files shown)


devictoribero (approved):

demo/molecule/dropdownList/demo/hoc/withStateSingle.js
```diff
@@ -23,12 +26,17 @@ const withStateSingle = BaseComponent => {
...
-      const {handleSelect, props} = this
+      const {handleSelect, innerRef, props} = this
       return (
         <div>
           <pre>{JSON.stringify(this.state)}</pre>
           <div className={CLASS_DEMO_LIST}>
-            <BaseComponent {...props} value={value} onSelect={handleSelect} />
+            <BaseComponent
+              {...props}
+              innerRef={innerRef}
```
jcger: wouldn't this work by just using regular react "ref" in order to avoid having to define innerRef in the DropdownList components?

jcger:
what do you want to do with this ref afterwards? we should avoid using ref for anything we could solve declaratively and therefore I would be against exposing it "as a feature"

juanmaguitar:
@jcger, The purpose of this task, is adding the behaviour of passing this _"reference manager"_ from a parent component so this parent component can access this `DropdownList` component to do things like `focus` its elements  

This is necessary to add keyboard support in the `MoleculeSelect` component, to (for example) focusing the first element of the list when `ArrowDown` is pressed first time the  `MoleculeSelect` is opened
 
To pass this _"reference manager"_ (`React.createRef()`) to this `DropdownList` from the outside, the best way is via props, don't you think?

juanmaguitar:
Found a more elegant solution for `MoleculeSelect` that doesn't require adding this `innerRef` to `MoleculeDropdownList`

PR #590, opened 2018-12-19
feat(atom/tag): normalize form of icons passed via props according to agreement

juanmaguitar:
Normalize way of passing icons to the component according to https://github.schibsted.io/scmspain/frontend-all-wiki/pull/48

Also added stopPropagation on `onClose` event

BREAKING CHANGE:
icons passed as components

- 🌎 Online Demo → https://sui-components-atom-tag-icons.now.sh/workbench/atom/tag/demo

### DIFF
--- a/components/atom/tag/src/Actionable/index.js
+++ b/components/atom/tag/src/Actionable/index.js
@@ -11,7 +11,7 @@ const getClassNames = function({className}) {
 }
 
 const ActionableTag = function({
-  icon: Icon,
+  icon,
   href,
   iconPlacement,
   label,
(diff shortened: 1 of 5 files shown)


jcger:
I pushed how it should be. The stateless component we mock in our playgrounds as icons must be treated like the es5 transpiled react component that our sui-svg tools spit out

At the end you will end up having something like
```
import Icon from 'my-svg-as-react-components-lib/lib/icon'

const MyApp = () =>
   <SomeSUIComponent icon={<Icon />} />
```
And the sui component must be able to work this way

Sorry, I had no time to talk with you in detail about it, I hope this example helps you.

juanmaguitar:
@jcger Oks, thanks for the commit.
I've added the proper notes to the WIKI so there's a clear reference there regarding this
https://github.schibsted.io/scmspain/frontend-all-wiki/pull/49

jcger (approved):

danivm (approved):

PR #619, opened 2019-01-29
Feature/atom input scss refactor

Osgar:
Added the call of the styles of the component to the readme and corrected the example of the form that called the button with b minuscule. 
Changes in the variables and in the use of input attributes by pseudo-classes and pseudo-elements.

### DIFF
--- a/components/atom/input/package.json
+++ b/components/atom/input/package.json
@@ -10,7 +10,7 @@
   },
   "dependencies": {
     "@s-ui/component-dependencies": "1",
-    "imask": "3.4.0"
+    "imask": "3.3.0"
   },
   "keywords": [],
   "author": "",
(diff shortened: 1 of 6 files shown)


components/atom/input/package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "@s-ui/react-atom-input",
-  "version": "3.3.0",
```
midudev: Don't update manually the version as it will be updated for you when releasing the new version of the component.

midudev (dismissed):

components/atom/input/package.json
Osgar: I left the package.json as it was. Thank you

midudev (approved):

jcger (approved):

devictoribero (approved):
I think it's a good improvement:tada:
Good job! :clap:

PR #623, opened 2019-02-07
feat(atom/label): prepare component to admit pass onClick handler

juanmaguitar:


jcger (approved):

jcger:
Don't you need to add a default empty onClick fn?

juanmaguitar:
@jcger Not in this case. As this handler is directly associated to the `onClick` prop of the proper element, it doesn't matter if it arrives as undefined. In this case it will do nothing
👉 https://codesandbox.io/s/jp48qmz1zw

danivm (approved):

PR #625, opened 2019-02-07
Select field

juanmaguitar:
- 🌍 Demo Online → https://sui-components-select-field.now.sh/workbench/molecule/selectField/demo


components/molecule/selectField/README.md
```diff
@@ -0,0 +1,23 @@
+# MoleculeSelectField
+
+> Description
```
jcger: please fill the readme

juanmaguitar: Done! 👍
Also reviewed and improved the `MoleculeSelect` README → https://github.com/SUI-Components/sui-components/pull/627

components/molecule/selectField/src/index.js
```diff
@@ -0,0 +1,80 @@
...
+
+  /** Error message to display when error state  */
+  errorText: PropTypes.string,
+
+  /** Help Text to display */
+  helpText: PropTypes.string,
+
+  /** Boolean to decide if elements should be set inline */
+  inline: PropTypes.bool
+}
```
devictoribero, line 78: Would be good to add this to the `demo`. Could we? I would like to see how this affects the UI of the component

components/molecule/selectField/src/index.js
```diff
@@ -0,0 +1,80 @@
...
+    const {
+      label,
+      successText,
+      errorText,
+      helpText,
+      inline,
+      children, // eslint-disable-line
+      ...props
+    } = this.props
+    const errorState = getErrorState(successText, errorText)
```
devictoribero: I had to go to the specification of the `function` to know what this does.

Due to the function returns `true` or `false` depending on if it the components has an errors text, what do you think about naming it `hasErrors` or `hasErrorMessage`?

components/molecule/selectField/src/index.js
juanmaguitar: This `prop` is not really part of the final component specification. I decided to let it just in case there's some need of it in some projects but from a UX point of view it seems it's not a recommended way of using these components in a form.

Having said that, I'll add this use case (i also think this use case should be in the demo)

components/molecule/selectField/src/index.js
juanmaguitar: You're right. `hasErrors` it is

demo/molecule/selectField/playground
```diff
@@ -0,0 +1,94 @@
+return (<MoleculeSelect />)
+/*
+const withState = BaseComponent => {
```
jcger: shouldn't this be removed?

components/molecule/selectField/README.md
```diff
@@ -0,0 +1,114 @@
...
+
+## Usage
+
+After importing the component `MoleculeSelectField` like this
+
+```javascript
+import MoleculeSelecyField from '@s-ui/react-molecule-select-field'
+import MoleculeSelectOption from '@s-ui/react-molecule-dropdown-option'
+
+const IconCloseTag = () => <span>x</span>
```
jcger, line 21: Not using this, looks like you forgot to add the required closeIcon in all examples

components/molecule/selectField/src/index.js
```diff
@@ -0,0 +1,80 @@
...
+import MoleculeSelect from '@s-ui/react-molecule-select'
+
+const hasErrors = (success, error) => {
+  if (error) return true
+  if (success) return false
+}
+
+class MoleculeSelectField extends Component {
+  refSelect = React.createRef()
+  id = this.props.id || `MoleculeSelect-${+new Date()}`
```
jcger: looks like we are not automagically settings ids in other components so I would remove this feature and add `isRequired` to the id propType in order to keep consistency across components. 

Also, developers are kind of forced to control the id in order to make it work with Formik or a custom form manager, so please remove it and just make it required

jcger (changes requested):

components/molecule/selectField/src/index.js
jcger: as this shouldn't have been implemented because design hasn't define this behaviour, we will let the prop as it is but without adding an example to our demos and creating a tech debt issue in order to remove it as soon as possible

components/molecule/selectField/README.md
juanmaguitar: There was no example with multiselection mode that is the only _mode_ that requires this icon. Added this use case!

demo/molecule/selectField/playground
juanmaguitar: Removed!

components/molecule/selectField/README.md
```diff
@@ -0,0 +1,114 @@
...
+```sh
+$ npm install @s-ui/react-molecule-select-field --save
+```
+
+## Usage
+
+After importing the component `MoleculeSelectField` like this
+
+```javascript
+import MoleculeSelecyField from '@s-ui/react-molecule-select-field'
```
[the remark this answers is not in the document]
juanmaguitar: Fixed!

components/molecule/selectField/src/index.js
juanmaguitar: Updated as `isRequired` prop

jcger (approved):

adriadoz (approved):

PR #629, opened 2019-02-12
feat(molecule/select): readme reviewed

juanmaguitar:


devictoribero (approved):

components/molecule/select/README.md
```diff
@@ -151,10 +151,8 @@ import MoleculeSelectOption from '@s-ui/react-molecule-dropdown-option'
 import withDynamicOptions from './hoc/withDynamicOptions'
 import {withStateValue} from '@s-ui/hoc'
 
-// some function that gets a `{query}` and returns asynchronoudly a list of values
-import {getAsyncCountriesFromQuery} from './services'
-
 const MoleculeSelectWithState = withStateValue(MoleculeSelect)
+const options = ['John','Paul','George','Ringo']
```
devictoribero, line 155: Is it mandatory to be an array of strings?
If that's the case, could be good to add `PropTypes.arrayOf(PropTypes.string)`
If it's not mandatory, It could be good to add the shape

rmoralp (approved):

components/molecule/select/README.md
juanmaguitar: Well, options is not even a prop of any component so there's no need of setting any `propType`. The component `MoleculeSelect` is flexible enough to admit a a set of option as `children` built from any type of structure (array, objects,...)

devictoribero: my bad @juanmaguitar

PR #630, opened 2019-02-12
feat(molecule/field): inline takes 100% of width available

juanmaguitar:


devictoribero (approved):
For me it's a bit strange that the `inline prop` makes the components act as a `block`.

devictoribero (commented):
missclick on approve

devictoribero (changes requested):
I thought that clicking comment would dismiss the approve 😆

juanmaguitar:
@tureey That's right. But this `inline` is not a CSS `inline`, it's a name for the prop that will display the elements in a line. And these components are supposed to take the 100% of the container, so we can decide the width of the field, by setting the width of the container

What do you think? Maybe another name for the prop would be less confusing?

devictoribero:
Yes it makes sense now @juanmaguitar!

I can't find a name for that.. :( if anyone has any idea... share it please! If not, let it as it's now

devictoribero (approved):

rmoralp (approved):

jcger:
Isn't inline the default described by ux? In this case, the default value should be true

No demo example?

juanmaguitar:
@jcger If I remember correctly, I was asked by UX specifically to remove these inline examples of the demo because they were no part of the specification
Anyway, as we're adding this use case for `fieldSelect` it would have sense to also add this use case to `fieldInput` and `fieldTextarea`

PR #633, opened 2019-02-14
feat(examples): added inline & form use case

juanmaguitar:
Added a few more use cases:
- `inline` mode
- `inline` mode w/ button (typical form in one line)

This PR is dependant on this other one https://github.com/SUI-Components/sui-components/pull/632 so no merge should be done here until this other one is fully approved and merged


midudev (dismissed):

juanmaguitar:
Accoding w/ https://github.com/SUI-Components/sui-components/pull/625#discussion_r256809352 we've decided not to add inline examples (for now)

PR #637, opened 2019-02-18
Add border radius to input/button

igncalleja:
This PR adds support for border-radius property in atom/input and atom/button. 

Although is not a breaking change because the default initial value stays as `0`, it could cause issues for users who relay on specifity/lack of prop setting to create their own styles.

Also, we are aware this implementation will not work correctly with the current version of the "addons". Users that may want to use border-radius AND addons will have to set specific styles for the corners they don't want to see rounded.

screenshot 2019-02-18 at 16 32 18

### DIFF
--- a/components/atom/input/src/Input/Component/index.scss
+++ b/components/atom/input/src/Input/Component/index.scss
@@ -2,13 +2,15 @@ $bgc-atom-input-disabled: $c-gray-lightest !default;
 $c-atom-input-placeholder: $c-gray-dark !default;
 $m-atom-input-type-left: $m-m !default;
 $m-atom-input-type-right: $m-m !default;
+$bdrs-atom-input: 0 !default;
 
 $base-class: '.sui-AtomInput-input';
 $class-read-only: '#{$base-class}--readOnly';
 
 #{$base-class} {
   @extend %sui-atom-input-input;
-
+  border-radius: $bdrs-atom-input;
+  
   &--size {
     width: inherit;
   }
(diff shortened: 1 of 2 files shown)


carlosvillu (dismissed):

components/atom/input/src/Input/Component/index.scss
```diff
@@ -2,13 +2,15 @@ $bgc-atom-input-disabled: $c-gray-lightest !default;
 $c-atom-input-placeholder: $c-gray-dark !default;
 $m-atom-input-type-left: $m-m !default;
 $m-atom-input-type-right: $m-m !default;
+$bdrs-atom-input: 0 !default;
```
jcger: please use `$bdrs-none` instead of 0 https://github.com/SUI-Components/sui-theme/blob/master/src/settings/_box-style.scss#L4

jcger (changes requested):

jcger (approved):

components/atom/input/src/Input/Component/index.scss
igncalleja: Done! 

What's the advantage of using `$bdrs-none`? Is there any scenario where `$bdrs-none` is going to be different from 0?

tomasmax (approved):

components/atom/input/src/Input/Component/index.scss
jcger: I would say it's to have a common language with UX, we also have stuff like c-white

rmoralp (approved):

devictoribero (approved):

PR #643, opened 2019-02-19
SUIC-106 fix(atom/input): fix min-height setting in atom input

igncalleja:
This PR fixes a bug with the atom-input size ([https://jira.scmspain.com/browse/SUIC-106](https://jira.scmspain.com/browse/SUIC-106)) where the "small" size is not correctly applying its styles.

Although the best approach would have been modifying in SUI-theme the placeholder `%sui-atom-input-input` used, that would have been a breaking change, so I have opted for the more conservative approach of resetting the min-height for the component and declaring it for every size modifier.

### DIFF
--- a/components/atom/input/src/Input/Component/index.scss
+++ b/components/atom/input/src/Input/Component/index.scss
@@ -9,6 +9,7 @@ $class-read-only: '#{$base-class}--readOnly';
 
 #{$base-class} {
   @extend %sui-atom-input-input;
+  min-height: auto;
   border-radius: $bdrs-atom-input;
   
   &--size {
@@ -60,6 +61,7 @@ $class-read-only: '#{$base-class}--readOnly';
   @each $type, $attr in $sizes-atom-input {
     &-#{$type} {
       height: $attr;
+      min-height: $attr;
     }
   }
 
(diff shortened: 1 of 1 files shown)


jcger (approved):

carlosvillu (approved):

midudev (approved):

components/atom/input/src/Input/Component/index.scss
```diff
@@ -60,6 +61,7 @@ $class-read-only: '#{$base-class}--readOnly';
   @each $type, $attr in $sizes-atom-input {
     &-#{$type} {
       height: $attr;
+      min-height: $attr;
```
kikoruiz, line 64: Strange... Shouldn't work only with the `height` attribute?

kikoruiz (approved):

tomasmax (approved):

components/atom/input/src/Input/Component/index.scss
igncalleja: Yes, it should, I'm keeping the `min-height` for compatibility reasons, just-in-case anyone relies on it.

However, as a general rule I don't think it's a good idea setting a fixed height for text/form elements. We should play only with line-height and padding, but inputs are a bit special with line-heights...

PR #655, opened 2019-03-26
feat(molecule/tabs): fix issue in demo

juanmaguitar:
- ⚙️ Task → https://jira.scmspain.com/browse/SUIC-118


demo/molecule/tabs/playground
```diff
@@ -1,189 +1 @@
...
-              items={itemsNoIcons}
-              activeTab={0}
-              handleChange={handleChange}
-              handleClickInDisabledTabs
-          />
-          <br/><br/>
-        </div>
-    </div>
-)
+return <p>This shouldn't be displayed</p>
```
midudev, line 1: ¿?

juanmaguitar: Because, the folder `demo` should take control. I've noticed in some cases, when there's an error in `demo` this `playground` content is displayed
So this is like a quick _warning_ that something goes wrong

jcger (approved):

components/molecule/tabs/src/hoc/withStateActiveTab.js
```diff
@@ -24,7 +24,7 @@ const withStateActiveTab = BaseComponent => {
     }
 
     componentDidMount() {
-      const {children} = this.prop
+      const {children} = this.props // eslint-disable-line
```
midudev, line 27: Why is that line disabled? Because PropTypes check? We might want to just add PropTypes instead disabling it... 🤔

midudev (approved):

PR #670, opened 2019-04-18
Atom tag - version 1.2

rmoralp:
**Including the Large size.** 
image

**40 px clickable area of close icon**
Icon hover
image

Icon hover area
image

Icon clickable area:
image
image
image

[Paper](https://paper.dropbox.com/doc/SUI-Tags--AbdM4X3X~lteJFBLC_u3utMFAg-bOSL4L5TFYZqPaK0wnCHb)

### DIFF
--- a/components/atom/tag/src/Standard.js
+++ b/components/atom/tag/src/Standard.js
@@ -17,8 +17,8 @@ const StandardTag = ({className, closeIcon, icon, label, onClose}) => {
         {label}
       </span>
       {closeIcon && (
-        <span className="sui-AtomTag-secondary-closeable" onClick={handleClick}>
-          <span className="sui-AtomTag-secondary-closeableIcon sui-AtomTag-secondary-icon">
+        <span className="sui-AtomTag-closeable" onClick={handleClick}>
+          <span className="sui-AtomTag-closeableIcon sui-AtomTag-secondary-icon">
             {closeIcon}
           </span>
         </span>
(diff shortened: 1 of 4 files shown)


components/atom/tag/src/Standard.js
```diff
@@ -17,11 +17,10 @@ const StandardTag = ({className, closeIcon, icon, label, onClose}) => {
         {label}
       </span>
       {closeIcon && (
-        <span
-          className="sui-AtomTag-secondary-closeable sui-AtomTag-secondary-icon"
-          onClick={handleClick}
-        >
-          {closeIcon}
+        <span className="sui-AtomTag-secondary-closeable" onClick={handleClick}>
```
kikoruiz: I know this classname was added previously, but I don't understand the concept `secondary` here, is it a modifier or another type of tag? Anyway, classname is not fitting well the SUIT CSS convention, it should be `sui-AtomTagSecondary-closeable`, although for me it would be better have these classnames: `sui-AtomTag-closeable` as parent and then `sui-AtomTag-closeableIcon`.

rmoralp: A tag could have 2 icons. Left hand side icon container has "sui-AtomTag-icon" class and right hand side icon container has "sui-AtomTag-secondary-icon" class

rmoralp: I have changed "sui-AtomTag-secondary-closeable" to "sui-AtomTag-closeable"

kikoruiz (approved):

adriadoz (approved):

components/atom/tag/src/Standard.js
rmoralp: Issue created to manage this https://github.com/SUI-Components/sui-components/issues/674

PR #671, opened 2019-04-18
feat(atom/switch): Allow value prop

AitorRodriguez990:
## Description

`AtomSwitch` component reads `initialValue` prop once but I need to load a value depending on an API request.

**Now the toggle value can be managed from its parent component** using the new `value` prop I added.

🎉 This version is backwards compatible with the existing version. 

## Tasks

- [x] Add new `value` prop.
- [x] Apply logic to read `value` prop.

## Demo

Grabaciondepantalla201904

### DIFF
--- a/components/atom/switch/README.md
+++ b/components/atom/switch/README.md
@@ -22,15 +22,16 @@ import AtomSwitch from '@s-ui/react-atom-switch'
 
 return (
   <AtomSwitch
-    initialValue={false}
-    size='default'
-    type='toggle'
     disabled={false}
+    initialValue={false}
     label='Label'
     labelLeft='Off'
-    labelRight='On'
     labelOptionalText='Optional label'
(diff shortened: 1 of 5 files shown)


juanmendez90 (dismissed):

adriadoz (dismissed):

juanmaguitar:
Can you please add a Use Case in the demo displaying an example where this new prop is used?

AitorRodriguez990:
@juanmaguitar Demo added :)

midudev:
Let's add some comments and more useful info in the README in order to explain that this component could be `uncontrolled` by using the `defaultValue` and `controlled` by using the `value`prop (and thus required to handle the state in order to send it over and get the UI updated).

midudev (approved):

kikoruiz (approved):

PR #676, opened 2019-04-22
Feature/data counter

juanmaguitar:
- ⚙️Task → https://jira.scmspain.com/browse/SUIC-50
- 📄Paper → https://paper.dropbox.com/doc/SUI-Data-counter-TqR9qBw4WRr5l5gNMvvEE
- 🌍Online Demo → https://sui-components-feature-data-counter.now.sh/workbench/molecule/dataCounter/demo


components/molecule/dataCounter/package.json
```diff
@@ -0,0 +1,23 @@
...
+  "main": "lib/index.js",
+  "scripts": {
+    "build": "npx rimraf ./lib && npx mkdirp ./lib && npm run build:js && npm run build:styles",
+    "build:js": "../../../node_modules/.bin/babel --presets sui ./src --out-dir ./lib",
+    "build:styles": "../../../node_modules/.bin/cpx './src/**/*.scss' ./lib"
+  },
+  "dependencies": {
+    "@s-ui/component-dependencies": "1",
+    "@s-ui/hoc": "^1.18.0",
+    "@s-ui/react-atom-input": "^3.9.0",
```
jcger: please remove the caret and use major version for our packages

components/molecule/dataCounter/src/index.js
```diff
@@ -0,0 +1,134 @@
...
+        <div
+          className={cx(
+            CLASS_INPUT_CONTAINER,
+            `${CLASS_INPUT_CONTAINER}--${size}`
+          )}
+        >
+          <AtomButton
+            disabled={disabled || value === min}
+            onClick={decrementValue}
+            type="secondary"
```
jcger: you should use a constant for both button type props are it should be always the same

components/molecule/dataCounter/src/index.js
```diff
@@ -0,0 +1,134 @@
...
+          >
+            -
+          </AtomButton>
+          <AtomInput
+            id={id}
+            disabled={disabled}
+            size={size}
+            charsSize={charsSize}
+            placeholder={placeholder}
+            value={value}
```
jcger: input propType is string or bool but this is a number https://github.com/SUI-Components/sui-components/blob/master/components/atom/input/src/index.js#L88

components/molecule/dataCounter/src/index.js
```diff
@@ -0,0 +1,143 @@
...
+
+  const incrementValue = e => {
+    if (isBelowMaxValue) onChange(e, {value: value + 1})
+  }
+
+  const decrementValue = e => {
+    if (isOverMinValue) onChange(e, {value: value - 1})
+  }
+
+  const handleKeyDown = e => {
```
rmoralp: Copied from Paper: "Input field value can also be changed by the user using the keyboard so Input field must receive focus as well as the buttons, but only number values are allowed."

Should numbers keys have to be allowed? I think that should be

juanmaguitar: Done!

components/molecule/dataCounter/src/index.js
```diff
@@ -0,0 +1,196 @@
...
+            -
+          </AtomButton>
+          <AtomInput
+            id={id}
+            disabled={disabled}
+            size={size}
+            charsSize={charsSize}
+            value={internalValue}
+            onKeyDown={handleKeyDown}
+            onChange={handleKeyDown}
```
rmoralp: onChange I think that is doesn't needed here. It already has onKeyDown

components/molecule/dataCounter/src/index.js
```diff
@@ -0,0 +1,196 @@
...
+
+  const addDigit = (e, {key: digit}) => {
+    const newValue = internalValue + digit
+    if (newValue.length <= 2) {
+      setInternalValue(newValue)
+      onChange(e, {value: newValue})
+    }
+  }
+
+  const handleKeyDown = e => {
```
rmoralp: Why not using {value} coming from AtomInput?
I can't add a digit before other. It always is being added at the end

jcger (dismissed):

components/molecule/dataCounter/src/index.js
juanmaguitar: Right! I have improved both the code and the UX based on your comments

components/molecule/dataCounter/src/index.js
```diff
@@ -67,28 +65,19 @@ const MoleculeDataCounter = ({
...
-
-  const addDigit = (e, {key: digit}) => {
-    const newValue = internalValue + digit
-    if (newValue.length <= 2) {
-      setInternalValue(newValue)
-      onChange(e, {value: newValue})
+  const handleChange = (e, {value}) => {
+    const numericValue = parseInt(value, 10)
+    if (value.length <= 2 && !isNaN(numericValue)) {
+      const strNewValue = String(numericValue)
```
rmoralp: This 3 lines are repeated 3 times in`incrementValue`, `decrementValue` and here `handleChange`. What do you think about wrap it in a function and reutilize this code?

juanmaguitar: done!

rmoralp (dismissed):

components/molecule/dataCounter/src/index.js
```diff
@@ -0,0 +1,185 @@
...
+  const incrementValue = e => {
+    if (isBelowMaxValue) {
+      const nValue = internalValue === '' ? min : parseInt(internalValue) + 1
+      assignValue(e, {nValue})
+    }
+  }
+
+  const decrementValue = e => {
+    if (isOverMinValue) {
+      const nValue = internalValue === '' ? min : parseInt(internalValue) - 1
```
jcger, line 65: you are repeating this code, you should move it into its own function

components/molecule/dataCounter/src/index.js
```diff
@@ -0,0 +1,185 @@
...
+            value={internalValue}
+            onKeyDown={handleKeyDown}
+            onChange={handleChange}
+          />
+          <AtomButton
+            disabled={incrementDisabled}
+            onClick={incrementValue}
+            type="secondary"
+          >
+            +
```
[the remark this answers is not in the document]
juanmaguitar: I don't know. Should they?

jcger (approved):

rmoralp (approved):

PR #682, opened 2019-04-24
feat(molecule/select): added disabled property

AgonisticKatai:
image

> ### Applies to MoleculeSelectSingleSelection component only

### DIFF
--- a/components/molecule/select/src/index.js
+++ b/components/molecule/select/src/index.js
@@ -13,6 +13,7 @@ import {getCurrentElementFocused} from '@s-ui/js/lib/dom'
 
 const BASE_CLASS = `sui-MoleculeSelect`
 const CLASS_FOCUS = `${BASE_CLASS}--focus`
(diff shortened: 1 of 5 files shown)


components/molecule/select/src/index.js
```diff
@@ -13,6 +13,7 @@ import {getCurrentElementFocused} from '@s-ui/js/lib/dom'
 
 const BASE_CLASS = `sui-MoleculeSelect`
 const CLASS_FOCUS = `${BASE_CLASS}--focus`
+const CLASS_DISABLED = `${BASE_CLASS}--disabled`
```
kikoruiz: Disabled state should be treated as state classname following the SUIT conventions: `.is-disabled`

AgonisticKatai: DONE!

kikoruiz (approved):

jcger (approved):

JoanClaret (approved):

midudev (approved):

PR #685, opened 2019-04-25
Atom tag responsive functionality

rmoralp:
MoleculeSelect uses InputTags and this one uses Tags.
In MoleculeSelect and also in InputTags, the tags should be displayed as "large" in mobile in order to keep the best possible UX getting a bigger "close button clickable area"

### DIFF
--- a/components/atom/tag/src/index.scss
+++ b/components/atom/tag/src/index.scss
@@ -11,7 +11,6 @@ $p-atom-tag-l: 0 $p-l !default;
 $p-atom-tag-m: 0 $p-l !default;
 $p-atom-tag-s: 0 $p-m !default;
 $w-atom-tag-clickable: 32px !default;
-$mq-atom-tag-responsive-breakpoint: 's' !default;
 
 @mixin icon-atom-tag($type) {
   @include sui-icon--small;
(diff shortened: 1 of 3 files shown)


JoanClaret (dismissed):

AgonisticKatai (dismissed):

components/atom/tag/src/index.scss
```diff
@@ -10,6 +11,7 @@ $p-atom-tag-l: 0 $p-l !default;
 $p-atom-tag-m: 0 $p-l !default;
 $p-atom-tag-s: 0 $p-m !default;
 $w-atom-tag-clickable: 32px !default;
+$mq-atom-tag-responsive-breakpoint: 's' !default;
```
jcger: Did ux define that this should be for every breakpoint? I mean, this may break the coherence between other components. You should be using hardcoded 's'  so everyone uses its 's' breakpoint

jcger (changes requested):

kikoruiz (dismissed):

components/atom/tag/src/index.scss
jcger: in which paper is this defined?

rmoralp: You are right. I talk about it with Marc Garrido and moved 's' value to the media mixin

JoanClaret (approved):

jcger (approved):

PR #688, opened 2019-04-25
fix(molecule/textareaField): fix textareaField missing first event value

AgonisticKatai:
text-area-field-bug

### Without getDerivedStateFromProps
text-area-field

### DIFF
--- a/components/molecule/textareaField/src/hoc/WithCharacterCount.js
+++ b/components/molecule/textareaField/src/hoc/WithCharacterCount.js
@@ -9,11 +9,6 @@ const WithCharacterCount = BaseComponent => {
       messageAtomTextarea: ''
     }
 
-    static getDerivedStateFromProps(nextProps) {
-      if (nextProps.value === '') return {value: ''}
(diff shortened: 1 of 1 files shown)


juanmaguitar:
This `getDerivedStateFromProps` is needed for the case when we submit a form and we need to _reset_ the values of the inputs inside of it, while it continues working every internal state of every component with its own stuff (count characters and so)

AgonisticKatai:
Ok @juanmaguitar,thx. I don't understand the answer very well but I see that the method is necessary (I already imagined that if it was for something 🤓). I just wanted to put on record that it is generated by the getDerivedStateFromProp method, which causes unwanted behavior when entering the first value in the MoleculeTextareaField component when wrapped by WithCharacterCount. I have tried to solve it but I don't find the key. Do you have any idea where the solution might come from?

juanmaguitar (changes requested):
Issue solved in https://github.com/SUI-Components/sui-components/pull/690
You can close this PR

PR #695, opened 2019-05-08
feat(molecule/collapsible): Do hideText prop to optional

AitorRodriguez990:
## Description

I think `hideText` prop is not necessary because we may not want to show it. 

In addition to this, I check `collapsed && showText || hideText` to render the action button and not drawing an empty button. This means that if the button has not `showText` or `hideText`, the component will not be rendered.

### Before

molecule-collapsible-before

### After

molecule-collapsible-after


AitorRodriguez990:
My team's UX will move it into the UXers team 👍

jcger:
Would it be ok to close this until its a thing?

PR #700, opened 2019-05-16
Feature/molecule table

juanmaguitar:
- 📄 Paper → https://paper.dropbox.com/doc/SUI-Table-wscpL60x6o3Pma5A56Mgn
- ⚙︎Task → https://jira.scmspain.com/browse/SUIC-52
- 🌎 Demo Online → https://sui-components-feature-molecule-table.now.sh/workbench/molecule/table/demo

✅Ok from UX → https://jira.scmspain.com/browse/SUIC-52?focusedCommentId=826114&page=com.atlassian.jira.plugin.system.issuetabpanels:comment-tabpanel#comment-826114


components/molecule/table/src/index.js
```diff
@@ -0,0 +1,127 @@
...
+  /** minimal value in range */
+  dataSource: PropTypes.object,
+
+  /** minimal value in range */
+  columns: PropTypes.object,
+
+  /** component including actions over rows */
+  actions: PropTypes.node,
+
+  /** component including actions over rows */
```
adriadoz, line 114: Definition should be for mobile prop

components/molecule/table/src/index.scss
```diff
@@ -0,0 +1,116 @@
...
+    font-size: $fz-table-title;
+  }
+
+  thead tr th {
+    color: $c-table-th;
+    font-size: $fz-table-thead-cell;
+    font-weight: normal;
+  }
+
+  thead tr td {
```
adriadoz, line 74: Avoid using element selectors if possible

components/molecule/table/src/index.js
```diff
@@ -0,0 +1,127 @@
...
+  ...props
+}) => {
+  const columns = actions
+    ? [...originalColumns, getActionsHeaderConfig(actions)]
+    : originalColumns
+
+  return (
+    <div className={BASE_CLASS}>
+      {mobile ? (
+        renderMobileMode({title, dataSource, columns})
```
adriadoz, line 81: Why mobile mode depends on a prop? Shouldn't be the component itself "auto" responsive?

adriadoz (commented):
There are two requirements in the paper that I couldn't find in the demo:
- A selector to change the number of rows displayed
- Rows have to be focusable

components/molecule/table/src/index.js
juanmaguitar: Not really. Because of some situations when doing SSR it was agreed w/ the enablers that a better strategy is to delegate to `LayoutMediaQuery` all the responsibility of detecting the size of the viewport so we can display the right version of any component via props
Maybe @jcger @carlosvillu or @midudev can give you more info about this

juanmaguitar:
@adriadoz regarding your comments about the specifications of the paper reflected in the component, this implementation has already been approved by the UX owner 
It was agreed w/ UX that a better solution was to create 2 component: `MoleculeTable` (the one in this PR) and another one `MoleculeFieldTable` (still in development) which will contain all the "extra" components that should go along w/ a table as described in the paper

components/molecule/table/src/index.js
```diff
@@ -0,0 +1,127 @@
...
+
+const renderMobileMode = ({title, dataSource, columns}) => { // eslint-disable-line
+  return (
+    <div className={CLASS_MOBILE_MODE}>
+      <h2>{title}</h2>
+      {dataSource.map((row, indexRow) => {
+        return (
+          <div key={indexRow} className={CLASS_MOBILE_MODE_ROW}>
+            {Object.keys(row).map((keyTitle, index) => {
+              const column = columns.filter(
```
midudev:
```
const column = columns.find(column => column.key === keyTitle)
```

components/molecule/table/src/index.js
```diff
@@ -0,0 +1,127 @@
...
+                column => column.key === keyTitle
+              )[0]
+
+              let dataCell = ''
+              if (typeof row[keyTitle] === 'object' && column.render) {
```
midudev: Extract to a function:
```js
getDataCell({ column, rowContent }) {
  return typeof rowContent === 'object' && column.render
    ? column.render(rowContent)
    : rowContent 
}

// ...
getDataCell({ column, rowContent: row[keyTitle]})
```

juanmaguitar: Done! Love your feedback. Suggestion and code ready to use 😁🔝

components/molecule/table/src/index.js
juanmaguitar: 👍

PR #705, opened 2019-05-31
feat(molecule/breadcrumb): removed deprecated dependency

juanmaguitar:
removed deprecated dependency as explained here → https://jira.scmspain.com/browse/SUIC-81


midudev (dismissed):

components/molecule/breadcrumb/src/utils.scss
```diff
@@ -0,0 +1,50 @@
+$breakpoints: (
```
midudev: Mmmm I think breakpoint-up is still valid in the next sui-theme: https://github.com/SUI-Components/sui-theme/blob/cf07c47c68541bdb8a2f550235e96fe52bd34d83/src/utils/_breakpoints.scss#L18

Why exactly do you need this utils file?

juanmaguitar: I'm getting this error if I remove the utils file

Captura de pantalla 2019-06-03 a las 14 47 23

I've put in this file all functions needed for the current code so I can remove the dependency without errors

jcger: you cannot use media-breakpoint-`down`, this task is about refactor in order to use media-breakpoint-`up` and make the component mobile first

juanmaguitar: oks, fixed!

JoanClaret (approved):

midudev (approved):

jcger (approved):

PR #712, opened 2019-06-18
feat(atom/input): add new native props

davidmartin84:


jcger (dismissed):

JoanClaret (dismissed):

components/atom/input/src/Input/Component/index.js
```diff
@@ -69,7 +69,9 @@ class Input extends Component {
       type,
       value,
       charsSize,
-      tabIndex
+      tabIndex,
+      maxlength,
```
midudev: I think could be a good idea to follow the camelCase rule for these props as well. `maxLength` and `autoComplete` as then for the `input` is needed like this and thus make it more sense.

davidmartin84: done!

juanmaguitar (approved):

midudev (approved):

PR #716, opened 2019-06-19
feat(molecule/accordion): add variable icon size

davidmartin84:


adriadoz (approved):

components/molecule/accordion/src/index.scss
```diff
@@ -110,7 +110,7 @@
     svg {
       fill: currentColor;
       transform: scale(1);
-      width: 10px;
+      width: $sz-accordion-tab-icon;
```
jcger, line 113: shouldn't this be `$w-accordion-tab-icon` as it is the width¿

jcger (approved):

PR #722, opened 2019-06-26
Feat/sui atom checkbox

a-ferrante:


jcger (approved):

juanmaguitar:
Is there a paper for this new component? What is then the role of the types `RADIO` & `CHECKBOX` in `AtomInput` → https://sui-components.now.sh/workbench/atom/input/demo ?

Captura de pantalla 2019-06-27 a las 8 25 29

a-ferrante:
Hi @juanmaguitar, here's the link to the UX-Spec document:
https://github.com/SUI-Components/UX-Definitions/blob/master/Atom/Checkbox-and-Radiobutton.md

I was not around when the discussion around inputs happened so i cannot provide input on the other question but the definition was already there, it just seems like the component wasn't implemented yet.

components/atom/checkbox/src/index.scss
```diff
@@ -0,0 +1,33 @@
+@import '~@schibstedspain/sui-theme/lib/index';
+
+$maw-atom-checkbox: 32px !default;
+
+.react-AtomCheckbox {
```
rmoralp, line 5: All the sui-components has `sui` prefix. Why are you using `react` prefix?

rmoralp:
Why look&feel does not match with paper?
Demo:
image

Paper:
image

components/atom/checkbox/src/index.js
```diff
@@ -0,0 +1,50 @@
+import React, {Fragment} from 'react'
+import PropTypes from 'prop-types'
+
+const AtomCheckbox = ({id, value, disabled, checked, onChange, ...props}) => {
```
rmoralp, line 4: I think that onChange should be required or just has an empty arrow function as a default value

rmoralp (changes requested):
@jcger will review with paper

JoanClaret:
as @juanmaguitar  is pointing out, since this component is already developed i don't get why to develop it again instead of use or improve the existing one.

a-ferrante:
Based on the discussions held i will close this PR and we will revisit the inclusion of this components later once Arturo is back.

PR #725, opened 2019-06-27
Feature/atom button iteration

danivm:
[SUIC-128](https://jira.scmspain.com/browse/SUIC-128)

Add new types and colors.

Captura de pantalla 2019-06-28 a las 8 35 04


components/atom/button/src/index.scss
```diff
@@ -29,40 +41,11 @@ $icon-sz-atom-button-large: $sz-icon-m !default;
...
-$bgc-atom-button-negative-secondary-focus: $bgc-atom-button-dark !default;
-$bxsh-atom-button-negative-primary: none !default;
-
-// Accent
-$c-atom-button-accent: $c-white !default;
-$bgc-atom-button-accent-focus: $c-accent-dark !default;
-$bdc-atom-button-accent-focus: $c-accent-dark !default;
-$bgc-atom-button-accent: $c-accent !default;
-$bdc-atom-button-accent: $c-accent !default;
+$bdrs-atom-buton-first-outline: $bdrs-m 0 0 $bdrs-m !default;
```
rmoralp: typo in word button

rmoralp (dismissed):

JoanClaret (dismissed):

adriadoz (dismissed):

emiliovillu (dismissed):

rmoralp (dismissed):

components/atom/button/src/index.js
```diff
@@ -5,7 +5,22 @@ import cx from 'classnames'
 import Button from './Button'
 
 const CLASS = 'sui-AtomButton'
-const TYPES = ['primary', 'accent', 'secondary', 'tertiary']
+const TYPES = ['solid', 'outline', 'flat']
```
jcger, line 8: Where is the definition of all of this? I checked https://github.com/SUI-Components/UX-Definitions/blob/master/Atom/Buttons.md and types and colors are named different 🤷‍♂

jcger (changes requested):

jcger (dismissed):

ruben-martin-lozano (dismissed):

components/atom/button/src/index.js
jcger: I just figured out that now our code > their definition

emiliovillu (approved):

adriadoz (approved):

JoanClaret (approved):

kikoruiz (commented):
Sorry, I couldn't give my comment before, but I don't agree with this PR for 2 main reasons:

1. Compatibility not kept with the current version, so that having a new version of this component, the most used everywhere, will force us to update all our deps in every such components in order to not to have different versions of the button in our final bundle.
2. Last but not least, the CSS generated from the new Sass code is more than 7x the current bundle size, and that's something we should take care as much as we can (I'm attaching an screenshot of both versions compiled in my computer).

Screenshot 2019-07-01 at 21 36 59

Since solving the second issue will make this feature to take a long time to be ready in PROD, I would consider to go on with the solution suggested in https://github.com/SUI-Components/sui-components/pull/730.

turolopezsanabria:
Hello, 
¿este tema cómo sigue? (he puesto este mismo comment en https://github.com/SUI-Components/sui-components/pull/730)

midudev:
We're going to merge this one: https://github.com/SUI-Components/sui-components/pull/730

PR #737, opened 2019-07-10
feat(molecule/thumbnail): updated README accorging to new props available

juanmaguitar:
Update README according to new props available in the component
Also force new MAJOR version (previous task involving this component, already merged, should have added this mark but it didn't) as there are new props available and old ones deprecated as an updated AtomImage is used

BREAKING CHANGE:
new props caused by internal use of AtomImage

### DIFF
--- a/components/molecule/thumbnail/README.md
+++ b/components/molecule/thumbnail/README.md
@@ -24,6 +24,7 @@ return (<MoleculeThumbnail
 ```
 
 ### Properties
+
 #### Shape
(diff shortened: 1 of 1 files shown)


jcger:
does this mean that the last minor version contains breaking changes?

juanmaguitar:
> does this mean that the last minor version contains breaking changes?

@jcger Yes. but it hasn't be released yet. It will be released after this PR is merged

rmoralp (approved):

jcger (approved):

adriadoz (approved):

PR #740, opened 2019-07-22
refactor(atom/backToTop): migration to hooks

juanmaguitar:
- 🌍Demo Online → https://sui-components-atom-back-to-top-hooks-migration.now.sh/workbench/atom/backToTop/demo

### DIFF
--- a/components/atom/backToTop/src/index.js
+++ b/components/atom/backToTop/src/index.js
@@ -27,35 +27,32 @@ const AtomBackToTop = ({
   const [show, setShow] = useState(false)
   const [hover, setHover] = useState(false)
 
-  let intervalRef = useRef()
+  const intervalRef = useRef()
 
-  useEffect(
-    () => {
-      const container = getTarget(refContainer)
-      const {scrollHeight, clientHeight} = container
-      const halfHeight = Math.floor((scrollHeight - clientHeight) / 2)
+  useEffect(() => {
+    const container = getTarget(refContainer)
+    const {scrollHeight, clientHeight} = container
+    const halfHeight = Math.floor((scrollHeight - clientHeight) / 2)
 
-      const handleScroll = () => {
-        const {scrollTop} = container
-        if (scrollTop > halfHeight) {
-          if (!show) {
-            setShow(true)
-          }
-        } else {
-          if (show) {
-            setShow(false)
-            setHover(false)
-          }
+    const handleScroll = () => {
+      const {scrollTop} = container
+      if (scrollTop > halfHeight) {
+        if (!show) {
+          setShow(true)
+        }
+      } else {
+        if (show) {
+          setShow(false)
+          setHover(false)
         }
       }
-
-      container.addEventListener('scroll', handleScroll)
-      return () => {
-        container.removeEventListener('scroll', handleScroll)
-      }
-    },
-    [refContainer, show]
-  )
+    }
+    container.removeEventListener('scroll', handleScroll)
+    container.addEventListener('scroll', handleScroll)
+    return () => {
+      container.removeEventListener('scroll', handleScroll)
+    }
+  }, [refContainer, show])
 
   const scrollStep = () => {
     const container = getTarget(refContainer)
--- a/components/atom/backToTop/src/index.js
+++ b/components/atom/backToTop/src/index.js
@@ -51,7 +51,6 @@ const AtomBackToTop = ({
       }
       container.addEventListener('scroll', handleScroll)
       return () => {
-        clearInterval(intervalId)
         container.removeEventListener('scroll', handleScroll)
       }
     },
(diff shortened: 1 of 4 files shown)


components/atom/backToTop/src/index.js
```diff
@@ -16,95 +16,83 @@ const STYLES = {
...
+      container.addEventListener('scroll', handleScroll)
+      return () => {
+        container.removeEventListener('scroll', handleScroll)
+      }
+    },
+    [clientHeight, intervalId, scrollHeight, show]
+  )
+
+  const scrollStep = () => {
+    const {scrollTop} = container
```
midudev, line 58: I would say you have to extract the value from `container.current`.

components/atom/backToTop/src/index.js
```diff
@@ -16,95 +16,83 @@ const STYLES = {
...
-    const {intervalId} = this
+const AtomBackToTop = ({
+  iconTop: IconTop,
+  textTop,
+  scrollSteps,
+  scrollIntervalTime,
+  refContainer,
+  style
+}) => {
+  const [show, setShow] = useState(null)
```
midudev: I see you're using show as boolean. We might want to put as initial value `false` as well.

components/atom/backToTop/src/index.js
```diff
@@ -16,95 +16,83 @@ const STYLES = {
...
+  const [hover, setHover] = useState(false)
+
+  let intervalId = 0
+  let container = useRef(null)
+
+  container = getTarget(refContainer)
```
midudev: This could break SSR, getting the target and the properties below should be done in an effect.

components/atom/backToTop/src/index.js
juanmaguitar: The `getTarget` method used to get the DOM container is already returning the `target.current`

components/atom/backToTop/src/index.js
juanmaguitar: 👍

components/atom/backToTop/src/index.js
juanmaguitar: Moved getting the container element inside the effect 👍

components/atom/backToTop/src/index.js
```diff
@@ -1,4 +1,6 @@
-import React, {PureComponent} from 'react'
+/* eslint-disable */
```
midudev: 👀

components/atom/backToTop/src/index.js
```diff
@@ -16,95 +18,80 @@ const STYLES = {
...
+  textTop,
+  scrollSteps,
+  scrollIntervalTime,
+  refContainer,
+  style
+}) => {
+  const [show, setShow] = useState(false)
+  const [hover, setHover] = useState(false)
+
+  let intervalId = 0
```
midudev: I think this should be a `ref` otherwhise, the value will be last in each render.

components/atom/backToTop/src/index.js
```diff
@@ -16,95 +18,80 @@ const STYLES = {
...
-          <span className={CLASS_ICON}>
-            <IconTop />
-          </span>
-        )}
-        {textTop && <span className={CLASS_TEXT}>{textTop}</span>}
-      </button>
-    )
-  }
+  const hoverButton = () => setHover(true)
+  const unhoverButton = () => setHover(true)
```
midudev: `const unhoverButton = () => setHover(false)`

components/atom/backToTop/src/index.js
juanmaguitar: 🤭removed

components/atom/backToTop/src/index.js
juanmaguitar: Well, the `ref` is actually being passed via `refContainer` prop.

components/atom/backToTop/src/index.js
juanmaguitar: fixed!

components/atom/backToTop/src/index.js
midudev: I mean the `intervalId` value that it might be a ref.

```
const intervalId = useRef()
```

Otherwise, the value could be recreated for every render and you will lose the value.

juanmaguitar: oks, fixed!

midudev (dismissed):

midudev:
There's a conflict to deal with. I suppose after the sui-lint update.

AgonisticKatai (dismissed):

juanmaguitar:
Because of conflicts w/ other tasks merged in master and to avoid problems, another PR from an updated version of `master` has been created (https://github.com/SUI-Components/sui-components/pull/762)

PR #752, opened 2019-08-01
feat(molecule/select): implement readOnly property on MultipleSelection and SingleSelection

jordevo:
On mobile web, when tapping on a `MoleculeSelectFieldMultiSelection` or `MoleculeSelectFieldSingleSelection`, the input keyboard pops up.
The `readOnly` property prevents that from happening on the input component at an atomic level but it seemed like the molecules were not passing down that property. Now it is.
I've added markup to the `molecule/select` demo for that case. A demo can be found here: https://public.jordimunoz.now.sh

### DIFF
--- a/components/molecule/select/src/components/MultipleSelection.js
+++ b/components/molecule/select/src/components/MultipleSelection.js
@@ -22,6 +22,7 @@ const MoleculeSelectFieldMultiSelection = props => {
     keysSelection,
     id,
     size,
+    readOnly,
     optionsData = {}
   } = props
 
@@ -56,6 +57,7 @@ const MoleculeSelectFieldMultiSelection = props => {
         optionsData={optionsData}
         autocomplete="off"
         noBorder
+        readOnly={readOnly}
       />
       <MoleculeDropdownList
         checkbox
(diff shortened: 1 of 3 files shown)


danivm (approved):

AgonisticKatai (approved):

juanmaguitar:
I don't understand this task. If you check the current demos at https://sui-components.now.sh/workbench/molecule/select/demo you can verify that the final `input` rendered by react already has the proper `readonly` attribute.

Captura de pantalla 2019-08-02 a las 10 16 45

If the final purpose of this PR is to have the `readonly` attribute in the final `input` this is being done actually. 

Besides this **the internal `input` used by this component should always have this `readonly` attribute** (to avoid browser autocomplete for example), so this should be handled internally (the final developer shouldn't be worried about passing this extra `prop`)

juanmaguitar (changes requested):
https://github.com/SUI-Components/sui-components/pull/752#issuecomment-517608632

jordevo:
@juanmaguitar I'm afraid I might have been tinkering with an older version or something... I just knew about this https://github.com/SUI-Components/sui-components/pull/747 so I reckon this PR can be closed. Thanks! 👍

PR #759, opened 2019-08-08
refactor(atom/upload): hooks migration

juanmaguitar:
- 🌎 Online Demo → https://sui-components-atom-upload-hooks-migration.now.sh/workbench/atom/upload/demo

### DIFF
--- a/components/atom/upload/src/index.js
+++ b/components/atom/upload/src/index.js
@@ -1,6 +1,8 @@
-import React, {PureComponent} from 'react'
+import React, {lazy, Suspense} from 'react'
 import PropTypes from 'prop-types'
 import cx from 'classnames'
+
+const Dropzone = lazy(() => import('react-dropzone'))
 
 const STATUSES = {
   ACTIVE: 'active',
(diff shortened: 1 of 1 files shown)


components/atom/upload/src/index.js
```diff
@@ -16,70 +18,42 @@ const CLASS_BLOCK_TEXT_SECONDARY = `${CLASS_BLOCK_TEXT}-secondary`
...
-            onDrop={onDrop}
-          >
-            {this.renderStatusBlock(status)}
-          </Dropzone>
-        )
-      )
-    }
+  if (Object.values(STATUSES).includes(status)) {
+    return (
+      <Suspense fallback={null}>
```
midudev: 👀 I think this could break Server Side Rendering as Suspense is not supported on the server.

midudev: To fix this, it should be used only behind a state change with a useEffect.

juanmaguitar: Fixed! `Suspense` will never be rendered if `ready` is not true which is set only in `useEffect` that will not be executed in SSR

desko27 (approved):

midudev (approved):

PR #781, opened 2019-09-16
Add all UX Definitions into SUI Components

turolopezsanabria:
Adding all the definitions as Markdown, in the same folder as the components.
Ticket: https://jira.scmspain.com/browse/SUIC-129
Name of the UX specs: UXDEF.MD

### DIFF
--- a/components/atom/backToTop/UXDEF.md
+++ b/components/atom/backToTop/UXDEF.md
@@ -1,0 +1,126 @@
+# SUI - Back to Top
+A simple way to return back to the top of a long content page.
+
+|   Status          | WIP |
+|   :----           |   :---- |
+|   Current version |   1.0|
+|   Category        |   Atom |
+|   Owners          |   UX @Irene L, @Chris J |
+
+## Structure
+
+Back to Top consists in a floating element located on the bottom right corner of the page and with a [z-index](https://paper.dropbox.com/doc/z-index-WIP--AMqq1DAVVk5QOOeNOojmYuRrAg-4QvLsXUoWXeWNDiQSzCBc) position.
+
+![](https://d2mxuefqeaa7sj.cloudfront.net/s_2B6461E0A5AACA052E49BE35E96B0E3C48FC369AADACF7054C77A66FA69B339E_1536830620353_1-+Back+to+Top+-+Structure.png)
+
+## Behavior
+
+Back to Top shows up when the user scrolls down beyond 50% of the visible part of the page and disappears when the user reaches that same limit.
+When the user clicks or taps on the Back to Top, the page scrolls to the top of the page and the Back to Top disappears.
+
+
+## Types
+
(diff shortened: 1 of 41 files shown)


components/atom/backToTop/UXDEF.md
```diff
@@ -0,0 +1,126 @@
+# SUI - Back to Top
+A simple way to return back to the top of a long content page.
+
+|   Status          | WIP |
+|   :----           |   :---- |
+|   Current version |   1.0|
+|   Category        |   Atom |
+|   Owners          |   UX @Irene L, @Chris J |
```
kikoruiz, line 8: Could be great to use real links to people if they have access to GitHub 😉

components/atom/backToTop/UXDEF.md
```diff
@@ -0,0 +1,126 @@
...
+Visuals should be custom on each vertical. This specs are for a SUi white label design:
+
+FONT
+font-size: #fs-m (font-size:16px & line-height:20px) 
+
+ICON
+size: #m 
+
+COLORS DARK CONTRAST
+background $bgc-gray-d2
```
kikoruiz, line 59: I don't know if it would be better to write some kind of metalanguage instead of using the real Sass variables names. What do you think about?

kikoruiz:
I think `SPECS.md` is a great name for the new documents 😃

components/atom/backToTop/UXDEF.md
turolopezsanabria: a "+1" total to this, to implement once all Designers have gitHub (and I know their usernames claro)

components/atom/backToTop/UXDEF.md
turolopezsanabria: yeap, not only colours but the same with all vars

turolopezsanabria:
about the name: yes, but the idea is that the notes in the new tab are not "specs" exactly, but notes and ux comments that complete and add more context to the actual React components.
The specs (instructions to develop) should live somewhere else, and be valid only to code the component and QA, but later dismiss. What do you think?

turolopezsanabria:
the idea is to go as Lean as possible. Put everything together and later, once everything is in the same place, to do a HUGE work rewriting, adjusting, cleaning each of these documents.

components/atom/backToTop/UXDEF.md
midudev: +1

midudev (approved):

midudev:
I already used UXDEF.md as the name of the file for look for. @turolopezsanabria if you want to change this, please, contact me before merging. Thanks!

turolopezsanabria:
ok perfect, I will not change the name for now

jcger (approved):

kikoruiz:
Totally agree after your explanation about the name @turolopezsanabria 👍 Good work!

PR #787, opened 2019-09-18
fix(atom/input): fix input:disabled look on iOS/safari

jordevo:
**Background**
iOS Safari is rendering the input element of a disabled input atom with a 0.4 opacity and a subtle shade under the top border which makes it inconsistent with the look & feel of other SUI components that make use of the input atom such as Molecule Select. 

**Goal**
To fix this inconsistency.
Before this change: 
image
After this change:
image

**Tradeoffs**
This adds a bit of browser-specific attributes to the codebase.

### DIFF
--- a/components/atom/input/src/Input/Component/index.scss
+++ b/components/atom/input/src/Input/Component/index.scss
@@ -38,6 +38,8 @@ $class-read-only: '#{$base-class}--readOnly';
     &:not(#{$class-read-only}) {
       background: $bgc-atom-input-disabled;
     }
+    opacity: 1;
+    -webkit-appearance: none;
   }
 
   &:focus {


rmoralp (approved):

components/atom/input/src/Input/Component/index.scss
```diff
@@ -38,6 +38,8 @@ $class-read-only: '#{$base-class}--readOnly';
     &:not(#{$class-read-only}) {
       background: $bgc-atom-input-disabled;
     }
+    opacity: 1;
+    -webkit-appearance: none;
```
JoanClaret, line 42: did you try to just add `appearance: none;` and let autoprefixer/webpack do the magic?

rmoralp: Agree

jordevo: thought of doing that but tbh I wasn't sure how we're dealing with autoprefixer and since the issue is only occurring on Safari I just left it with the -webkit- prefix.

JoanClaret: if you just add the regular declaration, then every single vertical can choose which browsers want to prefix by configuring its bundler. I believe this is more aligned with the SUI philosophy.

jordevo: gotcha :)

devictoribero:
🚨 Care, the build is failing and it's not because of a code related to this `PR`. Is this usual? Do we have issues deploying to `Zeit`?

`Error: Error: Cannot find module '/home/travis/.now.json'`

**EDIT: I see it happens with another `PR`'s too, so I guess it's well known**

devictoribero (approved):
Would be good as it's been pointed out to try first `appearance: none;`.

I don't know if with this property the `postcss` will add the `auto prefix`

PR #789, opened 2019-09-18
fix(molecule/select): set opacity: 1 and appearance: none to disabled input element

jordevo:
… when used in molecule select.

**Background**
iOS Safari is rendering the input element of a disabled Atom Input within a Molecule Select with a 0.4 opacity and a subtle shade under the top border which makes it inconsistent with the look & feel of other SUI components. 

**Goal**
To fix this inconsistency.

Before this change: 
image

After this change:
image

**Tradeoffs**
This adds a bit of browser-specific attributes to the codebase.

### DIFF
--- a/components/molecule/select/src/index.scss
+++ b/components/molecule/select/src/index.scss
@@ -40,6 +40,11 @@ $bgc-atom-input-disabled: $c-gray-lightest !default;
         }
       }
 
+      #{$class-select-atom-input-input}:disabled {
+        opacity: 1;
+        -webkit-appearance: none;
+      }
+
       #{$class-select-atom-input-input}:first-child,
       #{$class-select-atom-input-tags} {
         padding-right: $pr-select-atom-input-tags;


components/molecule/select/src/index.scss
```diff
@@ -40,6 +40,11 @@ $bgc-atom-input-disabled: $c-gray-lightest !default;
         }
       }
 
+      #{$class-select-atom-input-input}:disabled {
+        opacity: 1;
```
kikoruiz, line 44: This won't work if you move this properties into the line 39? It seems it's applying to the same selector...

jordevo: Won't work as we want. I think whatever there is on L39 applies to #{$class-select-atom-input-input} within an explicitly disabled molecule select. In our case we want to make sure our opacity and appearance is applying to the disabled input element of the atom input, whether or not the molecule select containing it is disabled. Does it make sense?

kikoruiz: ok, it makes sense

kikoruiz (approved):

midudev (approved):

PR #794, opened 2019-09-25
fix/range default value

a-ferrante:


components/atom/slider/src/index.js
```diff
@@ -127,6 +128,9 @@ AtomSlider.propTypes = {
   /** value  */
   value: PropTypes.number,
 
+  /** defaultValue prop that set initial positions of handles */
+  defaultValue: PropTypes.array,
```
jcger: should be a number in case of slider and array in case of range

jcger (approved):

kikoruiz (approved):

jordevo (approved):

PR #797, opened 2019-09-30
Feat/remove fixed type

jcger:
main reason of this change is that we need to be able to set different `designs` to different buttons inside this component but it's setting the same for all. 

I'm going to remove it completely and create a major because of:
- The button component prioritizes the `type` property higher than `design` or `color` which could be handled by adding some logic here but would make this component even more sui-atom-button dependent
- The`type` prop is already deprecated within the atom button component
- Removing the `type` prop makes this component reusable wrapping other type of buttons or components that aren't the sui-atom-button component, the dependency would be completely removed

### DIFF
--- a/components/molecule/buttonGroup/package.json
+++ b/components/molecule/buttonGroup/package.json
@@ -9,8 +9,7 @@
     "build:styles": "../../../node_modules/.bin/cpx './src/**/*.scss' ./lib"
   },
   "dependencies": {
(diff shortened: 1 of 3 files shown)


a-ferrante (approved):

midudev (approved):

turolopezsanabria:
👍

turolopezsanabria (approved):

PR #799, opened 2019-10-02
Atom radio button component

juanmaguitar:
- ⚙︎Task → https://jira.scmspain.com/browse/SUIC-133
- 🌎Demo Online → https://sui-components-atom-radio-button-component.now.sh/workbench/atom/radioButton/demo


components/atom/radioButton/src/index.scss
```diff
@@ -0,0 +1,64 @@
+@import '~@schibstedspain/sui-theme/lib/index';
+
+$w-atom-radiobutton: 15px !default;
```
jcger: all hardcoded values should be tokens

components/atom/radioButton/src/index.js
```diff
@@ -0,0 +1,84 @@
+import React from 'react'
+import PropTypes from 'prop-types'
+import cx from 'classnames'
+
+const BASE_CLASS = 'sui-AtomRadioButton'
+const CLASS_HIDDEN = `is-hidden`
+
+const ERROR_STATES = {
```
giodelabarrera: I get the feeling of how you use within the components this object is more like response states. If so, I think that for greater readability it should be `RESPONSE_STATUS`

components/atom/radioButton/src/index.scss
juanmaguitar: All values replaced by the proper token but this one:
$bw-atom-radiobutton-checked: 6px !default;

As there is no token w/ this value and I need exactly this value to emulate the effect reflected in the specification

components/atom/radioButton/src/index.js
juanmaguitar: Removed ERROR_STATES logic as this component doesn't really need to be aware of this (the specification doesn't reflect any behaviour in this level regarding errors)

kikoruiz (approved):

jcger (approved):

PR #806, opened 2019-10-08
feat(atom/backToTop): add min height prop

quinwacca:
With that optional prop, we now can define a minimum height in pixels to show back to top button.
```
/** Minimum height (in pixels) to show button */
minHeight: PropTypes.number,
```
_The problem I want to solve_: 
To show the button when the user at the half of the scrollHeight is good but in the case, we got a scroll of maybe 6000px, the user doesn't have a quick way to back to top until he arrives at 3000px.
With that prop, we can have the option to show that button before, higher on the page, at maybe 1000px.

Please, share your thoughts about it!

### DIFF
--- a/components/atom/backToTop/src/index.js
+++ b/components/atom/backToTop/src/index.js
@@ -18,11 +18,12 @@ const STYLES = {
 
 const AtomBackToTop = ({
   iconTop: IconTop,
+  minHeight,
   textTop,
-  scrollSteps,
-  scrollIntervalTime,
-  refContainer,
-  style
+  scrollSteps = 100,
+  scrollIntervalTime = 50,
+  refContainer = document.body,
+  style = STYLES.DARK
 }) => {
   const [show, setShow] = useState(false)
   const [hover, setHover] = useState(false)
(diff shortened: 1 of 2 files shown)


rmoralp (approved):

AitorRodriguez990 (approved):

jordevo (approved):

components/atom/backToTop/src/index.js
```diff
@@ -18,11 +18,12 @@ const STYLES = {
...
   iconTop: IconTop,
+  minHeight,
   textTop,
-  scrollSteps,
-  scrollIntervalTime,
-  refContainer,
-  style
+  scrollSteps = 100,
+  scrollIntervalTime = 50,
+  refContainer = document.body,
```
midudev, line 25: I realized now that this could potentially be broken on server-side rendering.

components/atom/backToTop/src/index.js
```diff
@@ -35,7 +36,7 @@ const AtomBackToTop = ({
       const {scrollTop, scrollHeight, clientHeight} = container
       const halfHeight = Math.floor((scrollHeight - clientHeight) / 2)
 
-      if (scrollTop > halfHeight) {
+      if (scrollTop > halfHeight || scrollTop > minHeight) {
```
midudev: What happens if `minHeight` is `undefined`? Or the new prop is mandatory (that would be a major) or we do something in order to avoid breaking the component if no `minHeight` is passed by props.

midudev (approved):

components/atom/backToTop/src/index.js
quinwacca: So is it better to define default values on `.defaultProps`?

quinwacca: Or the problem is `document.body`

PR #814, opened 2019-10-10
Feat/molecule notification rounded corners

davecarter:
## PR Summary

This PR allows any vertical to implement a better-customized icon and text color in `Sui-Molecule-Notification`

In addition, now you can decide if you want the notification with rounded corners via a prop with pre-defined sizes:

```javascript
BRDS_SIZE = {
  extraLarge: 'xl',
  large: 'l',
  medium: 'm',
  small: 's',
  extraSmall: 'xs'
}
```

### Screenshot

Screenshot 2019-10-10 at 12 53 32

### DIFF
--- a/components/molecule/notification/src/_settings.scss
+++ b/components/molecule/notification/src/_settings.scss
@@ -2,6 +2,7 @@ $w-notification: 100% !default;
 $z-notification: 1000 !default;
 $mh-notification: 500px !default;
 $sz-notification-icon: $sz-icon-m !default;
+$bdrs-notification: 8px !default;
 
 // Colors
 
@@ -19,6 +20,26 @@ $notification-type-font-colors: (
   warning: $c-white,
   error: $c-white,
   system: $c-white
+) !default;
+
+// Positive Text color
+
+$notification-type-positive-text-colors: (
+  info: $c-primary,
+  success: $c-success,
+  warning: $c-alert,
+  error: $c-error,
+  system: $c-system
+) !default;
+
+// Positive Icon color
+
+$notification-type-positive-icon-colors: (
+  info: $c-primary,
+  success: $c-success,
+  warning: $c-alert,
+  error: $c-error,
+  system: $c-system
 ) !default;
 
 // Effects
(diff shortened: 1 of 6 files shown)


components/molecule/notification/src/_settings.scss
```diff
@@ -2,6 +2,7 @@ $w-notification: 100% !default;
 $z-notification: 1000 !default;
 $mh-notification: 500px !default;
 $sz-notification-icon: $sz-icon-m !default;
+$bdrs-notification: 8px !default;
```
arnau-rius: Taking advantage of this PR, we should review `border-radius` tokens with @jcger in the next SUI meeting, because in INMHO we should extend the tokens for border-radius property.

arnau-rius (approved):

components/molecule/notification/src/_settings.scss
JoanClaret: Agree. in the past we've got `$bdrs-base`
https://github.com/SUI-Components/sui-theme/blob/master/src/settings-compat-v7/_box-style.scss#L4
but now is deprecated.

components/molecule/notification/src/index.js
```diff
@@ -203,6 +208,7 @@ MoleculeNotification.defaultProps = {
   effect: true,
   onClose: () => {},
   position: 'relative',
+  roundedCorners: false,
```
JoanClaret: maybe we could move all those default props to the new React definition.

JoanClaret (approved):

JoanClaret:
should we adjust the SPECS? looks like so.
https://github.com/SUI-Components/sui-components/blob/master/components/molecule/notification/UXDEF.md#visual

components/molecule/notification/src/_settings.scss
turolopezsanabria: I do agree with the tokens for Border Radius.
I suggest we continue with this PR as it is, and we move the border-radius in parallel

davecarter:
Definitely!
Thanks for pointing that out Mr @JoanClaret 😃

tomasmax (approved):

turolopezsanabria:
👍

carlosvillu (approved):

JoanClaret (approved):

PR #828, opened 2019-10-21
(molecule/autosuggestField) - feature/contrast label

a-ferrante:
Allow autosuggest field to use a contrast label setting a prop, this can be used when the autosuggestField is used on dark backgrounds. 

Also changes on the AtomLabel component to have a "CONTRAST" type that can be overridden by each theme

Online demo: https://sui-components-pr-828.now.sh/workbench/atom/label/demo


juanmaguitar:
@a-ferrante Can you please add a link to the demo (of this PR) including a use case of this feature?

a-ferrante:
@juanmaguitar done! have in mind that the demo for the autosuggestField would not reflect the changes since the AtomLabel hasn't been released yet since both are in the same PR

jcger (approved):

ferransimon (approved):

PR #830, opened 2019-10-22
Feat/slider inverse colors

a-ferrante:
Use a prop to invert the colors of the slider track and the slider rail

### DIFF
--- a/components/atom/slider/src/index.js
+++ b/components/atom/slider/src/index.js
@@ -53,7 +53,7 @@ const AtomSlider = ({
   valueLabelFormatter,
   hideTooltip,
   defaultValue,
-  inverseColors
+  invertColors
 }) => {
   const [ready, setReady] = useState(false)
   const [handleComponent, setHandle] = useState({component: null})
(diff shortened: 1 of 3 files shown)


jcger (approved):

components/atom/slider/src/index.js
```diff
@@ -151,7 +157,9 @@ AtomSlider.propTypes = {
   /* callback to format the value shown as label */
   valueLabelFormatter: PropTypes.func,
   /* flag to hide tooltip if wanted */
-  hideTooltip: PropTypes.bool
+  hideTooltip: PropTypes.bool,
+  /* If true it will invert the colors for selected track and rail */
+  invertColors: PropTypes.bool
```
ferransimon, line 162: I know that what it technically does is invert colors but I think the prop of the component should describe not its technically use but its meaning, and it is to invert the evaluation from min to max or the other way around, maybe `invertRange` ?

ferransimon (approved):

components/atom/slider/src/index.js
a-ferrante: i disagree since the slider its not technically a range, we are not inverting any evaluation, we are just using a visual change to give a different feeling/UX. I think its clearer for a prop name to describe what it does to the component without any assumption about the goal that you want to accomplish. 
Using another name like invertRange or something like this can cause confusion, since we are not doing any of that.

PR #837, opened 2019-10-23
fix(atom/tag): include classes from className prop in AtomTag classNames

jordevo:
Currently, `AtomTag` is not doing anything with whatever is sent as `className` prop.
The purpose of this PR is to fix this apparent bug.

### DIFF
--- a/components/atom/tag/src/index.js
+++ b/components/atom/tag/src/index.js
@@ -34,14 +34,15 @@ const filterKeys = (obj, listOfProps) =>
   }, {})
 
 const AtomTag = props => {
-  const {href, icon, onClick, size, responsive} = props
+  const {href, icon, onClick, size, responsive, className} = props
 
   const isActionable = onClick || href
   const classNames = cx(
     'sui-AtomTag',
     `sui-AtomTag-${size}`,
     responsive && 'sui-AtomTag--responsive',
-    icon && 'sui-AtomTag-hasIcon'
+    icon && 'sui-AtomTag-hasIcon',
+    className
   )
 
   /**
--- a/demo/atom/tag/playground
+++ b/demo/atom/tag/playground
@@ -23,6 +23,12 @@ return (
         onClose={() => alert('close!')}
         closeIcon={<CloseIcon />}
       />
+      <AtomTag
+        label="Tag with custom className"
+        className="custom-className"
+        onClose={() => alert('close!')}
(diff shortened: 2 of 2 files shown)


JoanClaret (approved):

juanmaguitar:
`sui-components` are not supossed to have classNames in order to control its customization only by pre-defined SASS variables in the component
Can you please confirm this @jcger @carlosvillu @midudev @turolopezsanabria ?

components/atom/tag/src/index.js
```diff
@@ -34,14 +34,15 @@ const filterKeys = (obj, listOfProps) =>
   }, {})
 
 const AtomTag = props => {
-  const {href, icon, onClick, size, responsive} = props
+  const {href, icon, onClick, size, responsive, className} = props
```
kikoruiz, line 37: Passing a `className` as prop is so dangerous because you're defining your component as "free" of styles, and our components has been defined regarding some "closed" styles. But it's true that this specific component should allow us to define some custom color variations, is that what you need @jordevo ?

kikoruiz: If so, you could do something like I've committed in other branch: https://github.com/SUI-Components/sui-components/commit/ed721ae0f7d4cd03f9cdb18b1f0d9be087134d49. This way any vertical could define in its own theme the types (variations) they need:

__VERTICAL THEME file__
```css
$atom-tag-types: (
  'financed': (
    bgc: red,
    c: white
  ),
  'top': (
    bgc: blue,
    c: white
  )
);
```

__VERTICAL COMPONENT file__
```js
<AtomTag type='financed' />
```

How do you see this solution? @turolopezsanabria @juanmaguitar @carlosvillu @midudev @jcger

turolopezsanabria:
I agree with the above, I would rather set all the possible colours as props

components/atom/tag/src/index.js
turolopezsanabria: plus 1 to this

juanmaguitar: A very elegant solution. Yes, sir 👍

jordevo: I get your point, @kikoruiz ! thanks for your solution! 👏

kikoruiz: Alright, I'm going on with this solution, thanks!

PR #839, opened 2019-10-23
[feat] allow to set custom atom tag types

kikoruiz:
### Description

`AtomTag` is a very atomic and fool component that in many cases needs some extra customisation by the app where is used like color and background color information. Right now it's not possible to do so despiste of overwriting the `sui` component class from the high order component that uses it, for that reason I'm suggesting to set a new prop `type` to allow this simple and controlled customisation.

### Usage

__YOUR APP THEME__ (Sass)
```css
$atom-tag-types: (
  'financed': (
    bgc: red,
    c: white
  ),
  'top': (
    bgc: blue,
    c: white
  )
);
```

__YOUR HIGH ORDER COMPONENT__
```js
<AtomTag type='financed' />
```

### DIFF
--- a/components/atom/tag/README.md
+++ b/components/atom/tag/README.md
@@ -25,4 +25,29 @@ import AtomTag from '@schibstedspain/sui-atom-tag'
 />
 ```
 
+### Tag types
+
+If you want to customize your tag you can pass a prop to identify this type and you also need to set your custom set of types in Sass:
+
+**Your theme file** (Sass)
+
+```css
+$atom-tag-types: (
+  "alert": (
+    bgc: red,
+    c: white
+  ),
+  "warning": (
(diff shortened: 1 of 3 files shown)


jordevo (approved):
great, elegant solution! well done! 👏

rmoralp (approved):

devictoribero (approved):
Super elegant LOL! :clap: 

The bad thing is that we have no control over the colors at all, BUT I think it's totally worth it for this component.


Another solution that comes to my mind is to create a `prop` as you said but with `oneOf([])`

```
type: PropTypes.oneOf([primary, accent, default, info, warning, success, ...])
```
With this solution that is more restrictive, UX should be only using those.

PR #859, opened 2019-11-06
feat(molecule/pagination): bump moleculeButtonGroup 

a-ferrante:
Following the release of the new MoleculeButtonGroup version we can now update the version on this component.
This will allow to set different designs on the navigation and page buttons 🎉 

https://sui-components-pr-859.now.sh/workbench/molecule/pagination/demo

### DIFF
--- a/components/molecule/pagination/src/index.js
+++ b/components/molecule/pagination/src/index.js
@@ -15,12 +15,20 @@ const BASE_CLASS = 'sui-MoleculePagination'
 const CLASS_PREV_BUTTON_ICON = 'sui-MoleculePagination-prevButtonIcon'
 const CLASS_NEXT_BUTTON_ICON = 'sui-MoleculePagination-nextButtonIcon'
 
-// eslint-disable-next-line react/prop-types
-const PageButton = ({onSelectPage, page, ...props}) => {
+const PageButton = ({onSelectPage, page, design, ...props}) => {
   const _onSelectPage = e => {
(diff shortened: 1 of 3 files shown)


ferransimon (approved):

components/molecule/pagination/src/index.js
```diff
@@ -35,7 +43,11 @@ const MoleculePagination = ({
   nextButtonIcon: NextButtonIcon,
   onSelectPage,
   compressed,
-  hideDisabled
+  hideDisabled,
+  selectedPageButtonDesign,
+  nonSelectedPageButtonDesign,
```
jordevo, line 48: doesn't matter much but... why not just `unSelectedPageButtonDesign` or even `pageButtonDesign`?

midudev (approved):

components/molecule/pagination/src/index.js
a-ferrante: i think pageButtonDesing would infer that it will be the design for all the page buttons and that is  innacurate, thats the point of having two props, one for selected and other one for the non selected pages.

nonSelected and unSelected are synonyms so i think both prop names are clear enough right?

jordevo: fair enough...

IMHO `unSelected` sounds more precise to me

jcger (approved):

adriadoz (approved):

PR #877, opened 2019-11-21
fix(molecule/select): remove useEffect when populating data

jordevo:
Not sure if we're actually fixing a bug here.

`MoleculeSelect` is using useRef and useEffect to inform its `MoleculeSingleSelection` and `MoleculeMultipleSelection`'s  `optionsData` prop.

When we use a `useEffect` hook on a parent component to fetch data and populate the `MoleculeSelect`'s children *AND* we have a preselected value (i.e. a `value` prop passed down from the parent) the user is not seeing it as selected unless the user interacts with the component.

Before these changes:
moleculeSelect_useeffect2

With these changes:
moleculeSelect_useeffect

### DIFF
--- a/components/molecule/select/src/index.js
+++ b/components/molecule/select/src/index.js
@@ -1,4 +1,4 @@
-import React, {useState, useEffect, useRef} from 'react'
+import React, {useState, useRef} from 'react'
 import PropTypes from 'prop-types'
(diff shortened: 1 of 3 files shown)


jordevo:
Closed as a better solution has been suggested by @kikoruiz and @midudev : to use useState .

PR #878, opened 2019-11-21
[fix] set default options data to molecule select

kikoruiz:
`MoleculeSelect` is using useRef and useEffect to inform its `MoleculeSingleSelection` and `MoleculeMultipleSelection`'s  `optionsData` prop.

When we use a `useEffect` hook on a parent component to fetch data and populate the `MoleculeSelect`'s children *AND* we have a preselected value (i.e. a `value` prop passed down from the parent) the user is not seeing it as selected unless the user interacts with the component.

Before these changes:
moleculeSelect_useeffect2

With these changes:
moleculeSelect_useeffect

### DIFF
--- a/components/molecule/select/src/index.js
+++ b/components/molecule/select/src/index.js
@@ -1,4 +1,4 @@
-import React, {useState, useEffect, useRef} from 'react'
+import React, {useState, useRef} from 'react'
 import PropTypes from 'prop-types'
 import cx from 'classnames'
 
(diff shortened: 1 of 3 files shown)


jcger (approved):

midudev (approved):

components/molecule/select/src/index.js
```diff
@@ -1,4 +1,4 @@
-import React, {useState, useEffect, useRef} from 'react'
+import React, {useState, useRef, useEffect} from 'react'
```
AgonisticKatai, line 1: 😬

AgonisticKatai (approved):

jordevo (approved):

components/molecule/select/src/index.js
kikoruiz: 🖕

PR #879, opened 2019-11-21
[fix] avoid onClose execution when show prop is false

giodelabarrera:
Fix to avoid the execution of the `onClose` prop in the first render.

If you see the demo as soon as you mount the notification component, the callback of the `onClose` prop is executed.
Look at console messages in: https://codesandbox.io/s/bug-onclose-sui-notification-tdcrx

The tasks that I have solved are:
- I group in a single function to close `handleClose` the `setIsOpen(false)` and the callback of the `onClose` prop.
- It eliminates dependencies of the `useEffect` that I consider unnecessary because they are values pointing to references through `useRef` (They do not change in the different renders).
- I delete the unnecessary `onClose` callback execution when the notification does not open. Code that was in the `else` of` if (show)`

### DIFF
--- a/components/molecule/notification/src/index.js
+++ b/components/molecule/notification/src/index.js
@@ -66,10 +66,9 @@ const MoleculeNotification = ({
     setShow(showFromProps)
   }, [showFromProps])
 
-  const getAutoCloseTime = useCallback(() => {
-    const time = AUTO_CLOSE_TIME[autoCloseTiming]
-    return time
-  }, [autoCloseTiming])
+  const autoCloseTimeInSeconds = AUTO_CLOSE_TIME[autoCloseTiming]
+    ? AUTO_CLOSE_TIME[autoCloseTiming]
+    : null
 
   const removeDelay = show => {
     const delay = show ? 1 : TRANSITION_DELAY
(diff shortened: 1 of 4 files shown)


components/molecule/notification/src/index.js
```diff
@@ -78,49 +77,37 @@ const MoleculeNotification = ({
...
 
-  const handleClickClose = e => {
-    setShow(false)
-  }
-
   useEffect(() => {
-    const autoCloseTimeInSeconds = getAutoCloseTime()
-
-    if (show && autoCloseTimeInSeconds) triggerAutoClose(autoCloseTimeInSeconds)
+    if (show) {
```
midudev: I think it was more readable with a single if, like before.

components/molecule/notification/src/index.js
```diff
@@ -78,49 +77,37 @@ const MoleculeNotification = ({
...
   }
 
+  const handleClose = useCallback(() => {
+    setShow(false)
+    onClose()
+  }, [onClose])
+
   const triggerAutoClose = useCallback(
     time => {
       autoCloseTimeout.current = setTimeout(() => {
```
midudev: `autoCloseTimeout.current = setTimeout(handleClose, time)`

components/molecule/notification/src/index.js
```diff
@@ -66,10 +66,9 @@ const MoleculeNotification = ({
     setShow(showFromProps)
   }, [showFromProps])
 
-  const getAutoCloseTime = useCallback(() => {
-    const time = AUTO_CLOSE_TIME[autoCloseTiming]
-    return time
-  }, [autoCloseTiming])
+  const autoCloseTimeInSeconds = AUTO_CLOSE_TIME[autoCloseTiming]
```
midudev: `const autoCloseTimeInSeconds = AUTO_CLOSE_TIME[autoCloseTiming]` should be enough. If not present, it will be `undefined` and thus a falsy value.

giodelabarrera: True. But I did so to differentiate it by giving a controlled value with `null`. I add that on a personal level I like to do this to differentiate these states.
`undefined`, not initialized and perhaps not identified.
`null`, initialized and identified with a null initial value.

but maybe in this situation there are no differences, I will change it. 😆

midudev (approved):

danivm (approved):

AitorRodriguez990 (approved):

quinwacca (approved):

desko27 (approved):

PR #882, opened 2019-11-25
chore(molecule/notification): add changelog and upgrade version

giodelabarrera:
Version upload manually because I got an error when execute `npm run release`.
The `Notification` component is already published in NPM with version` 1.13.0`

### DIFF
--- a/components/molecule/notification/package.json
+++ b/components/molecule/notification/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@s-ui/react-molecule-notification",
-  "version": "1.12.0",
+  "version": "1.13.0",
   "description": "",
   "main": "lib/index.js",
   "scripts": {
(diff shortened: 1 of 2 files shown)


components/molecule/notification/package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "@s-ui/react-molecule-notification",
-  "version": "1.12.0",
+  "version": "1.13.0",
```
davecarter, line 3: use a bug fix instead: `1.12.1`

giodelabarrera: Totally true. I'm sorry, it's already published with this version 😫

giodelabarrera: The commits I made were `fix` and `refactor`. I understand that the refactor has uploaded version?

danivm (approved):

PR #889, opened 2019-12-02
Style prop on Molecule CheckboxField

AitorRodriguez990:
## Description

Add a new prop to style checkboxes.

Related task: https://jira.scmspain.com/browse/SUIC-178

## Preview

image

image

### DIFF
--- a/components/molecule/checkboxField/src/index.js
+++ b/components/molecule/checkboxField/src/index.js
@@ -1,5 +1,6 @@
 import React from 'react'
 import PropTypes from 'prop-types'
+import cx from 'classnames'
 
 import MoleculeField from '@s-ui/react-molecule-field'
 import AtomCheckbox, {withCheckedValue} from '@s-ui/react-atom-checkbox'
@@ -18,11 +19,15 @@ const MoleculeCheckboxField = ({
   errorText,
   helpText,
   onChange,
+  styled,
   ...props
 }) => {
+  const className = cx(BASE_CLASS, {[`${BASE_CLASS}--styled`]: styled})
+
   const errorState = getErrorState(successText, errorText)
+
   return (
-    <div className={BASE_CLASS}>
+    <div className={className}>
       <MoleculeField
         name={id}
         label={label}
@@ -61,7 +66,10 @@ MoleculeCheckboxField.propTypes = {
   helpText: PropTypes.oneOfType([PropTypes.string, PropTypes.bool]),
 
   /** Boolean to decide if elements should be set inline */
-  inline: PropTypes.bool
+  inline: PropTypes.bool,
+
+  /* Style the checkbox component instead of use the default navigation styles */
+  styled: PropTypes.element
 }
 
 export default withCheckedValue(MoleculeCheckboxField)
--- a/components/molecule/checkboxField/src/index.js
+++ b/components/molecule/checkboxField/src/index.js
@@ -19,7 +19,7 @@ const MoleculeCheckboxField = ({
   errorText,
   helpText,
   onChange,
-  styled,
+  styled = false,
   ...props
 }) => {
   const className = cx(BASE_CLASS, {[`${BASE_CLASS}--styled`]: styled})
@@ -69,7 +69,7 @@ MoleculeCheckboxField.propTypes = {
   inline: PropTypes.bool,
 
   /* Style the checkbox component instead of use the default navigation styles */
-  styled: PropTypes.element
+  styled: PropTypes.bool
 }
 
 export default withCheckedValue(MoleculeCheckboxField)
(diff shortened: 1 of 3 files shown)


components/molecule/checkboxField/src/index.js
```diff
@@ -61,7 +66,10 @@ MoleculeCheckboxField.propTypes = {
   helpText: PropTypes.oneOfType([PropTypes.string, PropTypes.bool]),
 
   /** Boolean to decide if elements should be set inline */
-  inline: PropTypes.bool
+  inline: PropTypes.bool,
+
+  /* Style the checkbox component instead of use the default navigation styles */
+  styled: PropTypes.element
```
danivm: element or bool? 🤔

components/molecule/checkboxField/src/index.scss
```diff
@@ -1,12 +1,53 @@
...
+@import '~@schibstedspain/sui-theme/lib/settings-compat-v7/spacing';
+@import '~@schibstedspain/sui-theme/lib/settings-compat-v7/color';
+@import '~@schibstedspain/sui-theme/lib/settings-compat-v7/box-style';
 @import '~@s-ui/react-molecule-field/lib/index';
 @import '~@s-ui/react-atom-checkbox/lib/index';
 
+$bgc-atom-checkbox-styled: $c-white !default;
+$bgc-checked-atom-checkbox-styled: $c-primary !default;
+$c-atom-checkbox-styled: str-slice(#{$c-white}, 2);
+$bgi-atom-checkbox-styled: url('data:image/svg+xml;utf8,<svg xmlns="http://www.w3.org/2000/svg" viewBox="0 0 24 24"><path fill="%23#{$c-atom-checkbox-styled}" d="M9 16.2L4.8 12l-1.4 1.4L9 19 21 7l-1.4-1.4L9 16.2z"></path></svg>') !default;
```
midudev, line 11: I'm not sured about this. We should work in order to avoid this kind of things and let users put the icons using props instead.

rmoralp:
Given style is for AtomCheckbox, does it has sense to move this to AtomCheckbox component?
With this, we are adding AtomCheckbox css in other component and we are forced to use AtomCheckbox with MoleculeCheckboxField if we want to use a stylish checkbox, not enabling it in AtomCheckbox.

What do you think?

components/molecule/checkboxField/src/index.js
AitorRodriguez990: Fixed!

turolopezsanabria:
I agree with @rmoralp, it should style the Atom

AitorRodriguez990:
@rmoralp and @turolopezsanabria, I tried to do that but It's not possible to do without a label. Al least, I don't know-how.

components/molecule/checkboxField/src/index.scss
AitorRodriguez990: For the moment, I think nobody needs to change the icon. With this solution, each vertical will have the checkbox icon with his primary colour. If someone needs to change the icon, can iterate this component.

danivm (approved):

ruben-martin-lozano (approved):

components/molecule/checkboxField/src/index.scss
midudev: The same thing could be achieved using a normal svg. :P

AitorRodriguez990: Tell me more. How can I do that? I need to use a `background-image` on the `:before` label selector.

components/molecule/checkboxField/src/index.scss
```diff
@@ -1,12 +1,53 @@
 @import '~@schibstedspain/sui-theme/lib/index';
+@import '~@schibstedspain/sui-theme/lib/settings-compat-v7/spacing';
+@import '~@schibstedspain/sui-theme/lib/settings-compat-v7/color';
+@import '~@schibstedspain/sui-theme/lib/settings-compat-v7/box-style';
```
JoanClaret, line 4: we should try to avoid importing this `/settings-compat-v7` files which where meant to give compatibility to old sui components by defining deprecated variables with a non_standard format (for instance `$m-large` instead of `$m-l`) .  Do you really need them?

components/molecule/checkboxField/src/index.scss
```diff
@@ -1,12 +1,53 @@
...
   }
+
   label {
     cursor: pointer;
   }
+
+  &--styled {
+    .sui-MoleculeField-labelContainer {
+      .sui-AtomCheckbox {
+        display: none;
```
JoanClaret, line 29: too much specificity in this selector. This will increase the maintenance difficulty. is there any other way to do that?

JoanClaret:
This PR is breaking the component encapsulation and IMHO we should not style a component (`sui-AtomCheckbox`) from another one (`sui-MoleculeField`). This will generate unexpected side effects in the future.

components/molecule/checkboxField/src/index.scss
AitorRodriguez990: Done 😉

PR #890, opened 2019-12-02
fix(atom/checkbox): prevent execute onChangeFromProps as undefined

rmoralp:
Fix error when onChange is undefined.
image

EDIT: MoleculeSelect is not passing this prop

### DIFF
--- a/components/atom/checkbox/src/index.js
+++ b/components/atom/checkbox/src/index.js
@@ -21,8 +21,8 @@ const AtomCheckbox = ({
   id,
   name,
   disabled,
-  checked,
-  onChange: onChangeFromProps,
+  checked = false,
+  onChange: onChangeFromProps = () => {},
   errorState,
   ...props
 }) => {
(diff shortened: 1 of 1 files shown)


jordevo (approved):

midudev:
Then the `prop` is not required anymore. You should remove the `isRequired` property.

devictoribero (approved):

devictoribero:
@midudev This is something I always wanted to talk about. The fact this `prop` has a `defaultValue` doesn't mean it's not `required`, isn't it?

What I mean by this, is that it's still `required`, but it has a `defaultValue`. What does `require` mean for us? to pass it always every time we want to instance the component?

midudev:
@tureey `isRequired` means that the component should not work if the prop is not provided, as giving a value is mandatory. If a default value is provided, then it means the prop is not required anymore as you could ommit it and the component will work without problems.

devictoribero:
Ok! I always understood more in an API way. If the prop is required, but it has a default value, it's not provided by the user, but somehow its provided so it should work!

But I think that how you pointed out is how everyone understands it xD thanks @midudev !

rmoralp:
I have removed isRequired from proptypes.

carlosvillu (approved):

AgonisticKatai (approved):

midudev (approved):

PR #898, opened 2019-12-12
feat(organism/nestedCheckboxes): New Organism component NestedCheckboxes

AitorRodriguez990:
## Descripcion

Create a new, and the first 🎉, Organism component for nested checkboxes.

This Pull Request creates a new major version for `AtomCheckbox` and `MoleculeCheckboxField`.

This new component allows:

- Set N items that control the parent checked status.
- Set an icon to the parent when are items are checked, and another icon when only some items are. This is possible by `intermediate` state of `AtomCheckbox` and `MoleculeCheckboxField`.
- Use the `join` prop to remove the padding from items.
- Use the `defaultItemsVisibility` prop to show/hide items by default, and the `showItemsIcon` and `hideItemsIcon` prop to render a show/hide icon that will toggle the visibility of the items.

### All checked

image

### Some items checked

image

### No items checked

image

### Remove the items margin

image

#### Hide

image

#### Show

image

PR #901, opened 2019-12-18
feat(organism/nestedCheckboxes): changed to compound component

juanmendez90:
As requested in the previous version we changed the nestedCheckboxes organism  to a compound component
https://github.com/SUI-Components/sui-components/pull/898#discussion_r358686806

BREAKING CHANGE:
changed to compound component

### DIFF
--- a/components/organism/nestedCheckboxes/src/index.js
+++ b/components/organism/nestedCheckboxes/src/index.js
@@ -21,6 +21,7 @@ const OrganismNestedCheckboxes = ({
   labelParent = '',
   onChangeItem = () => {},
   onChangeParent = () => {},
+  onClickParent = () => {},
   showItems: showItemsProp = true,
   showItemsIcon: ShowItemsIcon
 }) => {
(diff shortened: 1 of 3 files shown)


components/organism/nestedCheckboxes/src/index.js
```diff
@@ -6,14 +6,17 @@ import MoleculeCheckboxField from '@s-ui/react-molecule-checkbox-field'
...
 const baseClass = 'sui-OrganismNestedCheckboxes'
 
-const checkItemIsChecked = ({checked}) => checked === true
+const checkItemIsChecked = ({props}) => {
+  const {checked} = props
+  return checked === true
+}
 
 const OrganismNestedCheckboxes = ({
+  children,
```
midudev, line 15: What happens if children is not provided? Shouldn't this be required?
And thus, avoid using all over the optional chaining?

juanmendez90: It wouldn't crash because I used optional chaining to avoid it, but you are right I will make it required.

adriadoz (approved):

nucliweb (approved):

jcger (approved):

PR #903, opened 2019-12-19
feat(atom/button): set accent hover color as a variable

nguasch:
In IJ we need to change the `background-color` and the `border-color` for the accent button on `hover`. For this, we need a specific variable only for this button (all buttons are building with an `@each`). For this reason I used an `@if`. Better solutions than this are welcome :)


beatrizip (approved):

PablitoGS (approved):

rmoralp (approved):
👏

PR #907, opened 2019-12-19
feat(molecule/select): Add selectSize prop to set input size

igncalleja:
This PR adds a `selectSize` prop to moleculeSelect so it can be passed to the select, which is using the atomInput. 

Right now the `size` prop only sets the size for the dropdownList, and not the main select.

### DIFF
--- a/components/molecule/select/src/index.js
+++ b/components/molecule/select/src/index.js
@@ -3,6 +3,7 @@ import PropTypes from 'prop-types'
 import cx from 'classnames'
 
 import {moleculeDropdownListSizes as SIZES} from '@s-ui/react-molecule-dropdown-list'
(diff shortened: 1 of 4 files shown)


javiauso (approved):

components/molecule/select/src/index.js
```diff
@@ -3,6 +3,7 @@ import PropTypes from 'prop-types'
 import cx from 'classnames'
 
 import {moleculeDropdownListSizes as SIZES} from '@s-ui/react-molecule-dropdown-list'
+import {inputSizes as SELECT_SIZES} from '@s-ui/react-atom-input'
```
arnau-rius, line 6: 👍

arnau-rius (approved):

tomasmax (approved):

jcger (approved):

PR #926, opened 2020-01-13
feat(atom/button): fill color of any value within svg

AgonisticKatai:
### Current behaviour
Captura de pantalla 2020-01-13 a las 12 43 44
Captura de pantalla 2020-01-13 a las 12 32 40

### Expected behaviour
Captura de pantalla 2020-01-13 a las 12 31 38

> By currently defining the style of the icons, only the color is applied to the `SVG` path. In some cases, the `SVG` includes `**ellipses**`, and you need to apply the same color as the rest of the icon

### DIFF
--- a/components/atom/button/src/index.scss
+++ b/components/atom/button/src/index.scss
@@ -196,7 +196,7 @@ $icon-stroke-color: currentColor !default;
       display: inline-flex;
     }
 
-    svg path {
+    svg {
       fill: $icon-fill-color;
       stroke: $icon-stroke-color;
     }


rmoralp (approved):

davidmartin84 (approved):

JoanClaret (approved):

jordevo (approved):

midudev (approved):
While this is a good fix, this could break some actual usage of svgs with buttons. I might be very careful with this. One fix could be using AtomIcon instead a plain svg for the button.

AgonisticKatai:
After pulling the thread on this behavior, we have seen that when using the AtomIcon, we still see that it does not apply the style correctly, and we have seen that the problem lies in how we generate the react components in our svg repo in motor, since by default it injects some inline styles, including a fill 119, 119, 119. This could be solved by adding an important !in the AtomIcon class that manages the color of the SVG, but I understand that the problem really is how we work in motor and that we should not apply any inline style to the svg's, I close this PR and open a technical-debt in engine to solve it.
Thx guys!

andresin87:
A softer fix could be add the other visible svg elements to the styling. This might not broke any existing current behavior with other verticals:

```
svg {
  circle, ellipse image, line, mesh, path, polygon, polyline, rect, text, use {
    fill: whtvr
    stroke: wthvr
  }
}
```

https://developer.mozilla.org/es/docs/Web/SVG/Element

PR #935, opened 2020-01-16
fix(molecule/selectField): fix on change propagation

juanmendez90:
Molecule/field was overriding the onChange prop sent to the molecule/select

### DIFF
--- a/components/molecule/selectField/src/index.js
+++ b/components/molecule/selectField/src/index.js
@@ -17,6 +17,7 @@ const MoleculeSelectField = ({
   inline,
   label,
   successText,
+  onChange,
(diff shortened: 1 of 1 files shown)


components/molecule/selectField/src/index.js
```diff
@@ -35,6 +36,7 @@ const MoleculeSelectField = ({
       inline={inline}
       label={label}
       name={id}
+      onChange={onChange}
```
adriadoz: It should be handleOnChange

adriadoz (approved):

jcger (approved):

PR #937, opened 2020-01-20
Fix/suic 258 border error and success in textareafield and autosuggestfield

arnau-rius:
### What does this PR do?

- add error and success state (and border color) in `molecule/autosuggestField` and `molecule/textField`. This is the actual behavior of input, select, etc components. 

#### Before
image

#### After
image

### Implementation

Inspiration from `molecule/select` component: https://github.com/SUI-Components/sui-components/blob/master/components/molecule/select/src/index.js

### Ticket

https://jira.scmspain.com/browse/SUIC-258

### DIFF
--- a/components/atom/textarea/src/index.js
+++ b/components/atom/textarea/src/index.js
@@ -8,8 +8,23 @@ const SIZES = {
   LONG: 'long'
 }
 
-const AtomTextarea = ({onChange, size, value, ...props}) => {
-  const className = cx(BASE_CLASS, `${BASE_CLASS}--${size}`)
+const ERROR_STATES = {
+  ERROR: 'error',
+  SUCCESS: 'success'
+}
+
+const getErrorStateClass = errorState => {
+  if (errorState) return `${BASE_CLASS}--${ERROR_STATES.ERROR}`
+  if (errorState === false) return `${BASE_CLASS}--${ERROR_STATES.SUCCESS}`
+  return ''
+}
+
+const AtomTextarea = ({onChange, size, value, errorState, ...props}) => {
+  const className = cx(
+    BASE_CLASS,
+    `${BASE_CLASS}--${size}`,
+    getErrorStateClass(errorState)
+  )
 
   const handleChange = ev => {
     const {value, name} = ev.target
@@ -36,7 +51,10 @@ AtomTextarea.propTypes = {
   onChange: PropTypes.func,
 
   /** Value (content) of the textarea */
-  value: PropTypes.string
+  value: PropTypes.string,
+
+  /** true = error, false = success, null = neutral */
+  errorState: PropTypes.bool
 }
 
 AtomTextarea.defaultProps = {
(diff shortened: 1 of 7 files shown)


components/atom/textarea/src/index.js
```diff
@@ -8,8 +8,23 @@ const SIZES = {
...
-const AtomTextarea = ({onChange, size, value, ...props}) => {
-  const className = cx(BASE_CLASS, `${BASE_CLASS}--${size}`)
+const ERROR_STATES = {
+  ERROR: 'error',
+  SUCCESS: 'success'
+}
+
+const getErrorStateClass = errorState => {
+  if (errorState) return `${BASE_CLASS}--${ERROR_STATES.ERROR}`
+  if (errorState === false) return `${BASE_CLASS}--${ERROR_STATES.SUCCESS}`
```
carlosvillu: False or Falsish ?!

Because `undefined !== false`

components/molecule/textareaField/src/index.js
```diff
@@ -19,6 +19,11 @@ const MoleculeTextareaField = WithCharacterCount(
     onChange,
     ...props
   }) => {
+    const hasErrors = (success, error) => {
+      if (error) return true
+      if (success) return false
+    }
```
carlosvillu: are you sure, that you need this function ?!

arnau-rius: From actual implementation in [molecule/select](https://github.com/SUI-Components/sui-components/blob/master/components/molecule/select/src/index.js), I guess that if error is true it doesn't mean that success is false and viceversa..

components/atom/textarea/src/index.js
arnau-rius: if errorState is truthy --> there is error. if errorState is false -->  there is not error and it is success. Otherwise there is not error and there is not success.

arnau-rius: it is not the best logic: merging truthy and strictly false in the same value (in my opinion). But there are other components that share this logic. See: https://github.com/SUI-Components/sui-components/blob/master/components/molecule/select/src/index.js#L26

carlosvillu (approved):

AgonisticKatai (approved):

components/atom/textarea/src/index.js
```diff
@@ -8,8 +8,23 @@ const SIZES = {
...
 }
 
-const AtomTextarea = ({onChange, size, value, ...props}) => {
-  const className = cx(BASE_CLASS, `${BASE_CLASS}--${size}`)
+const ERROR_STATES = {
+  ERROR: 'error',
+  SUCCESS: 'success'
+}
+
+const getErrorStateClass = errorState => {
```
davecarter: It might be easier to use `classNames` library to handle different class names based on some values

arnau-rius: it is using className in line 23..

davecarter (approved):

components/atom/textarea/src/index.js
arnau-rius: refactored for all components: molecule/textarea, molecule/autosuggest, atom/input, molecule/select

PR #943, opened 2020-01-27
fix(molecule/pagination): make molecule/pagination component render 

jordevo:
…`<ol>` tag instead of `<ul>`.

`<ol>` seems like a more appropriate HTML tag for a list made of elements on a pagination component. Not sure though why `<ul>` is used so raising this PR as a question.

### DIFF
--- a/components/molecule/pagination/src/index.js
+++ b/components/molecule/pagination/src/index.js
@@ -76,7 +76,7 @@ const MoleculePagination = ({
   const isHideNext = hideDisabled && !nextPage
 
   return (
-    <ul className={BASE_CLASS}>
+    <ol className={BASE_CLASS}>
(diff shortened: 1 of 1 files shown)


jcger (dismissed):

jcger (commented):
After researching a little bit, the right way to do this seems to either be ul or even better with [nav](https://stackoverflow.com/questions/27411894/can-i-use-the-nav-tag-for-pagination/27420775#27420775) so I'm gonna take back my approve

devictoribero:
Interesting... I always thought `<ol>` was the right way. Should we change the `<ul>` anyway but to `<nav>` then? @jcger @jordevo 

Is it a good practice to have more than one `<nav>` per page?

PR #944, opened 2020-01-27
feat(molecule/pagination): allow molecule/pagination to take a linkFactory prop and render links

jordevo:
`molecule/pagination` uses `atom/button` which by default renders `button` tags for every pagination element.

The idea behind this PR is to allow `molecule/pagination` to tell `atom/button` to render an anchor tag if a `renderLinks` prop and a linkFactory are passed.

### DIFF
--- a/components/molecule/pagination/src/index.js
+++ b/components/molecule/pagination/src/index.js
@@ -14,13 +14,19 @@ const BASE_CLASS = 'sui-MoleculePagination'
 const CLASS_PREV_BUTTON_ICON = 'sui-MoleculePagination-prevButtonIcon'
 const CLASS_NEXT_BUTTON_ICON = 'sui-MoleculePagination-nextButtonIcon'
 
-const PageButton = ({onSelectPage, page, design, ...props}) => {
+const PageButton = ({onSelectPage, page, design, linkFactory, ...props}) => {
   const _onSelectPage = e => {
     onSelectPage(e, {page})
   }
   return (
     <li className={`${BASE_CLASS}-item`}>
-      <AtomButton onClick={_onSelectPage} design={design} {...props} />
+      <AtomButton
+        onClick={_onSelectPage}
+        design={design}
+        {...props}
+        link
+        linkFactory={linkFactory(page)}
+      />
     </li>
   )
 }
@@ -31,7 +37,9 @@ PageButton.propTypes = {
   /** Current page selected */
   page: isValidPage,
   /** Design to be used for the page button. Design types 'solid', 'outline' or 'flat' */
-  design: PropTypes.string
+  design: PropTypes.string,
+  /** Factory used to create navigation links */
+  linkFactory: PropTypes.func
 }
 
 const MoleculePagination = ({
@@ -51,9 +59,7 @@ const MoleculePagination = ({
   nonSelectedPageButtonDesign,
   prevButtonDesign,
   nextButtonDesign,
-  link,
-  linkFactory,
-  href
+  linkFactory
 }) => {
   const paramsPagination = {
     page,
@@ -86,6 +92,8 @@ const MoleculePagination = ({
             onClick={handleClickPrev}
             design={prevButtonDesign}
             disabled={!prevPage}
+            link
+            linkFactory={linkFactory(page)}
           >
             {PrevButtonIcon && (
               <span className={CLASS_PREV_BUTTON_ICON}>
@@ -101,6 +109,7 @@ const MoleculePagination = ({
           page={page}
           design={selectedPageButtonDesign}
           onSelectPage={onSelectPage}
+          linkFactory={linkFactory}
         >
           {page}
         </PageButton>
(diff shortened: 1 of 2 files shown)


components/molecule/pagination/src/index.js
```diff
@@ -83,6 +92,7 @@ const MoleculePagination = ({
             onClick={handleClickPrev}
             design={prevButtonDesign}
             disabled={!prevPage}
+            {...linkProps(prevPage)}
```
rmoralp, line 112: If component does not receive linkFactory as a prop, it will break

jordevo: it won't, `linkProps` checks whether `linkFactory` is a function

devictoribero (approved):
As far as I can see, It's okay! Wait for more reviews tho, maybe someone can provide you with better feedback!

jcger (approved):

components/molecule/pagination/src/index.js
```diff
@@ -75,6 +78,12 @@ const MoleculePagination = ({
   const isHidePrev = hideDisabled && !prevPage
   const isHideNext = hideDisabled && !nextPage
 
+  const linkProps = pageNumber =>
+    typeof linkFactory === 'function' && {
+      link: true,
+      linkFactory: linkFactory(pageNumber)
```
kikoruiz: It's not a good idea to give to `linkFactory` the responsibility to manage de URL, I suggest to better move the following functionality to this component: https://github.com/SUI-Components/adevinta-spain-components/blob/master/components/pagination/basic/src/sui-pagination-item.js#L36. So, you'll need to define 3 new props: `linkFactory`, `createUrl` and `patternUrl` props (names can be modified).

kikoruiz: It's not a good idea to give to `linkFactory` the responsibility to manage the URL, I suggest to better move the following functionality to this component: https://github.com/SUI-Components/adevinta-spain-components/blob/master/components/pagination/basic/src/sui-pagination-item.js#L36. So, you'll need to define 3 new props: `linkFactory`, `createUrl` and `patternUrl` props (names can be modified).

kikoruiz (changes requested):

components/molecule/pagination/src/index.js
jordevo: just pushed with new changes, please review @kikoruiz :)

components/molecule/pagination/src/index.js
```diff
@@ -200,7 +230,20 @@ MoleculePagination.defaultProps = {
   selectedPageButtonDesign: 'solid',
   nonSelectedPageButtonDesign: 'flat',
   prevButtonDesign: 'flat',
-  nextButtonDesign: 'flat'
+  nextButtonDesign: 'flat',
+  createUrl: ({pageNumber, urlPattern}) => {
```
rmoralp: Could you decouple this default functions? Also, could you use default parameters?

components/molecule/pagination/src/index.js
```diff
@@ -75,6 +83,12 @@ const MoleculePagination = ({
   const isHidePrev = hideDisabled && !prevPage
   const isHideNext = hideDisabled && !nextPage
 
+  const linkProps = pageNumber =>
+    links && {
+      link: true,
+      linkFactory: linkFactory({href: createUrl({pageNumber, urlPattern})})
```
rmoralp, line 101: What about children and props?

components/molecule/pagination/src/index.js
kikoruiz: Yes, better move these `defaultProps` to default params in the component function.

components/molecule/pagination/src/index.js
jordevo: just fixed the default prop function :)

kikoruiz (approved):
Nice!

rmoralp (approved):

midudev (approved):

midudev:
With these changes in Fotocasa we might move to this component. :D

components/molecule/pagination/src/index.js
```diff
@@ -31,26 +33,44 @@ PageButton.propTypes = {
...
-  onSelectNext,
-  onSelectPrev,
+  onSelectNext = () => {},
+  onSelectPage = () => {},
+  onSelectPrev = () => {},
   page,
   totalPages,
-  showPages,
-  prevButtonText,
+  showPages = 10,
```
tureey: magic number  👀

components/molecule/pagination/src/index.js
```diff
@@ -31,26 +33,44 @@ PageButton.propTypes = {
...
+  prevButtonText = 'Previous',
```
tureey: magic text  👀

components/molecule/pagination/src/index.js
```diff
@@ -31,26 +33,44 @@ PageButton.propTypes = {
...
   prevButtonIcon: PrevButtonIcon,
-  nextButtonText,
+  nextButtonText = 'Next',
```
tureey: magic text  👀

devictoribero:
All thanks to @jordevo 👏

PR #946, opened 2020-01-28
feat(atom/popover): added showPopover value to control it from outside

devictoribero:
_As commented to @midudev I tested the behavior with @quinwacca so this behavior is totally retro compatible._

## Goal
Make possible to control when the `Popover` is open or not from its parent.

## Implementation
- Add a `showPopover ` prop that makes possible this behavior.
- Removed unneeded `<>`

### DIFF
--- a/components/atom/popover/src/index.js
+++ b/components/atom/popover/src/index.js
@@ -1,4 +1,4 @@
-import React, {useState, useRef, Suspense} from 'react'
+import React, {useEffect, useState, useRef, Suspense} from 'react'
 import PropTypes from 'prop-types'
 import {PLACEMENTS} from './config'
 
@@ -10,7 +10,6 @@ const DEFAULT_TRIGGER = 'legacy'
 const DEFAULT_DELAY = 0
 
 const Popover = React.lazy(() => import('reactstrap/lib/Popover'))
-let targetRef
 
 function AtomPopover({
   children,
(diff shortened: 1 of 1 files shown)


components/atom/popover/src/index.js
```diff
@@ -91,6 +90,8 @@ AtomPopover.propTypes = {
   content: PropTypes.element.isRequired,
   /** Popover children */
   children: PropTypes.node.isRequired,
+  /** Initial value for the show pop over */
```
midudev: Instead initial, should be something like `Controlled value for the showPopover state`.

midudev (approved):

jcger (approved):

components/atom/popover/src/index.js
```diff
@@ -19,15 +18,20 @@ function AtomPopover({
...
   onClose = () => {},
   onOpen = () => {},
-  placement = PLACEMENTS.BOTTOM
+  placement = PLACEMENTS.BOTTOM,
+  showPopover
 }) {
-  targetRef = useRef()
-  const [showPopover, setShowPopover] = useState(false)
+  const targetRef = useRef()
+  const [internalShowPopover, setInternalShowPopover] = useState(showPopover)
```
jordevo, line 25: being it a boolean, why not naming it `showInternalPopover` instead?

jordevo (approved):

components/atom/popover/src/index.js
devictoribero: any other idea about the naming @jcger @midudev? Would you understand it if we call it `showInternalPopover`or do you prefer as it is `internalShowPopover` ?

AgonisticKatai (approved):

PR #949, opened 2020-01-28
feat(atom/input): add new option to highlight input

quinwacca:
Add new states passed by props.

Before: 
We only could highlight the input with `success` or `error`, green or red. That's because we had a boolean prop (`errorState`), we couldn't have more than two values.

After:
We can highlight the input in different ways. Now, we need an `alert` orange border. So we add a string prop (`state`), which can receive `'success'`, `'error'` or `'alert'`.  In the future, more states can be added easily. Also, we exported from that component `inputStates`.

Captura de pantalla 2020-01-28 a las 16 14 46

Related Jiras:
https://jira.scmspain.com/browse/SUIC-262
https://jira.scmspain.com/browse/SUIC-263

### DIFF
--- a/components/atom/input/src/Input/Component/index.scss
+++ b/components/atom/input/src/Input/Component/index.scss
@@ -6,7 +6,6 @@ $bdrs-atom-input: $bdrs-none !default;
 
 $base-class: '.sui-AtomInput-input';
 $class-read-only: '#{$base-class}--readOnly';
-$c-atom-input-type: success $c-success, error $c-error, alert $c-alert;
 
 #{$base-class} {
   @extend %sui-atom-input-input;
@@ -53,7 +52,7 @@ $c-atom-input-type: success $c-success, error $c-error, alert $c-alert;
     }
   }
 
-  @each $state, $color in $c-atom-input-type {
+  @each $state, $color in $states-atom-input {
     &.sui-AtomInput-input--#{$state} {
       border-color: $color;
     }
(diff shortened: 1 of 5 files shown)


AgonisticKatai (approved):

components/atom/input/src/Input/Component/index.scss
```diff
@@ -52,7 +53,7 @@ $class-read-only: '#{$base-class}--readOnly';
     }
   }
 
-  @each $state, $color in $states-atom-input {
```
arnau-rius: IMHO since all form components use the variable: `$states-atom-input` that comes from this extend: `@extend %sui-atom-input-input;`, it is better to add the `alert` in the `$states-atom-input` variable instead of creating a new one (`$c-atom-input-type`), so the rest of components, for example: autosuggest, select, etc (that use the `$states-atom-input` variable) will also be able to have the `alert` if we add the corresponding tokens.

arnau-rius: otherwise if someday we change the states of the input in this variable `$states-atom-input` will not apply in this component. In other words, in this PR we are creating various versions ($states-atom-input and $c-atom-input-type) of the same 'truth' (the states of the input) taking into account that a lot of components use the input component.

arnau-rius: check for example here, `molecule/autosuggest` is using `$states-atom-input` variable: https://github.com/SUI-Components/sui-components/blob/329054791ae649ea0ae2bcbfc43860b9d34b4036/components/molecule/autosuggest/src/index.scss#L65

jordevo (approved):

components/atom/input/src/Input/Component/index.scss
quinwacca: Thanks for your help! I'm adding the `alert` state at `sui-theme` 👍

PR #951, opened 2020-01-29
Feature/molecule modal with url state

rmoralp:
I have created a HoC in order to make that a Modal can have a hash and open / close taking care of URL state.

- Every time it opens or closes, it should be saved in history
- onPopState changes it should open / close
- When users enter with modal hash, it should open

DEMO:
withUrlState


components/molecule/modal/src/HoC/WithUrlState.js
```diff
@@ -0,0 +1,76 @@
+import React, {useEffect, useState} from 'react'
+import PropTypes from 'prop-types'
+
+const getUrlWithHash = hashName => {
```
devictoribero: What about `createUrlWithHash` or `generateUrlWithHash`?

tureey: give it a default value `hashName  = ''`

rmoralp: When you call this method you should be aware of how / where this method get the url with hash. Using the names you propose, you are coupling the implementation to the name.

components/molecule/modal/src/HoC/WithUrlState.js
```diff
@@ -0,0 +1,76 @@
...
+      setIsPopStateChange(true)
+
+      if (hasHash) openModalTrigger()
+      else onClose()
+    }
+
+    const _handleChangeState = () => {
+      if (isPopStateChange) return setIsPopStateChange(false)
+
+      const nextUrl = isOpen ? getUrlWithHash(hash) : getUrlWithHash()
```
devictoribero: this looks confusing to me. We're calling a `getUrlWithHash()` without hash.

components/molecule/modal/src/HoC/WithUrlState.js
```diff
@@ -0,0 +1,76 @@
...
+
+      window.history.pushState(
```
devictoribero: This is less than 80 chars, make it be in 1 line manually and once you save, this will stay in one line I think

demo/molecule/modal/demo/MobileFitContentModal.js
```diff
@@ -0,0 +1,52 @@
...
+export default class MobileFitContentModal extends React.Component {
+  constructor(props) {
+    super(props)
+    this.state = {
+      open: false
+    }
+  }
+
+  handleOpenModal = () => {
+    this.setState({
```
devictoribero, line 19: again, make it be in 1 line manually and save, it should stay like that because it's less than 80 charrs

demo/molecule/modal/demo/MobileFitContentModal.js
```diff
@@ -0,0 +1,52 @@
...
+      open: true
+    })
+  }
+
+  handleCloseModal = () => {
```
devictoribero, line 24: What about :
```
handleCloseModal = () => this.setState({open: false})
```

devictoribero (approved):
IMHO looks good, but let's see what others have to say. Good job!

components/molecule/modal/src/HoC/WithUrlState.js
rmoralp: Why? What do you propose?

demo/molecule/modal/demo/MobileFitContentModal.js
rmoralp: It does not apply to this PR

demo/molecule/modal/demo/MobileFitContentModal.js
rmoralp: It does not apply to this PR

components/molecule/modal/src/HoC/WithUrlState.js
devictoribero: Since we can generate the URL without hash, what about calling it just `getUrl`?

components/molecule/modal/src/HoC/WithUrlState.js
```diff
@@ -0,0 +1,69 @@
...
+
+    const _handlePopState = ev => {
+      const hasHash = checkHash(hash)
+      setIsPopStateChange(true)
+
+      if (hasHash) openModalTrigger()
+      else onClose()
+    }
+
+    const _handleChangeState = () => {
```
jordevo: handleStateChange perhaps?

components/molecule/modal/src/HoC/WithUrlState.js
```diff
@@ -0,0 +1,69 @@
...
+ * - useLocation
+ */
+export default BaseComponent => {
+  const displayName = BaseComponent.displayName
+
+  if (typeof window === 'undefined') return BaseComponent
+
+  const WithUrlState = props => {
+    const {hash, openModalTrigger, onClose, ...rest} = props
+    const {isOpen} = rest
```
jordevo, line 18: why don't you get it from props destructuring as in L25?

AgonisticKatai (approved):

jordevo (approved):

components/molecule/modal/src/HoC/WithUrlState.js
rmoralp: I destructure `rest` in BaseComponent so it should keep there.

desko27 (approved):
Great work with nice docs and demos, congrats mate 🎉

components/molecule/modal/src/HoC/WithUrlState.js
```diff
@@ -0,0 +1,69 @@
...
+    }, [hash, isOpen]) // eslint-disable-line
+
+    useEffect(() => {
+      if (isPopStateChange) openModalTrigger()
+
+      window.onpopstate = _handlePopState
+      return () => {
+        window.onpopstate = emptyFn
+      }
+    }, []) // eslint-disable-line
```
desko27: Also, instead of not placing `isPopStateChange` as a dependency, we may just use `hash` and do something like:
```js
if (checkHash(hash)) openModalTrigger()
```

That way `eslint-disable-line` comment becomes avoidable, which is desired.

rmoralp: It must remains because hash not only changes when popState event triggered.
`_handleStateChange` should not modify history if hash changes only in popState (when user click next/prev arrows in browser)

PR #957, opened 2020-01-30
feat(molecule/field): add new prop: alertText

quinwacca:
Add a new text type, so we can apply an alert highlight to `molecule/inputField`

Captura de pantalla 2020-01-30 a las 10 30 16

### DIFF
--- a/components/molecule/field/src/index.js
+++ b/components/molecule/field/src/index.js
@@ -80,7 +80,7 @@ const MoleculeField = ({
       )}
       <div className={CLASS_INPUT_CONTAINER}>
         {!inline && extendedChildren}
-        {(successText || errorText || alertText) && (
+        {typeValidationText && (
           <AtomValidationText
             type={typeValidationText}
             text={statusValidationText}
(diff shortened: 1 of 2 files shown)


davidmartin84 (approved):

AgonisticKatai (approved):

components/molecule/field/src/index.js
```diff
@@ -72,7 +80,7 @@ const MoleculeField = ({
       )}
       <div className={CLASS_INPUT_CONTAINER}>
         {!inline && extendedChildren}
-        {(successText || errorText) && (
+        {(successText || errorText || alertText) && (
```
devictoribero: What about creating a variable before the return called:
`const hasValidationText = successText || errorText || alertText`
so we can reduce the cognitive load?

devictoribero (approved):

components/molecule/field/src/index.js
quinwacca: In fact, we can check the value of `typeValidationText`! Thanks!

nucliweb (approved):

PR #961, opened 2020-01-30
feat(atom/textarea): add state prop

quinwacca:
As in `atom/input` component, we add new states passed by props.

Before:
We only could highlight the `text area` with success or error, green or red. That's because we had a boolean prop (errorState), we couldn't have more than two values.

After:
We can highlight the `text area` in different ways. Now, we need an alert orange border. So we add a string prop (state), which can receive 'success', 'error' or 'alert'. In the future, more states can be added easily. Also, we exported from that component atomTextareaStates (this time we really do😄 !)

Captura de pantalla 2020-01-30 a las 16 30 18

### DIFF
--- a/components/atom/textarea/README.md
+++ b/components/atom/textarea/README.md
@@ -34,4 +34,12 @@ import AtomTextarea from '@s-ui/react-atom-textarea'
 <AtomTextarea placeholder="Write something cool here..." />
 ```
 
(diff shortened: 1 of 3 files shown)


jordevo (approved):
👏

jcger (approved):

PR #963, opened 2020-01-31
Feat/collapsible

nucliweb:
## In this Pull Request

- Added `alignContainer` prop to change the `text-align` in container element
- Removed unnecessary  CSS property and Sass var

### DIFF
--- a/components/molecule/collapsible/src/index.js
+++ b/components/molecule/collapsible/src/index.js
@@ -10,10 +10,15 @@ const BUTTON_CLASS = `${BASE_CLASS}-btn`
 const BUTTON_CONTENT_CLASS = `${BUTTON_CLASS}-content`
 const ICON_CLASS = `${BASE_CLASS}-icon`
 const MIN_HEIGHT = 100 // px
+const CONTENT_ALIGN = {
+  CENTER: 'center',
+  RIGHT: 'right'
+}
 
 const MoleculeCollapsible = ({
   onClose = () => {},
   onOpen = () => {},
+  alignContainer,
   children,
   height = MIN_HEIGHT,
   icon,
@@ -42,7 +47,7 @@ const MoleculeCollapsible = ({
     setShowButton(offsetHeight >= height)
     setMaxHeight(offsetHeight)
   }, [height])
-
+  console.log('alignContainer', alignContainer)
   const wrapperClassName = cx(`${BASE_CLASS}`, {
     [`${BASE_CLASS}--withGradient`]: withGradient,
     [COLLAPSED_CLASS]: collapsed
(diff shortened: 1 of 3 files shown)


devictoribero (approved):

adriadoz (approved):

arnau-rius:
Is this breaking backwards compatibility? The actual default behavior is aligning to the left. So, INMHO it should have as default value align left.

components/molecule/collapsible/src/index.js
```diff
@@ -90,6 +95,10 @@ const MoleculeCollapsible = ({
 MoleculeCollapsible.displayName = 'MoleculeCollapsible'
 
 MoleculeCollapsible.propTypes = {
+  /**
+   * Container align center || right
+   */
+  alignContainer: PropTypes.oneOf(Object.values(CONTENT_ALIGN)),
```
kikoruiz, line 101: I don't see the need to do it, we already have a Sass variable that allows us to customise it from our sites.

davecarter: I'd rather use a finite number of JS props to define components customizable properties instead of modifying Sass values. Feels more declarative to me.

components/molecule/collapsible/src/index.js
```diff
@@ -10,10 +10,15 @@ const BUTTON_CLASS = `${BASE_CLASS}-btn`
 const BUTTON_CONTENT_CLASS = `${BUTTON_CLASS}-content`
 const ICON_CLASS = `${BASE_CLASS}-icon`
 const MIN_HEIGHT = 100 // px
+const CONTENT_ALIGN = {
```
davecarter, line 13: you can expose this value out of the component

```javascript

export {CONTENT_ALIGN}
```
...so then you'll be able to use it:

```javascript

import MoleculeCollapsible, {CONTENT_ALIGN} from ' @s-ui/react-molecule-collapsible' 
```

components/molecule/collapsible/src/index.js
kikoruiz: Yep, I was wrong, in this case we could want more than one collapsible in our sites aligned differently, so it makes a lot of sense 👍

kikoruiz (approved):

ivanmlaborda (approved):

davecarter (approved):

nucliweb:
> Is this breaking backwards compatibility? The actual default behavior is aligning to the left. So, INMHO it should have as default value align left.

It's true, I leave it as it was to maintain compatibility with the verticals that have used the variable `$trs-collapsible-content` in their theme, thanks man.

ivanmlaborda (approved):

kikoruiz (approved):

AgonisticKatai (approved):

davecarter (approved):

PR #964, opened 2020-01-31
feat(molecule/select): add state prop

quinwacca:
Add new states passed by props.

Before:
We only could highlight the input with success or error, green or red. That's because we had a boolean prop (errorState), we couldn't have more than two values.

After:
We can highlight the **select** in different ways. Now, we need an alert orange border. So we add a string prop (state), which can receive 'success', 'error' or 'alert'. In the future, more states can be added easily. Also, we exported from that component **moleculeSelectStates**.

Captura de pantalla 2020-01-31 a las 11 29 01

### DIFF
--- a/components/molecule/select/src/index.js
+++ b/components/molecule/select/src/index.js
@@ -16,9 +16,10 @@ const BASE_CLASS = `sui-MoleculeSelect`
 const CLASS_FOCUS = `${BASE_CLASS}--focus`
 const CLASS_DISABLED = `is-disabled`
 
-const ERROR_STATES = {
+const SELECT_STATES = {
   ERROR: 'error',
-  SUCCESS: 'success'
+  SUCCESS: 'success',
+  ALERT: 'alert'
 }
 
 const ENABLED_KEYS = ['Enter', 'ArrowDown', 'ArrowUp']
@@ -38,6 +39,7 @@ const MoleculeSelect = props => {
     onToggle,
     children,
     errorState,
+    state,
     disabled,
     keysSelection,
     refMoleculeSelect: refMoleculeSelectFromProps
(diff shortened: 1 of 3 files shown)


components/molecule/select/src/index.js
```diff
@@ -230,3 +236,4 @@ MoleculeSelect.defaultProps = {
 export default withOpenToggle(MoleculeSelect)
 export {SIZES as moleculeSelectDropdownListSizes}
 export {SELECT_SIZES as moleculeSelectSizes}
+export {SELECT_STATES as moleculeSelectStates}
```
davecarter, line 239: 👍

demo/molecule/select/demo/index.js
```diff
@@ -124,6 +124,26 @@ const Demo = () => (
...
+      <h3>With state</h3>
+      <p>
+        State to highlight that can be <code>success</code>, <code>error</code>{' '}
+        or <code>alert</code>
+      </p>
+      <MoleculeSelect
+        placeholder="Select a Country..."
+        onChange={(_, {value}) => console.log(value)}
+        iconArrowDown={<IconArrowDown />}
+        state="alert"
```
davecarter: if you expose `SELECT_STATES`, it would be great to have an example of this implementation instead of an `"alert"` string:

```javascript

state={SELECT_STATES.ALERT}
```

quinwacca: True! Thanks!

quinwacca: Fixed!

AgonisticKatai (approved):

davecarter (approved):

davidmartin84 (approved):

PR #973, opened 2020-02-04
Feat/autosuggest button input

a-ferrante:
Add prop to be able to use the new input with a submit button

### DIFF
--- a/components/atom/input/src/Input/Features/WithButton/index.scss
+++ b/components/atom/input/src/Input/Features/WithButton/index.scss
@@ -1,5 +1,6 @@
 .sui-AtomInput-withButton {
   display: flex;
+  width: 100%;
 
   &-input {
     width: 100%;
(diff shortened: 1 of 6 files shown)


jordevo (approved):

turolopezsanabria:
thanks, The "focus" state is a bit weird, the field should occupy as much as possible of the full width of the container, right?

image

a-ferrante:
> thanks, The "focus" state is a bit weird, the field should occupy as much as possible of the full width of the container, right?
> 
> image

The issue is being handle in this PR also, but since there are changes that involves two components the builder its not able to include the changes required on the input to fix the issue. 
Not sure if the changes should be done in seperate PR since the changes are highly related 

This is the result of running all changes together!
Captura de pantalla 2020-02-04 a las 11 01 40

jcger (approved):

PR #978, opened 2020-02-06
Feat/select popover size

danivm:
Add sizes to `SelectPopover`

🔎   [Demo](https://sui-components-git-feat-select-popover-size.schibstedspain.now.sh/workbench/molecule/selectPopover/demo)

### DIFF
--- a/components/molecule/selectPopover/src/_settings.scss
+++ b/components/molecule/selectPopover/src/_settings.scss
@@ -1,4 +1,9 @@
-$bdrs-select-popover: 16px !default;
+$bdrs-select-popover: 20px !default;
 $bxsh-select-popover: 0 1px 4px 0 rgba(0, 0, 0, 0.24) !default;
 $mw-select-popover: 250px !default;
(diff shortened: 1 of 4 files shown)


components/molecule/selectPopover/src/_settings.scss
```diff
@@ -1,4 +1,9 @@
-$bdrs-select-popover: 16px !default;
+$bdrs-select-popover: 20px !default;
```
midudev, line 1: If we change the default value, this could be a breaking change if somebody is using it relying on the default value. Couldn't you just change this value in your theme?

danivm: This change is necessary because with the new sizes `16px` is not enought. 
Anyway, nobody is using this component yet (only us).

adriadoz (approved):

AitorRodriguez990 (approved):

midudev (approved):

giodelabarrera (approved):

PR #983, opened 2020-02-13
Feat/add on blur callback

arnau-rius:
- add onBlur callback to `molecule/autosuggest` and `atom/textArea`


components/atom/textarea/src/index.js
```diff
@@ -21,6 +21,7 @@ const DEFAULT_PROPS = {
 
 const AtomTextarea = ({
   onChange = DEFAULT_PROPS.onChange,
+  onBlur = DEFAULT_PROPS.onBlur,
```
oriolpuig, line 25: This onBlur is not defined at DEFAULT_PROPS constant, isn't it?

components/atom/textarea/src/index.js
```diff
@@ -14,14 +14,10 @@ const TEXTAREA_STATES = {
   ALERT: 'alert'
 }
 
-const DEFAULT_PROPS = {
-  size: SIZES.SHORT,
-  onChange: () => {}
-}
-
 const AtomTextarea = ({
-  onChange = DEFAULT_PROPS.onChange,
```
arnau-rius, line 23: for me it is better to define the defaultProps as default arguments (JS native way) but I don't know why this is the only component in sui world that defines the default props in such way. Ideas?

oriolpuig: I prefer this one instead of defaultProps.

oriolpuig (approved):

components/atom/textarea/src/index.js
davecarter: Using default props makes components API more resilient and is kinda more declarative...
So, what else? 😄

PR #990, opened 2020-02-18
fix(atom/button): hover styles when device has hover

adriadoz:
fix(atom/button): hover styles when device has hover

### DIFF
--- a/components/atom/button/src/index.scss
+++ b/components/atom/button/src/index.scss
@@ -110,9 +110,13 @@ $icon-stroke-color: currentColor !default;
 @mixin button-focused {
   &:not(#{$base-class}--disabled) {
     &#{$base-class}--focused,
-    &:active,
-    &:hover {
+    &:active {
       @content;
+    }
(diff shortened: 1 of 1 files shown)


desko27 (commented):
I love it but it doesn't work for IE, does it? 😭

https://caniuse.com/#feat=mdn-css_at-rules_media_hover

adriadoz:
> I love it but it doesn't work for IE, does it? 😭
> 
> https://caniuse.com/#feat=mdn-css_at-rules_media_hover

Yes, but if we have to choose it is better to not have hover in IE or to avoid double touching in mobile devices?

desko27:
@adriadoz That's a convincing point and I'm not a big fan of supporting IE after all, let's see what others think 😄👍🏻

desko27 (approved):

rackso (approved):

midudev (approved):

nucliweb (approved):

PR #991, opened 2020-02-18
Feature/molecule select popover add open events

desko27:
## 🎯 Why?

We need some events to be aware of popover opening or closing, which is a different concept than 
 onAccept/onCancel events.

## 🤔 How?

I think this is a good chance to make `useEffect` shine a bit, instead of manually tracking down all `isOpen` updates 😄

## ↪️ About usePrevious
📌 https://usehooks.com/usePrevious/


giodelabarrera (approved):
⚓️⚛️

nucliweb (approved):

PR #998, opened 2020-02-21
Feature/atom checkbox add native style option

desko27:
## Add native style option

Adds the ability to go with the browser's native checkbox which was lost when we got to v2 major. This way we can update major version of this dependency on other components while still keeping the original behaviour/style. Also, it's a reasonable feature IMHO.

There are two ways of getting the native checkbox:
- Setting the new `isNative` prop to `true`
- Not setting any of the custom icon props (required for custom checkbox)

Screen-Recording-2020-02-21-at-1

### DIFF
--- a/components/atom/checkbox/src/index.js
+++ b/components/atom/checkbox/src/index.js
@@ -79,7 +79,7 @@ AtomCheckbox.propTypes = {
   checked: PropTypes.bool,
 
   /* AtomIcon when checkbox is checked */
-  checkedIcon: PropTypes.elementType.isRequired,
+  checkedIcon: PropTypes.elementType,
 
   /* Mark the input as intermediate */
   intermediate: PropTypes.bool,
--- a/components/atom/checkbox/src/index.js
+++ b/components/atom/checkbox/src/index.js
@@ -1,41 +1,61 @@
-import React from 'react'
+import React, {useRef} from 'react'
 import PropTypes from 'prop-types'
 import cx from 'classnames'
 
 const BASE_CLASS = 'sui-AtomCheckbox'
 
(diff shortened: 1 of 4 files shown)


components/atom/checkbox/src/index.js
```diff
@@ -1,41 +1,61 @@
...
   })
 
+  // Keep native indeterminate property updated every render
+  updateNativeIndeterminate()
+
   return (
     <label className={className}>
-      {checked && <CheckedIcon />}
-      {intermediate && !checked && <IntermediateIcon />}
+      {!isNative && checked && <CheckedIcon />}
```
adriadoz, line 49: just as a suggestion if using `isIntermediate`  it would be nicer to use `isChecked`

adriadoz (approved):

danivm (approved):

giodelabarrera (approved):

components/atom/checkbox/src/index.js
desko27: I totally agree the `is` prefix is cooler 👍

Unfortunately the currently exposed props were already named `checked` and `intermediate`, which cannot be changed without releasing a major,  while `isIntermediate` is just a new inner constant that is used to disable intermediate behaviour when `checked` prop is active (there may be better names for it 🤣). This makes `checked` prop to become the absolute priority if set along with `intermediate` prop (e.g. by accident).

We could rename outer `checked` prop to `isChecked` just for inner usage, but I think it's convenient to avoid that kind of [mental mapping](https://github.com/ryanmcdermott/clean-code-javascript#avoid-mental-mapping) and keep original prop names the same when possible.

PR #1008, opened 2020-03-02
feat(molecule/photoUploader): fix rejected files callback

quinwacca:
Improve rejected file callback

Before: only executing when a not accepted file is dropped.

Now: executed in four cases
- when a not accepted file is dropped.
- dropped file exceeds the maximum size.
- dropped file is repeated.
- added file fails when loading.
Also, a reason string will be sent in the callback with the file.

This callback will be useful to track user's behavior.

### DIFF
--- a/components/molecule/photoUploader/src/fileTools.js
+++ b/components/molecule/photoUploader/src/fileTools.js
@@ -10,7 +10,7 @@ export const filterValidFiles = ({
   files,
   filesToBeFiltered,
   acceptedFileMaxSize,
-  callbackPhotosRejected,
+  handlePhotosRejected,
   setMaxSizeError,
   setMaxPhotosError,
   allowUploadDuplicatedPhotos,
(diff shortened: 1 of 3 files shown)


components/molecule/photoUploader/src/fileTools.js
```diff
@@ -2,53 +2,78 @@ import {formatToBase64} from './photoTools'
...
-  DEFAULT_IMAGE_ROTATION_DEGREES
+  DEFAULT_IMAGE_ROTATION_DEGREES,
+  REJECT_FILES_REASONS
 } from './config'
 
 export const filterValidFiles = ({
   files,
   filesToBeFiltered,
   acceptedFileMaxSize,
+  callbackPhotosRejected,
```
giodelabarrera: Maybe you can use a naming with handle: `handlePhotosReject`

quinwacca: Renamed! :)

davidmartin84 (approved):

arnau-rius (approved):

PR #1027, opened 2020-03-17
feat(atom/card): Add media prop optional in atom card

izeller:
We need a Card without image. For that reason, the media prop in atom card should be optional

### DIFF
--- a/components/atom/card/src/index.js
+++ b/components/atom/card/src/index.js
@@ -45,9 +45,13 @@ const AtomCard = ({
       onClick={redirectToHref}
       onKeyDown={redirectOnEnter}
     >
-      <div className={CLASS_MEDIA}>
-        <Media />
-      </div>
+      <>
+        {Media && (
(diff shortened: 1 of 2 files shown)


danivm (approved):

davidmartin84 (approved):

components/atom/card/src/index.js
```diff
@@ -45,9 +45,13 @@ const AtomCard = ({
       onClick={redirectToHref}
       onKeyDown={redirectOnEnter}
     >
-      <div className={CLASS_MEDIA}>
-        <Media />
-      </div>
+      <>
```
midudev: I think Fragment is not required in this case.

midudev (approved):

jcger (approved):

components/atom/card/src/index.js
izeller: Removed!

ferransimon (approved):

PR #1042, opened 2020-03-23
feat(atom/textarea): mod styles

andresin87:
ADD style variables to customize the proper styles for Fotocasa purposes
- TASK: [https://jira.scmspain.com/browse/SUIC-298](https://jira.scmspain.com/browse/SUIC-298)
- BUILD: [https://sui-components-git-feature-suic-298.schibstedspain.now.sh/workbench/atom/textarea/demo](https://sui-components-git-feature-suic-298.schibstedspain.now.sh/workbench/atom/textarea/demo)

### DIFF
--- a/components/atom/textarea/src/index.scss
+++ b/components/atom/textarea/src/index.scss
@@ -5,6 +5,7 @@ $short-num-lines: 5 !default;
 $long-num-lines: 7 !default;
 
 $p-atom-textarea: 2px !default;
+$pm-atom-textarea: $p-m !default; // deprecated
 $pt-atom-textarea: $p-m !default;
 $pl-atom-textarea: $p-m !default;
 $fz-atom-textarea: $fz-base !default;
--- a/components/atom/textarea/src/index.scss
+++ b/components/atom/textarea/src/index.scss
@@ -4,7 +4,9 @@
 $short-num-lines: 5 !default;
 $long-num-lines: 7 !default;
 
-$pm-atom-textarea: $p-m !default;
+$p-atom-textarea: 2px !default;
+$pt-atom-textarea: $p-m !default;
+$pl-atom-textarea: $p-m !default;
 $fz-atom-textarea: $fz-base !default;
 $lh-atom-textarea: $lh-l !default;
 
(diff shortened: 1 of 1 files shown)


components/atom/textarea/src/index.scss
```diff
@@ -4,7 +4,9 @@
 $short-num-lines: 5 !default;
 $long-num-lines: 7 !default;
 
-$pm-atom-textarea: $p-m !default;
+$p-atom-textarea: 2px !default;
```
midudev, line 7: This is breaking the component API, we should try to keep previous compatibility while adding the new feature. Otherwise, we should publish a major and could be problematic to spread the new version. :)

components/atom/textarea/src/index.scss
```diff
@@ -4,7 +4,10 @@
...
+$pm-atom-textarea: $p-m !default; // deprecated
+$pt-atom-textarea: $p-m !default;
```
midudev: `$pt-atom-textarea` and `$pl-atom-textarea` default value should be `$pm-atom-textarea` (the deprecated one). As if it was used before, we should keep the value used.

components/atom/textarea/src/index.scss
```diff
@@ -4,7 +4,10 @@
 $short-num-lines: 5 !default;
 $long-num-lines: 7 !default;
 
-$pm-atom-textarea: $p-m !default;
+$p-atom-textarea: 2px !default;
```
midudev, line 7: I think the default one for this should be `0`, as it was before.

carlosvillu (approved):

components/atom/textarea/src/index.scss
andresin87: Done. there was an error on `pl-whatever` i actually update

components/atom/textarea/src/index.scss
midudev: Then, better using `initial`.

midudev (approved):

components/atom/textarea/src/index.scss
andresin87: Finally we 2 decided to set it to 2px cuz initial is not returning us the desired behavior

PR #1043, opened 2020-03-23
feat(atom/card): add new highlight bgc hover variable

jordevo:
This will allow us to overwrite the background colour on hover of a card that's highlighted.

Kapture 2020-03-23 at 12 54 38

### DIFF
--- a/components/atom/card/src/index.scss
+++ b/components/atom/card/src/index.scss
@@ -7,6 +7,7 @@ $p-atom-card-small: 0 !default;
 $bgc-atom-card-hover: transparent !default;
 $bxsh-atom-card-hover: none !default;
 $c-atom-card-highlight: $c-highlight !default;
+$bgc-atom-card-highlight-hover: $bgc-atom-card-hover !default;
 $w-atom-card-media: 33% !default;
 $mr-atom-card-media: 0 !default;
 $mq-responsive-breakpoint-name: 'm' !default;
@@ -50,6 +51,11 @@ $class-media: '#{$base-class}-media';
 
   &--highlight {
     background: $c-atom-card-highlight;
+    @media (hover: hover) {
+      &:hover {
+        background: $bgc-atom-card-highlight-hover;
+      }
+    }
   }
 
   &--vertical {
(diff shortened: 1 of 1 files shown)


andresin87 (approved):

midudev (approved):

components/atom/card/src/index.scss
```diff
@@ -7,6 +7,7 @@ $p-atom-card-small: 0 !default;
 $bgc-atom-card-hover: transparent !default;
 $bxsh-atom-card-hover: none !default;
 $c-atom-card-highlight: $c-highlight !default;
+$bgc-atom-card-highlight-hover: $bgc-atom-card-hover !default;
```
rmoralp: In order to not break other verticals that could be using Highlight and does not want to have different hover it should be `$c-atom-card-highlight` by default. Then we should override it in our theme, isn't it?

components/atom/card/src/index.scss
```diff
@@ -50,6 +51,11 @@ $class-media: '#{$base-class}-media';
 
   &--highlight {
     background: $c-atom-card-highlight;
+    @media (hover: hover) {
```
nucliweb, line 54: 🤩

nucliweb (approved):

components/atom/card/src/index.scss
midudev: Good catch. I'm guessing if the `hover` is already changing for the highlight with the hover style. Then the default value should be `transparent !default;`.

jordevo: Currently, `atom/card` is using `$bgc-atom-card-hover` as its `bgc` when on hover on a highlighted card, so using `$c-atom-card-highlight` would actually be a breaking change.

midudev: What Rafa is pointing is that the value of `$bgc-atom-card-hover` could have changed to other value making that the hover of the highlight could change as well unexpectedly.

So, the default value of the hover should be something that mimics the previous behaviour.

rmoralp: I understand it bad.
Now highlight hover removes the highlight, so Jordi does not break it. Just keep the bugfeature 😅 
If we want to persists this bugfeature Jordi is doing it good. If not, Jordi can use `$c-atom-card-highlight` by default in hover.

@midudev what do you think?

midudev: I would recommend to keep the bugfeature and add a comment about it in the code. :)

jordevo: done

PR #1051, opened 2020-03-26
docs(examples): add themes to complex demos as well

midudev:
This PR allows demos styles to be imported by every theme.

Before:
image

Now:
image


andresin87 (approved):

tomasmax (approved):

jcger (approved):
should the `import './styles.scss'` that was done in this demo/index.js files be removed then?

midudev:
@jcger I removed it before in this PR and did a rollback as the default styles were not shown as expected. I think there's some corner case in the studio to handle this when no theme is selected. Before removing, I would like to investigate further to avoid problems.

Keeping it the "only thing" that's happening is that the styles are being added twice.

carlosvillu (approved):

PR #1061, opened 2020-03-31
Feature/add node label to radio

ferransimon:
We need to be able to use custom Labels for our RadioButtonField. In some cases, we would like to add extra components that are not part of the interactive label of the radioButton itself like in this i.e 
image

where the "+ info" part should not trigger the onchange of the radiobutton. This PR tries to solves that by inverting the dependency of the AtomLabel to the client of the RadioButtonField.

### DIFF
--- a/components/molecule/field/src/index.scss
+++ b/components/molecule/field/src/index.scss
@@ -24,7 +24,6 @@ $atomHelpText-class: '.sui-AtomHelpText';
 
       #{$atomLabel-class} {
         align-items: center;
-        margin-left: $ml-molecule-field-label;
       }
     }
   }
--- a/components/molecule/field/src/index.js
+++ b/components/molecule/field/src/index.js
@@ -23,20 +23,33 @@ const MoleculeLabel = ({
   name,
   onClickLabel
 }) => {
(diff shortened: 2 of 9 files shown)


andresin87:
Nice! could you please add the same tests that molecule/checkboxField have to the behavior of this component?

andresin87 (approved):

jcger:
Please add a description about what this is and what it is solving, if possible, with screenshots

components/molecule/field/src/index.scss
```diff
@@ -24,7 +24,6 @@ $atomHelpText-class: '.sui-AtomHelpText';
 
       #{$atomLabel-class} {
         align-items: center;
-        margin-left: $ml-molecule-field-label;
```
jcger, line 27: wouldn't this break somewhere?

ferransimon: Not it won't in order to both label and nodelabel to have the same behavior a div wrapper have been added to add this padding to both cases. This have been done in order to ensure retrocompatibility

jcger (approved):

PR #1062, opened 2020-03-31
feat(molecule/breadcrumb): remove breadcrumb padding

AitorRodriguez990:
## Description

To remove Molecule Breadcrumb padding because it's not its responsibility.


andresin87 (approved):

danivm (approved):

nguasch (changes requested):
This change requires a major version :(

PR #1063, opened 2020-03-31
feat(atom/card): using $bgc-atom-card-hover to keep styles backwards …

PablitoGS:
…compatible


andresin87 (approved):

andresin87 (changes requested):
I think you dont need the final "default" written value

nucliweb (approved):

beatrizip (approved):

PR #1066, opened 2020-04-03
feat(molecule/dropdownList): do not render list if is not visible

arnau-rius:
- Performance improvement: do not render list if is not visible.

### DIFF
--- a/components/molecule/dropdownList/src/index.js
+++ b/components/molecule/dropdownList/src/index.js
@@ -78,6 +78,8 @@ const MoleculeDropdownList = ({
     ev.stopPropagation()
   }
 
+  if (!visible) return null
+
   return (
     <ul
       ref={refDropdownList}
--- a/components/molecule/dropdownList/src/index.js
+++ b/components/molecule/dropdownList/src/index.js
@@ -112,6 +112,7 @@ MoleculeDropdownList.propTypes = {
 }
 
 MoleculeDropdownList.defaultProps = {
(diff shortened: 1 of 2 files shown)


davidmartin84 (approved):

carlosvillu (approved):

tomasmax (approved):

jcger (approved):

test/molecule/dropdownList/index.js
```diff
@@ -4,19 +4,90 @@
...
+        // Given
+        const props = {}
+        // When
+        const {container} = setup({
+          ...testDefaultProps,
+          ...props
+        })()
+        // The
+        expect(container).to.be.not.undefined
+        expect(container.children.length).to.be.equal(0)
```
andresin87: The default value is now undefined-- > (!undefined) = true , which prevents the renderby default. Is it giving any problem to your verticals?

andresin87 (commented):
I added some tests

components/molecule/dropdownList/src/index.js
```diff
@@ -78,6 +78,8 @@ const MoleculeDropdownList = ({
     ev.stopPropagation()
   }
 
+  if (!visible) return null
```
midudev, line 81: This new behaviour should be behind a prop as sometimes, for SEO reasons, some verticals could want to still render the content.

andresin87 (approved):

PR #1067, opened 2020-04-03
feat(molecule/textareaField): added prop updateInternalValue 

manjarcode:
added prop updateInternalValue in order to update the internal value of the AtomTextarea.

Also refactored the demo for MoleculeTextareaField, in order to add the new example of updateInternalValue.


demo/molecule/textareaField/demo/withUpdateInternalValue.js
```diff
@@ -0,0 +1,34 @@
+import React, {useState} from 'react'
+import MoleculeTextareaField from '../../../../components/molecule/textareaField/src'
+import Button from '../../../../components/atom/button/src'
```
jcger: only use relative path for the current component you want to show the demo of, this button should be the @s-ui/react-atom-button dependency

jcger (approved):

demo/molecule/textareaField/demo/withUpdateInternalValue.js
manjarcode: ok I will update it!

components/molecule/textareaField/src/index.js
```diff
@@ -29,6 +29,7 @@ const MoleculeTextareaField = ({
   autoHideHelpText = false,
   helpText,
   value = '',
+  updateInternalValue = '',
```
carlosvillu: I don't get why we need to props, `value` and `updateInternalValue`. If what we want is a controlled component we should expose only two props: `value` and `onChange` and if both of these props are not properly defined, I think that we can assume that the client want to use the uncontrolled version.

andresin87:
I think that actual `value` prop must be called `initialValue`, the required behavior of `value` must give the consumer the oportunity to modify its inner `value` on every component, and `updateInternalValue ` suggests me a `bolean` of `function` naming type.

components/molecule/textareaField/src/index.js
manjarcode: I agree with you @carlosvillu but I debated this with @jcger and we decided to create a new prop in order to ensure the retro-compatibility of the compontent.

javiauso (approved):

components/molecule/textareaField/src/index.js
arnau-rius: And what about creating a new (major) version with only `value` and `onChange` properly used instead of adding layers of props? @carlosvillu  @andresin87 @jcger

manjarcode: @arnau-rius the final version will only contain `value `and `onChange`, so not `updateInternalValue` nor `inicialValue`, I'm working on it right now

arnau-rius: good news!

carlosvillu (approved):

demo/molecule/textareaField/demo/withState.js
```diff
@@ -0,0 +1,17 @@
+import React from 'react'
+
+export default BaseComponent =>
```
andresin87:
```
const handleChange = setter =>(e, {value}) {
    setter(value)
}

export default BaseComponent => ({value, onChange, ...otherProps}) => {
  const [ownValue, setValue] = useState(value || '')
  return <BaseComponent {...otherProps} value={ownValue}, onChange={useCallback(handleChange(setValue))} />
}
```
or extend from `PureComponent`

andresin87 (changes requested):

andresin87 (approved):

PR #1070, opened 2020-04-06
feat(atom/input): add max and min prop to atom inpu

arnau-rius:
- Add max, min and step prop to atom/input component.

Note: When input `type` attribute is `number` there are two properties `max` and `min` that are different than the already defined props `maxLength` and `minLength`.

https://developer.mozilla.org/en-US/docs/Web/HTML/Element/input/number#max
https://developer.mozilla.org/en-US/docs/Web/HTML/Element/input/number#min

### DIFF
--- a/components/atom/input/src/Input/Component/index.js
+++ b/components/atom/input/src/Input/Component/index.js
@@ -69,6 +69,7 @@ const Input = ({
   minLength,
   min,
   max,
+  step,
   autoComplete,
   onChange = DEFAULT_PROPS.ON_CHANGE,
   onEnter = DEFAULT_PROPS.ON_ENTER,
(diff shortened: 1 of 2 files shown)


midudev (approved):

javiauso (approved):

igncalleja (approved):

andresin87:
could you please add the `step` attribute which is directly related to the `max`-`min` input[number] values?

arnau-rius:
> could you please add the `step` attribute which is directly related to the `max`-`min` input[number] values?

done

andresin87 (approved):

quinwacca (approved):

PR #1079, opened 2020-04-07
fix(molecule/select): onKeyDown logic, don't closeList before selection

tomasmax:
MoleculeSelect component doesn't work onKeyDown.
It automatically closes when clicking Down key a couple of times.
Fix this bug and add new logic to close dropdown when clicking outside.

**TASK**: https://jira.scmspain.com/browse/SUIC-221
**DEMO LINK**: https://sui-components-git-fix-molecule-select-onkeydown.schibstedspain.now.sh/workbench/molecule/select/demo

### DIFF
--- a/components/molecule/select/src/index.js
+++ b/components/molecule/select/src/index.js
@@ -71,18 +71,40 @@ const MoleculeSelect = props => {
     }
   )
 
-  useEffect(() => {
-    setOptionsData(getOptionData(children))
-  }, [children])
-
   const closeList = ev => {
     onToggle(ev, {isOpen: false})
     ev.preventDefault()
     ev.stopPropagation()
   }
 
+  const handleClick = ev => {
+    if (disabled) return
+    if (!refMoleculeSelect.current.contains(ev.target)) {
+      // outside click
+      closeList(ev)
+      setFocus(false)
+    } else {
+      refMoleculeSelect.current.focus()
+      setFocus(true)
+    }
+  }
+
+  useEffect(() => {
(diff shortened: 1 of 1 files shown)


nucliweb (approved):

adriadoz (approved):

igncalleja (approved):

javiauso (approved):

andresin87:
I think the PR event title must be ~~onKeyDown~~ onMouseDown

components/molecule/select/src/index.js
```diff
@@ -71,18 +71,40 @@ const MoleculeSelect = props => {
...
 
+  const handleClick = ev => {
+    if (disabled) return
+    if (!refMoleculeSelect.current.contains(ev.target)) {
+      // outside click
+      closeList(ev)
+      setFocus(false)
+    } else {
+      refMoleculeSelect.current.focus()
+      setFocus(true)
```
rmoralp: This `handleClick` is a callback for an event listener in document. So if MoleculeSelect does not contains the target you always focus MoleculeSelect. IMHO this is a bad practice because user possibly does not want to interact with this component always that him/her makes a mousedown or touchend.

tomasmax: If `MoleculeSelect` or its children don't contain the event target it loses the focus and closes option list. If `MoleculeSelect` contains the target it gets focused. This is a common way to check if you're clicking outside the element.

carlosvillu (approved):

andresin87 (approved):

PR #1106, opened 2020-04-29
feat(molecule/selectPopover): add hide actions prop

desko27:
We have a use case where interactive elements residing inside the content are **sometimes** enough to proceed, so bottom actions of `molecule/selectPopover` are not necessary.

Proposed `hideActions` prop just prevents them from rendering if passed.

### DIFF
--- a/demo/molecule/selectPopover/demo/index.js
+++ b/demo/molecule/selectPopover/demo/index.js
@@ -23,6 +23,7 @@ const Demo = () => {
   const [size, setSize] = useState(selectPopoverSizes.MEDIUM)
   const [placement, setPlacement] = useState(selectPopoverPlacements.RIGHT)
   const [hasEvents, setHasEvents] = useState(false)
+  const [actionsAreHidden, setActionsAreHidden] = useState(false)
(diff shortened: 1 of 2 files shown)


nucliweb (approved):

demo/molecule/selectPopover/demo/index.js
```diff
@@ -79,17 +80,32 @@ const Demo = () => {
...
+      <div>
+        <label>
+          <input
+            type="checkbox"
+            checked={actionsAreHidden}
+            onChange={ev => setActionsAreHidden(ev.target.checked)}
+          />
+          Actions are hidden
+        </label>
+      </div>
```
nucliweb, line 102: 👏🏼

emiliovillu (approved):

demo/molecule/selectPopover/demo/index.js
desko27: Thanks for the recommendation @nucliweb 😄

andresin87 (approved):

PR #1112, opened 2020-05-04
feat(molecule/select): remove important !important property

PablitoGS:


nguasch (approved):

components/molecule/select/src/index.scss
```diff
@@ -64,7 +64,7 @@ $bgc-atom-input-disabled: $c-gray-lightest !default;
       }
 
       svg {
-        fill: $c-select-list-arrow-fill !important;
+        fill: $c-select-list-arrow-fill;
```
jcger, line 67: couldn't this change break something in some vertical?

PablitoGS: We shouldn't use it, I think, ...in fact we can't overwrite it from IJ right now. We are going to wait for a review of this PR

andresin87: With the older behabior, the rule which could give us problems with this new version might appear with `!important` flag too (in the verticals) in order to be giving a resulting visible effect (not ignored).

PablitoGS: I'm going to close this PR, for now the fill value is working ok. We can discuss it in a SUI weekly in case we can do something to remove it. Thanks!

PR #1123, opened 2020-05-15
feat(behavior/device): add behavior deviceType component

andresin87:
## Behavior/Device
**TASK**: 

### Types of changes

ADD the Behavior device to in order get the device type from the `window.navigator.userAgent`

Some verticals need to diferenctiate behaviors of componnets depending on user device type usage


components/behavior/device/package.json
```diff
@@ -0,0 +1,18 @@
+{
+  "name": "sui-behavior-device",
```
midudev: `@s-ui/react-behavior-device`

midudev:
Is this package going to be used as well inside SUI-Components? I mean I would like to avoid having to load this component or its dependencies. In Fotocasa we're using an injected object from the server in order to avoid loading `ua-parser-js`on the client and I wouldn't like to be forced to load anyway because a subdependency is using it. :D

oriolpuig (approved):
👏

PR #1125, opened 2020-05-18
feat(atom/tag): add prop to avoid truncating large labels

a-ferrante:


components/atom/tag/src/index.js
```diff
@@ -41,14 +41,15 @@ const filterKeys = (obj, listOfProps) =>
...
+  const {href, icon, onClick, size, responsive, type, extendedLabel} = props
   const isActionable = onClick || href
   const classNames = cx(
     'sui-AtomTag',
     `sui-AtomTag-${size}`,
     type && `sui-AtomTag--${type}`,
     responsive && 'sui-AtomTag--responsive',
-    icon && 'sui-AtomTag-hasIcon'
+    icon && 'sui-AtomTag-hasIcon',
+    extendedLabel && 'sui-AtomTag--extended-label'
```
andresin87, line 52: My opinion is that `extendedLabel`naming is not enough declarative. maybe a naming which contains the ellipsis content declaration should be more explicit

PR #1133, opened 2020-05-21
feat(molecule/photoUploader): add preview to be returned on callback,…

quinwacca:
Add `previewUrl` as a string returned on callback, useful if you wanna preview images outside the `photoUploader` component.

Also, fix a callback bug produced when all images are deleted.
Before: when all images are deleted, no callback.
Now: when all images are deleted, an empty array is returned.

### DIFF
--- a/components/molecule/photoUploader/src/index.js
+++ b/components/molecule/photoUploader/src/index.js
@@ -110,11 +110,21 @@ const MoleculePhotoUploader = ({
   const _callbackPhotosUploaded = list => {
     if (list.length) {
       const blobsArray = list.reduce((array, currentFile) => {
-        const {blob, url, isNew, isModified, hasErrors, file} = currentFile
-        array.push({blob, url, isNew, isModified, hasErrors, file})
+        const {
+          blob,
+          url,
+          isNew,
+          isModified,
+          hasErrors,
+          file,
+          preview
+        } = currentFile
+        array.push({blob, url, isNew, isModified, hasErrors, file, preview})
         return [...array]
       }, [])
       callbackPhotosUploaded(blobsArray)
+    } else {
+      callbackPhotosUploaded([])
     }
   }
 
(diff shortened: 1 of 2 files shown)


nucliweb (approved):

davidmartin84 (approved):

components/molecule/photoUploader/src/index.js
```diff
@@ -108,14 +108,28 @@ const MoleculePhotoUploader = ({
...
   const _callbackPhotosUploaded = list => {
-    if (list.length) {
-      const blobsArray = list.reduce((array, currentFile) => {
-        const {blob, url, isNew, isModified, hasErrors, file} = currentFile
-        array.push({blob, url, isNew, isModified, hasErrors, file})
-        return [...array]
-      }, [])
-      callbackPhotosUploaded(blobsArray)
-    }
+    const blobsArray = list.reduce((array, currentFile) => {
```
rmoralp, line 111: 🍆 IMO array map will do the job and be easier to read

rmoralp (approved):

arnau-rius (approved):

components/molecule/photoUploader/src/index.js
```diff
@@ -108,14 +108,28 @@ const MoleculePhotoUploader = ({
...
+        preview
+      } = currentFile
+      array.push({
+        blob,
+        url,
+        isNew,
+        isModified,
+        hasErrors,
+        file,
+        previewUrl: preview
```
andresin87, line 128: I think is better to expose the original `preview` argument attribute. It can be so much confusing this inner silent naming changes

quinwacca: Yes, it is confusing. Is it a good idea to rename `preview` argument everywhere to `previewUrl` ? I believe it's a clearer name than only `preview`. What do you think?

PR #1140, opened 2020-05-27
Feat/remove class name prop

andresin87:
remove some className props not needed

### DIFF
--- a/components/atom/tag/src/index.js
+++ b/components/atom/tag/src/index.js
@@ -80,7 +80,6 @@ AtomTag.propTypes = {
   /**
    * Custom classes
    */
-  className: PropTypes.string,
   label: PropTypes.string.isRequired,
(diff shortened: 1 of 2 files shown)


components/atom/tag/src/index.js
```diff
@@ -80,7 +80,6 @@ AtomTag.propTypes = {
   /**
    * Custom classes
```
jcger, line 81: I'd say this comment should be removed too, looks like its referring the className prop

jcger:
I can't see no breaking changes defined in your commit messages so this would be released as a minor, you sure about this?

andresin87:
> I can't see no breaking changes defined in your commit messages so this would be released as a minor, you sure about this?

I think so. The prop is not used in the component for both cases. I would reword the commits to do not force a major

PR #1145, opened 2020-05-29
feat(atom/table): refactor component, added cellPadding & borderBottom props

nucliweb:
## In this Pull Request

- [x] Improve and fix README documentation
- [x] Added `cellPadding` prop to define the cell padding, now by default the cells don't have padding. 
- [x] Added `borderBottom` prop to define a default (hard coded) border bottom, now by default the cells don't have padding. 
- [x] Added props documentation to have the API documentation in SUI Studio
- [x] Added new sample in the demo

## Demo

table

### DIFF
--- a/components/atom/table/src/index.js
+++ b/components/atom/table/src/index.js
@@ -8,6 +8,7 @@ const CELL_TYPE = {
 }
 
 const CELL_PADDING = {
+  none: false,
   xs: 'xs',
   s: 's',
   m: 'm',
@@ -20,7 +21,7 @@ const AtomTable = ({
   body,
   foot,
   fullWidth,
-  cellPadding = false,
+  cellPadding = CELL_PADDING.none,
   borderBottom = false
 }) => {
   const hasHead = Boolean(head?.length)
@@ -29,7 +30,7 @@ const AtomTable = ({
   const tableClass = cx(`${baseClass}`, {
     [`${baseClass}--fullWidth`]: fullWidth
   })
-  const headerClass = cx(`${baseClass}-cell ${baseClass}-headerCell`, {
+  const headerClass = cx(`${baseClass}-cell`, `${baseClass}-headerCell`, {
     [`${baseClass}-cell--${cellPadding}`]: cellPadding
   })
 
@@ -39,7 +40,7 @@ const AtomTable = ({
         <thead>
           <tr>
             {head.map((element, index) => (
-              <th key={index} className={`${headerClass}`}>
+              <th key={index} className={headerClass}>
                 {element}
               </th>
             ))}
@@ -135,5 +136,5 @@ AtomTable.propTypes = {
   borderBottom: PropTypes.bool
 }
 
-export {CELL_TYPE as AtomTableCellTypes, CELL_PADDING as AtomTableCellPadding}
+export {CELL_TYPE as atomTableCellTypes, CELL_PADDING as atomTableCellPadding}
 export default AtomTable
--- a/components/atom/table/src/index.js
+++ b/components/atom/table/src/index.js
@@ -7,12 +7,30 @@ const CELL_TYPE = {
   data: 'td'
 }
 
-const AtomTable = ({head, body, foot, fullWidth}) => {
+const CELL_PADDING = {
+  xs: 'sizeXS',
+  s: 'sizeS',
+  m: 'sizeM',
+  l: 'sizeL',
+  xl: 'sizeL'
+}
+
+const AtomTable = ({
+  head,
+  body,
+  foot,
+  fullWidth,
+  cellPadding,
+  borderBottom = false
+}) => {
   const hasHead = Boolean(head?.length)
   const hasFoot = Boolean(foot?.length)
   const baseClass = 'react-AtomTable'
   const tableClass = cx(`${baseClass}`, {
     [`${baseClass}--fullWidth`]: fullWidth
+  })
+  const headerClass = cx(`${baseClass}-cell ${baseClass}-headerCell`, {
+    [`${baseClass}-cell--${cellPadding}`]: cellPadding
   })
 
   return (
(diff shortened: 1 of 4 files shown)


components/atom/table/src/index.js
```diff
@@ -78,19 +95,45 @@ const AtomTable = ({head, body, foot, fullWidth}) => {
...
+   */
+  fullWidth: PropTypes.bool,
+  /**
+   * Add a default border bootom to all cells
+   */
+  borderBottom: PropTypes.bool
 }
 
-export {CELL_TYPE as AtomTableTypes}
+export {CELL_TYPE as AtomTableTypes, CELL_PADDING as AtomTableCellPadding}
```
andresin87: export {CELL_TYPE as AtomTableCellType, CELL_PADDING as AtomTableCellPadding}

components/atom/table/src/index.js
```diff
@@ -43,7 +58,9 @@ const AtomTable = ({head, body, foot, fullWidth}) => {
                 colspan = 1
               } = cell
               const cellClassName = cx(`${baseClass}-cell`, {
-                [`${baseClass}-cell--noWrap`]: isNowrap
+                [`${baseClass}-cell--noWrap`]: isNowrap,
+                [`${baseClass}-cell--${cellPadding}`]: cellPadding,
```
andresin87: if cellpadding is undefined this will be added. Better removing this css class value if there is no defined cellpadding

andresin87 (changes requested):

components/atom/table/src/index.js
kikoruiz: Yep, the same in the classname above. In order to prevent it, if `cellPadding` prop is not required, it could be fine to set a default value, don't you think @nucliweb?

nucliweb: Yes, just added...
```
const AtomTable = ({
  head,
  body,
  foot,
  fullWidth,
  cellPadding = false,
  borderBottom = false
}) => {
```

components/atom/table/src/index.js
```diff
@@ -7,24 +7,39 @@ const CELL_TYPE = {
   data: 'td'
 }
 
-const AtomTable = ({head, body, foot, fullWidth}) => {
+const CELL_PADDING = {
+  xs: 'sizeXS',
```
kikoruiz: This naming convention `sizeXS` seems to be new, has there been an agreement about it? Perhaps it's enough to call it just `xs`.

components/atom/table/src/index.js
andresin87:
```js
cx(`${baseClass}-cell`,
  {
    [`${baseClass}-cell--noWrap`]: isNowrap,	                		 
    ... cellPadding && {[`${baseClass}-cell--${cellPadding}`]: cellPadding},
    ... borderBottom && {[`${baseClass}-cell--borderBottom`]: borderBottom},
  })
})
```

andresin87: Only add classes if props are defined. This will not add extra classes to the existant table usages. remove the default prop values added to the function... not needed

andresin87 (changes requested):

kikoruiz (approved):

components/atom/table/src/index.js
```diff
@@ -7,24 +7,39 @@ const CELL_TYPE = {
...
+  cellPadding = false,
+  borderBottom = false
+}) => {
   const hasHead = Boolean(head?.length)
   const hasFoot = Boolean(foot?.length)
   const baseClass = 'react-AtomTable'
   const tableClass = cx(`${baseClass}`, {
     [`${baseClass}--fullWidth`]: fullWidth
   })
+  const headerClass = cx(`${baseClass}-cell ${baseClass}-headerCell`, {
```
rmoralp:
```suggestion
  const headerClass = cx(`${baseClass}-cell`, `${baseClass}-headerCell`, {
```

components/atom/table/src/index.js
```diff
@@ -7,24 +7,39 @@ const CELL_TYPE = {
...
     <table className={tableClass}>
       {hasHead && (
         <thead>
           <tr>
             {head.map((element, index) => (
-              <th
-                key={index}
-                className={`${baseClass}-cell ${baseClass}-headerCell`}
-              >
+              <th key={index} className={`${headerClass}`}>
```
rmoralp:
```suggestion
              <th key={index} className={headerClass}>
```

components/atom/table/src/index.js
kikoruiz: If `cellPadding` is `false`, the classname won't be added, not needed extra check, could you test it @nucliweb?

components/atom/table/src/index.js
```diff
@@ -78,19 +95,45 @@ const AtomTable = ({head, body, foot, fullWidth}) => {
...
+   */
+  fullWidth: PropTypes.bool,
+  /**
+   * Add a default border bootom to all cells
+   */
+  borderBottom: PropTypes.bool
 }
 
-export {CELL_TYPE as AtomTableTypes}
+export {CELL_TYPE as AtomTableCellTypes, CELL_PADDING as AtomTableCellPadding}
```
rmoralp: Given this const are not a constructor, I think that should be in camelCase instead PascalCase:

```suggestion
export {CELL_TYPE as atomTableCellTypes, CELL_PADDING as atomTableCellPadding}
```

components/atom/table/src/index.js
```diff
@@ -7,24 +7,39 @@ const CELL_TYPE = {
...
+  l: 'l',
+  xl: 'l'
+}
+
+const AtomTable = ({
+  head,
+  body,
+  foot,
+  fullWidth,
+  cellPadding = false,
```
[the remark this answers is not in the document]
nucliweb: We want without padding by default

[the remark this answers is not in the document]
nucliweb: What do you think about this approach?

```js
const CELL_PADDING = {
  none: false,
  xs: 'xs',
  s: 's',
  m: 'm',
  l: 'l',
  xl: 'l'
}

const AtomTable = ({
  head,
  body,
  foot,
  fullWidth,
  cellPadding = CELL_PADDING.none,
  borderBottom = false
}) => { 
```

andresin87 (approved):

components/atom/table/src/index.js
andresin87: true, we where just talking about that.

rmoralp (approved):

PR #1147, opened 2020-06-02
feat(molecule/notification): add line height to notification text

PablitoGS:


nguasch (approved):

andresin87 (commented):
This is producing a 1px hop between this version and the original one. I think the token value might be `inherit`

PablitoGS:
@andresin87 done

andresin87 (approved):

PR #1149, opened 2020-06-04
Feature/add upload callback

tomcask:
Callback that executes every time that new image is added or modified.
Returns the new Blob and old URL if exists


components/molecule/photoUploader/src/PhotosPreview.js
```diff
@@ -79,11 +81,17 @@ const PhotosPreview = ({
...
         list[index].blob = blob
-        setFiles(list)
         list[index].isModified = true
+
+        list[index].url = await callbackUploadPhotoHandler(
+          blob,
+          callbackUploadPhoto,
+          list[index].url
+        )
+        setFiles(list)
```
andresin87, line 94: 👍 Preventing mutation

andresin87 (approved):

davidmartin84 (approved):

PR #1154, opened 2020-06-09
feat(atom/panel): added border-radius

ferransimon:


components/atom/panel/src/index.scss
```diff
@@ -1,5 +1,6 @@
 @import '~@s-ui/theme/lib/index';
 
+$bdrs-atom-panel: 0px !default;
```
midudev: Es raro que no se queje, pero diría que los valores que son 0 no van con unidades.

En lugar de `0px` debería ser `0`.

midudev (approved):

nucliweb (approved):

javiauso (approved):

components/atom/panel/src/index.scss
```diff
@@ -32,6 +33,7 @@ $vertical-alignments-atom-panel: ('top', 'center', 'bottom') !default;
   }
 
   &-color {
+    border-radius: $bdrs-atom-panel;
```
PablitoGS: Creo que no deberiamos asociar el border radius al panel color, en todo caso, en la vista mobile el panel-color no tiene border radius en IJ

components/atom/panel/src/index.scss
```diff
@@ -1,5 +1,6 @@
 @import '~@s-ui/theme/lib/index';
 
+$bdrs-atom-panel: 0 !default;
```
andresin87: `$bdrs-atom-panel: $bdrs-none !default`

andresin87 (changes requested):

components/atom/panel/src/index.scss
andresin87: @s-ui/theme/lib/settings/_box-style.scss

components/atom/panel/src/index.scss
turolopezsanabria: If we think mobile first, the `$bdrs` should be `none` by default, and apply a media query to be rounded only for Desktop.

andresin87: Both Components, ColorPanel && ImagePannel must include its baseClassName = `sui-atom-panel`

components/atom/panel/src/index.scss
ferransimon: I moved that to a component prop so now I left the default value to `$bdrs-l` so in the demo we can see a difference between a floating panel or a non-floating panel

components/atom/panel/src/index.scss
ferransimon: They do not have this baseClassName, there is a baseClassName for every single Panel (color, image) if we need to do a refactor on this maybe it gets out of the scope of this PR

andresin87 (approved):

PR #1157, opened 2020-06-15
Feat/atom tooltip always shown

oriolpuig:
### Types of changes

Everytime we want to show `AtomTooltip` once the reference component is loaded without a client event (touch, hover, etc) we need to mock/fire this client event from outside the component and then change `isOpen` state to `true`.

So, in Fotocasa, he have detected this logic might be a new feature for tooltip and move this code inside component.

This PR creates a new flag to mock firing a client event and don't break isOpen feature.

### DIFF
--- a/components/atom/tooltip/src/index.js
+++ b/components/atom/tooltip/src/index.js
@@ -46,9 +46,7 @@ class AtomTooltip extends Component {
       'reactstrap/lib/Tooltip'
     )
       .then(module => module.default)
-      .then(Tooltip => {
-        this.setState({Tooltip})
-      })
+      .then(Tooltip => this.setState({Tooltip}))
   }
 
   extendChildren() {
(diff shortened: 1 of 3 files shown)


javiauso (approved):

components/atom/tooltip/src/index.js
```diff
@@ -86,6 +86,12 @@ class AtomTooltip extends Component {
     target.oncontextmenu = this.handleContextMenu
     target.addEventListener('mouseover', this.disableTitle)
     target.addEventListener('mouseout', this.restoreTitle)
+
+    // We are emulating a client event to force display
+    if (target && this.props.alwaysShown) {
```
andresin87: I think this emulated behavior can change the default behavior of form elements focussing when  some vertical page loads. I ignore if there is some product setting by default some input ready to be completed when some page loads.

danivm (approved):

andresin87:
# UPDATE
this pr updates the behavior of isOpen to offer the possibility to take the control of its own with that prop even on the onMounting lifecycle (this was not possible already)

andresin87 (approved):

PR #1160, opened 2020-06-16
feat(molecule/photoUploader): fix to avoid rotation based on exif, if…

quinwacca:
Fix to avoid rotation based on EXIF, if the browser has this feature by default.

Browsers affected: Since Chrome 81, and Safari 13.1 (also Chrome Mobile and iOS). Firefox with some bugs, not fully support this.

Both WebKit (iOS) and Android (Chrome) have just recently changed the default behavior of the image-orientation CSS property. While it was none before, it is now from-image. This means: Before, they ignored the EXIF data of an image by default, while they are now using it to auto-correct the image. Which breaks our own rotation auto-correction.

Has been tested on:

- [x] Chrome (desktop)
- [x] Safari (desktop and iOS device)
- [x] Firefox (desktop)
- [x] Samsung Browser (Android device)
- [x] Chrome Mobile (Android and iOS device)

### DIFF
--- a/components/molecule/photoUploader/src/photoTools.js
+++ b/components/molecule/photoUploader/src/photoTools.js
@@ -19,15 +19,24 @@ export function formatToBase64({
       reader.onload = () => {
         getExifOrientation(file)
           .then(getExifOrientationResult => {
-            switch (getExifOrientationResult) {
-              case (3, 4):
-                options.rotation = 180
-                break
-              case (5, 6):
-                options.rotation = 90
-                break
-              case (7, 8):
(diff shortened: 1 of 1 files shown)


davidmartin84 (approved):

nucliweb (approved):

jordevo (approved):
good one, mate! 👏

arnau-rius (approved):

AgonisticKatai (approved):

andresin87 (approved):

danivm (approved):

PR #1165, opened 2020-06-22
feat(Root): fix padding in demo table

xaviermariaconejo:
1) Fix padding in demo tables

2) We have noticed that some components are using the user agent styles, which is a bug. But since we should have deployed that studio visually correct, we have removed the reset styles. There still is the problem that the components that are using the user agent styles have bug, and we should try to all the different frontals use the same reset styles.

### DIFF
--- a/demo/atom/actionButton/playground
+++ b/demo/atom/actionButton/playground
@@ -190,6 +190,7 @@ function ButtonsTable({children, style}) {
   return (
     <table
       width="auto"
+      className="sui-StudioTable"
       cellPadding="8"
       cellSpacing="0"
       style={{background: '#fff', padding: 15}}
@@ -217,6 +218,7 @@ function ButtonsColorsTable({children, style}) {
   return (
     <table
       width="auto"
+      className="sui-StudioTable"
       cellPadding="8"
       cellSpacing="0"
       style={{background: '#fff', padding: 15}}
--- a/demo/atom/button/playground
+++ b/demo/atom/button/playground
@@ -484,6 +484,7 @@ function ButtonsTable({children, style}) {
   return (
     <table
       width="auto"
+      className="sui-StudioTable"
       cellPadding="8"
       cellSpacing="0"
       style={{background: '#fff', padding: 15}}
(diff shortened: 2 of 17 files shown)


adriadoz (approved):

demo/package.json
```diff
@@ -8,5 +8,8 @@
       "packagesFolder": ".",
       "deepLevel": 3
     }
+  },
+  "dependencies": {
+    "undefined": "^0.1.0"
```
oriolpuig: Could you remove that dependency? As far as I know, it's a known issue at sui-mono.

demo/package.json
```diff
@@ -8,5 +8,8 @@
       "packagesFolder": ".",
       "deepLevel": 3
     }
+  },
+  "dependencies": {
```
midudev: Remove this.

package.json
```diff
@@ -30,11 +31,11 @@
     "@s-ui/mono": "1",
     "@s-ui/precommit": "2",
     "@s-ui/studio": "8",
-    "fs-extra": "8.1.0",
-    "globby": "10.0.1",
-    "husky": "0.13.4",
-    "validate-commit-msg": "2.12.2",
-    "walker": "1.0.7"
+    "fs-extra": "8",
```
oriolpuig: Why we have updated the version pointing the major one? Normally we do that only on our private packages or SUI packages.

oriolpuig (approved):

package.json
midudev: We don't rely on major versions for external dependencies so, just use the fixed versions.

demo/package.json
xaviermariaconejo: Yes, sorry

package.json
xaviermariaconejo: Removed!

xaviermariaconejo: Fixed!

demo/package.json
xaviermariaconejo: done

midudev (approved):

PR #1180, opened 2020-07-14
Feature/atom table action

ivanmlaborda:
- Pass onClick callback to atom/table returning the index row pressed.
- Applied hover/action styles to rows when atom/table is actionable (receive an onClick callback).

### DIFF
--- a/components/atom/table/src/index.js
+++ b/components/atom/table/src/index.js
@@ -21,10 +21,12 @@ const AtomTable = ({
   foot,
   fullWidth,
   cellPadding,
-  borderBottom
+  borderBottom,
+  onClick
 }) => {
   const hasHead = Boolean(head?.length)
   const hasFoot = Boolean(foot?.length)
+  const isActionable = Boolean(onClick)
   const baseClass = 'react-AtomTable'
   const tableClass = cx(`${baseClass}`, {
     [`${baseClass}--fullWidth`]: Boolean(fullWidth)
@@ -32,6 +34,11 @@ const AtomTable = ({
   const headerClass = cx(`${baseClass}-cell`, `${baseClass}-headerCell`, {
     [`${baseClass}-cell--${cellPadding}`]: Boolean(cellPadding)
   })
+  const rowClass = cx(`${baseClass}-row`, {
+    [`${baseClass}-row--actionable`]: isActionable
+  })
+
+  const handleOnClick = index => onClick(index)
 
   return (
     <table className={tableClass}>
@@ -49,7 +56,11 @@ const AtomTable = ({
 
       <tbody>
         {body.map((row, index) => (
-          <tr key={index}>
+          <tr
+            key={index}
+            className={rowClass}
+            {...(isActionable && {onClick: () => handleOnClick(index)})}
+          >
             {row.map((cell, index) => {
               const {
                 type: Element = CELL_TYPE.data,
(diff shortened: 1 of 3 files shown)


components/atom/table/src/index.js
```diff
@@ -21,17 +21,24 @@ const AtomTable = ({
   foot,
   fullWidth,
   cellPadding,
-  borderBottom
+  borderBottom,
+  onClick
```
giodelabarrera: Maybe we could call the prop `onRowClick`. So we can give more semantics. 

```suggestion
  onRowClick
```

ivanmlaborda: It's a good point. Thank you!

tomcask (approved):

components/atom/table/src/index.js
giodelabarrera: If you initialize with a [noop](https://www.damiannicholson.com/posts/2018/03/06/react-noop-anti-pattern/) function it would also always be actionable and you could avoid internal checks.

```suggestion
  onClick = () => {}
```

giodelabarrera (approved):

andresin87 (commented):
you dnt need the isActionable bolean, you can directly check if the prop is already declared.

```
  const rowClass = cx(`${baseClass}-row`, {
    [`${baseClass}-row--actionable`]: onClick
  })
...
  {...(onClick && {onClick: () => handleOnClick(index)})}
```

components/atom/table/src/index.js
andresin87: this is not retrocompatible due to styling changes

ivanmlaborda: if we set () => {} as a default value for onRowClick prop, the evaluation always will be true and we cannot know when the component is receiving the onRowClick prop or not (for example, on this case when has the callback, the rows have a hover effect)

giodelabarrera: ok, noop then it doesn't work if you are interested in evaluating if you receive the value 👍🏼

ivanmlaborda:
> you dnt need the isActionable bolean, you can directly check if the prop is already declared.
> 
> ```
>   const rowClass = cx(`${baseClass}-row`, {
>     [`${baseClass}-row--actionable`]: onClick
>   })
> ...
>   {...(onClick && {onClick: () => handleOnClick(index)})}
> ```

Well, I use it for semantics, but you are right, maybe is not necessary

devictoribero (approved):

andresin87:
> > you dnt need the isActionable bolean, you can directly check if the prop is already declared.
> > ```
> >   const rowClass = cx(`${baseClass}-row`, {
> >     [`${baseClass}-row--actionable`]: onClick
> >   })
> > ...
> >   {...(onClick && {onClick: () => handleOnClick(index)})}
> > ```
> 
> Well, I use it for semantics, but you are right, maybe is not necessary

well, so better giving a more concrete naming like `isRowActionable`?

components/atom/table/src/index.js
ivanmlaborda: > this is not retrocompatible due to styling changes

the hover effect is only applied when onRowClick prop is received, and the prop is new too.

andresin87: Of course, i mean. Its not retrocompatible when adding noop as a default value of the mentioned new prop

ivanmlaborda: Yes of course, if we use the noop function, the styles will be applied in all cases ;)

andresin87 (approved):

PR #1190, opened 2020-07-23
Feature: dropdown option with line wrapping

jordevo:
- added prop `textWrap` which allows for different multi-line wrapping for lengthy options
- exported `textWrap` style options as `MoleculeDropdownOptionTextWrapStyles`: `NO_WRAP` (default), `TWO_LINES`, `THREE_LINES`, `LINE_WRAP`
- `line-height` turned into `min-line-height` for dropdown options
- added sass variables so padding can be controlled more granularly

This will now be possible:
image

### DIFF
--- a/components/molecule/dropdownOption/src/_settings.scss
+++ b/components/molecule/dropdownOption/src/_settings.scss
@@ -1,3 +1,4 @@
 $c-dropdown-option-disabled: color-variation($c-gray, 3) !default;
 $mih-dropdown-option: 40px !default;
+$h-dropdown-option: $mih-dropdown-option !default; // deprecated
 $bgc-dropdown-option-hover: $c-gray-lightest !default;
--- a/components/molecule/dropdownOption/src/_settings.scss
+++ b/components/molecule/dropdownOption/src/_settings.scss
@@ -1,3 +1,3 @@
 $c-dropdown-option-disabled: color-variation($c-gray, 3) !default;
-$h-dropdown-option: 40px !default;
+$mih-dropdown-option: 40px !default;
 $bgc-dropdown-option-hover: $c-gray-lightest !default;
--- a/components/molecule/dropdownOption/src/index.js
+++ b/components/molecule/dropdownOption/src/index.js
@@ -51,7 +51,8 @@ const MoleculeDropdownOption = ({
       `${CLASS_TEXT}--${MODIFIER_THREE_LINES}`,
     textWrap === TEXT_WRAP_STYLES.LINE_WRAP &&
       `${CLASS_TEXT}--${MODIFIER_LINE_WRAP}`,
-    textWrap === TEXT_WRAP_STYLES.NO_WRAP &&
+    ((!withTwoLinesText && !textWrap) ||
+      textWrap === TEXT_WRAP_STYLES.NO_WRAP) &&
       `${CLASS_TEXT}--${MODIFIER_NO_WRAP}`
   ])
 
--- a/components/molecule/dropdownOption/src/index.js
+++ b/components/molecule/dropdownOption/src/index.js
@@ -10,7 +10,7 @@ const BASE_CLASS = 'sui-MoleculeDropdownOption'
 const CLASS_CHECKBOX = `${BASE_CLASS}-checkbox`
 const MODIFIER_TWO_LINES = `twoLines`
 const MODIFIER_THREE_LINES = `threeLines`
-const MODIFIER_ELLIPSIS = 'ellipsis'
+const MODIFIER_NO_WRAP = 'noWrap'
 const MODIFIER_LINE_WRAP = 'lineWrap'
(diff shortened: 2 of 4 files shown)


adriadoz (approved):

davidmartin84 (approved):

andresin87:
How does this prop works combined with `withTwoLinesText`¿?

jordevo:
good question @andresin87 , in my opinion in case both props are set to true, we should only take `withLineWrap` into account.

devictoribero:
what if the prop is `textWrap` and we have many values? `none`, `partial` (2 lines), `total`?

This way we would stop dealing with booleans

AgonisticKatai (approved):

jordevo:
@devictoribero I wanted to make sure the change was backwards compatible with an already in use prop... but sounds like an interesting suggestion. We should make sure and document that `withTwoLines` would be deprecated in favour of this new prop. What do you think, @andresin87 ?

andresin87:
I think this is the right way cuz motos needs 3 lines and that prop is too much strict ( not as much abstract as needed). I will talk with @jcger in order to find a solution. It is used in ij.

jordevo:
@andresin87 since we're in need of this feature **ASAP** and we've got capacity to implement it right now, would you mind giving us feedback about what's the best course of action (prop with enum values, I assume) so that we can go on with it?

andresin87:
ok, let's do that:

- `textWrap` have higher priority over `withTwoLinesText`,
- it has the values @devictoribero says
- and its values are enumerated and also provided by the component

components/molecule/dropdownOption/src/index.scss
```diff
@@ -32,19 +39,30 @@ $p-molecule-dropdown-option: $p-l !default;
...
     &--ellipsis {
       text-overflow: ellipsis;
       white-space: nowrap;
     }
 
+    &--lineWrap {
+      line-height: $ln-molecule-dropdown-option-line-wrap-text;
+    }
+
     &--twoLines {
```
andresin87:
```
&--twoLines, &--threeLines {
	-webkit-box-orient: vertical;
	display: -webkit-box;
	line-height: $ln-molecule-dropdown-option-lined-text;
}
&--twoLines {
	-webkit-line-clamp: 2;
}
&--threeLines {
	-webkit-line-clamp: 3;
}
```
and line-height must be `lh-`and not `ln-`
[emmet](https://docs.emmet.io/cheat-sheet/)

components/molecule/dropdownOption/src/index.js
```diff
@@ -35,9 +45,14 @@ const MoleculeDropdownOption = ({
 
   const innerClassName = cx([
     CLASS_TEXT,
```
andresin87, line 47: I'm not sure this is giving the prior usage of `text-warp` over `withTwoLines` props

components/molecule/dropdownOption/src/index.scss
```diff
@@ -8,14 +8,20 @@ $bg-molecule-dropdown-option-selected: none !default;
 $fw-molecule-dropdown-option-selected-text: $fw-bold !default;
 $fz-molecule-dropdown-option-text: inherit !default;
 $fw-molecule-dropdown-option--highlight-mark: normal !default;
-$ln-molecule-dropdown-option-two-lines-text: $lh-m !default;
+$lh-molecule-dropdown-option-line-wrap-text: $lh-m !default;
```
andresin87, line 11: it might be retrocompatible, soo it should still be named
`$ln-molecule-dropdown-option-two-lines-text` but including this is deprecated and actual value is
`$lh-molecule-dropdown-option-line-wrap-text`and link the actual value

andresin87 (approved):

devictoribero (approved):

components/molecule/dropdownOption/src/index.scss
```diff
@@ -8,14 +8,21 @@ $bg-molecule-dropdown-option-selected: none !default;
...
+$ln-molecule-dropdown-option-two-lines-text: $lh-molecule-dropdown-option-line-wrap-text !default; // deprecated
```
andresin87, line 12: 👍

components/molecule/dropdownOption/src/_settings.scss
```diff
@@ -1,3 +1,3 @@
 $c-dropdown-option-disabled: color-variation($c-gray, 3) !default;
-$h-dropdown-option: 40px !default;
+$mih-dropdown-option: 40px !default;
```
andresin87, line 2: ok, but the same as b4... not retrocompatible

andresin87 (changes requested):
retrocompatibility

jordevo:
@andresin87 adding backwards compatibility

andresin87 (approved):

PR #1199, opened 2020-08-05
Feat/autosuggest add input mode and type props

arnau-rius:
- export `inputTypes` from `atom/input`
- add new props: `inputMode` [more info](https://developer.mozilla.org/en-US/docs/Web/HTML/Global_attributes/inputmode) and `type` in `molecule/autosuggest` component (already defined in `atom/input` component).

### DIFF
--- a/components/atom/input/src/index.js
+++ b/components/atom/input/src/index.js
@@ -143,4 +143,4 @@ AtomInput.propTypes = {
 AtomInput.displayName = 'AtomInput'
 
 export default AtomInput
-export {inputSizes, inputStates}
+export {inputSizes, inputStates, TYPES as inputTypes}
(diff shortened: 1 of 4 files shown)


Sabri1209 (approved):

devictoribero:
@arnau-rius I don't see the difference between `type` and `inputMode`. Could you explain it with your words so I could understand it, please 🙏 ? I thought it was the same.

In addition, is there any time that we will want to pass different values for each of them?

igncalleja (approved):

desko27 (approved):

andresin87:
@devictoribero check the proptypes. Input mode is for keyboard viewmode in mobile devices. Type is exactly what input type defines in native tag

andresin87 (approved):

PR #1200, opened 2020-08-06
feat(molecule/accordion): add default sass variable on molecule accor…

ivanmlaborda:
- add sass variable to molecule accordion to be able to display rounded borders on tabs.
Default is the same behavior as actual, no rounded borders

### DIFF
--- a/components/molecule/accordion/src/index.scss
+++ b/components/molecule/accordion/src/index.scss
@@ -13,6 +13,7 @@
   padding: 0;
   position: relative;
   width: 100%;
+  border-radius: $bdrs-accordion-tab;
 
   &:last-child {
     border-bottom: $bd-accordion-tab-container;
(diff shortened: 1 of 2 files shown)


devictoribero (approved):

rmoralp (approved):

desko27 (approved):

nucliweb (approved):

components/molecule/accordion/src/index.scss
```diff
@@ -99,6 +100,7 @@
     background-color: $bgc-accordion-tab-container;
     position: relative;
     width: 100%;
+    border-radius: $bdrs-accordion-tab;
```
jordevo, line 103: 🍆 if it's no different than the border radius for the other children, do you really need this property here (using the same `$bdrs-accordion-tab` variable)?

jordevo (approved):

components/molecule/accordion/src/index.scss
ivanmlaborda: Yes, it needs to be applied to both selectors.

AgonisticKatai (approved):

quinwacca (approved):

andresin87 (approved):

PR #1212, opened 2020-09-02
feat(atom/table): add new tokens

Sabri1209:
Add new tokens and properties to style the component:

*Background-color to header cell
*Background-color to alternate rows
*Border-radius to table

**Why is this PR needed?**
At this moment, the table won't allow Pijama Stripes, and custom background for the header.
Both features are needed for a MA user story, and we agree this is good for SUI to have as options (as opposed to have this implemented at brand level only)

**This is how the desired design looks like:**
table options

### DIFF
--- a/components/atom/table/src/index.scss
+++ b/components/atom/table/src/index.scss
@@ -25,7 +25,7 @@ $bgc-atom-table-row-actionable--active: $c-gray-light-3 !default;
   }
 
   &-row {
-    &:nth-child(2n + 2) {
+    &:nth-child(2n) {
       background-color: $bgc-atom-table-row;
     }
     &--actionable {
--- a/components/atom/table/src/index.scss
+++ b/components/atom/table/src/index.scss
@@ -1,7 +1,7 @@
 @import '~@s-ui/theme/lib/index';
 
 $bgc-atom-table-header-cell: transparent !default;
-$bgc-atom-table-row: $c-gray-light-3 !default;
+$bgc-atom-table-row: $c-gray-light-4 !default;
 $bdrs-atom-table: $bdrs-none !default;
 $bgc-atom-table-row-actionable--hover: $c-gray-light-4 !default;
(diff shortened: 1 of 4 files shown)


components/molecule/modal/src/index.js
```diff
@@ -48,7 +48,7 @@ const MoleculeModal = ({
 
   const closeModal = useCallback(
     ev => {
-      ev?.stopPropagation()
+      ev.stopPropagation()
```
jcger: why are you removing it? anyhow, as you did not commit this under the molecule/modal scope no release would be created

Sabri1209: because I was getting this error
Captura de pantalla 2020-09-02 a las 15 54 43

adriadoz (approved):

components/atom/table/src/index.scss
```diff
@@ -1,24 +1,33 @@
...
   }
 
   &-headerCell {
+    background-color: $bgc-atom-table-header-cell;
     font-weight: $fw-regular;
     text-align: left;
   }
 
   &-row {
+    &:nth-child(2n + 2) {
```
andresin87: 2n + 2 is the same as 2n ommiting the 0 row

andresin87: i dk what does this rule purpose but i think its too much specific for becoming generalist

turolopezsanabria: hmm, I think it's ok, it does what this PR is intending to do

turolopezsanabria (approved):

andresin87 (approved):

andresin87:
let (2n + 2) rule become again

turolopezsanabria:
Open question, more related to the direction we want to take
Do you guys see the PijamaStripes as a potential `Prop`?

This way, a theme can have both options if needed, and not being forced to one for ever

andresin87 (approved):

javiauso (approved):

AgonisticKatai (approved):

PR #1222, opened 2020-09-15
Feat/tag actionable

giodelabarrera:
# Autosuggest actionable from tags

Add the focus of the field from anywhere via `onClick` event

- Adds consistency by leaving a single input source with variable `props` and for the rest with `restProps` for `MoleculeAutosuggest`, `MoleculeAutosuggestField`.
- Move `MoleculeAutosuggestField` of class component to functional component
- Apply handler prefix rule for event handlers
- Fix warnings in demo

Sep-15-2020 14-49-39

### DIFF
--- a/components/molecule/autosuggest/src/index.js
+++ b/components/molecule/autosuggest/src/index.js
@@ -34,23 +34,24 @@ const getIsTypeableKey = key => {
   return key.length === 1 || keysEdit.includes(key)
 }
 
-const MoleculeAutosuggest = ({multiselection, ...props}) => {
-  const {
-    refMoleculeAutosuggest: refMoleculeAutosuggestFromProps,
-    children,
-    onToggle,
-    onChange,
-    onBlur,
-    onEnter,
-    onFocus,
-    isOpen,
-    keysCloseList,
-    keysSelection,
-    disabled,
-    errorState,
-    state
-  } = props
-
+const MoleculeAutosuggest = ({
+  multiselection,
+  refMoleculeAutosuggest: refMoleculeAutosuggestFromProps,
+  children,
+  onToggle,
+  onChange,
+  onBlur,
+  onEnter,
+  onFocus,
+  isOpen,
+  keysCloseList,
+  keysSelection,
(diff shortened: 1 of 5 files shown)


components/molecule/autosuggestField/src/index.js
```diff
@@ -17,57 +17,53 @@ const getState = ({successText, errorState, alertText}) => {
...
+  label,
+  onChange,
+  successText,
+  useContrastLabel,
+  children,
+  ...restProps
+}) {
+  const refAutosuggest = useRef()
+
+  const handleClick = () => {
```
andresin87, line 35: better memoized

```
const handleClick = useCallback( () => {
  const {current: domAutosuggest} = refAutosuggest
  if (domAutosuggest) domAutosuggest.focus()
}, [refAutosuggest])
```

giodelabarrera: I understand that you say it for performance, but I think that in this case it is not necessary.

The first is that the optimization can be more expensive than the benefit, for this case I think that it is more advantageous to execute the same code in each render, than to delegate the cost to React to check the change of the dependencies.

And the other reason is that in this case we are using a variable that has been created with `useRef` and does not change between different renders.

https://kentcdodds.com/blog/usememo-and-usecallback

andresin87 (approved):

danivm (approved):

adriadoz (approved):

components/molecule/autosuggest/src/index.js
```diff
@@ -34,23 +34,24 @@ const getIsTypeableKey = key => {
...
-    isOpen,
-    keysCloseList,
-    keysSelection,
-    disabled,
-    errorState,
-    state
-  } = props
-
+const MoleculeAutosuggest = ({
+  multiselection,
```
oriolpuig: I would suggest to order props to make code more readable. Here and during this PR, we could do it. It's a suggestion, feel free!

oriolpuig (approved):

PR #1225, opened 2020-09-17
Feature/molecule accordion default opened tab

ivanmlaborda:
Add defaultOpenedTabs prop to sui molecule Accordion component.

imatge


AgonisticKatai (approved):

adriadoz (approved):

andresin87:
It might also cover the muilti-opened-tab (Array on indexes)

ivanmlaborda:
> It might also cover the muilti-opened-tab (Array on indexes)

Maybe open more than one tab by default (but not all tabs) is a strange edge case. Isn't it?

And it will create problems with autoclose. If two tabs are opened by default and the user open another one, the two default tabs will be closed, and it will seem strange for the user.

andresin87:
> > It might also cover the muilti-opened-tab (Array on indexes)
> 
> Maybe open more than one tab by default (but not all tabs) is a strange edge case. Isn't it?
> 
> And it will create problems with autoclose. If two tabs are opened by default and the user open another one, the two default tabs will be closed, and it will seem strange for the user.

It needs to attend Accordion with no autoclose cases

seremejvaz (approved):

ivanmlaborda:
> > > It might also cover the muilti-opened-tab (Array on indexes)
> > 
> > 
> > Maybe open more than one tab by default (but not all tabs) is a strange edge case. Isn't it?
> > And it will create problems with autoclose. If two tabs are opened by default and the user open another one, the two default tabs will be closed, and it will seem strange for the user.
> 
> It needs to attend Accordion with no autoclose cases

Ok, I have refactored to support multiple default opened tabs. Now defaultOpenedTabs prop has to be an array of indexes (numbers).

andresin87 (approved):

PR #1226, opened 2020-09-18
fix(atom/upload): add css vars and remove atomIcon ussage

a-ferrante:
Add css vars to allow styling from the themes

Remove the usage of AtomIcon so we can determinate size and color of the icon from outside instead of always using a large one


components/atom/upload/package.json
```diff
@@ -11,7 +11,6 @@
   "dependencies": {
     "@s-ui/component-dependencies": "1",
     "@s-ui/react-atom-button": "1",
-    "@s-ui/react-atom-icon": "1",
```
jcger, line 14: this should be a major release but I can't find the proper commit message

a-ferrante: done!

jcger (approved):

andresin87 (approved):

PR #1234, opened 2020-09-30
fix(atom/button): use prop rel in AtomButton when it is a link

rmoralp:
`rel` attribute can have different values as you can see in: https://developer.mozilla.org/en-US/docs/Web/HTML/Attributes/rel#:~:text=The%20rel%20attribute%20defines%20the,which%20the%20attribute%20is%20found.

We need to enable the possibility of sent the value we want in props and use it. `nofollow`, `noopener`, `noreferrer` are common values`rel` could countains

### DIFF
--- a/components/atom/button/src/Button.js
+++ b/components/atom/button/src/Button.js
@@ -14,13 +14,12 @@ const Button = ({
 }) => {
   if (isSubmit) attrs.type = 'submit'
   if (isButton) attrs.type = 'button'
+
+  const defaultRel = target === '_blank' ? 'noopener' : undefined
+  const rel = attrs.rel || defaultRel
+
   return link ? (
-    <Link
-      {...attrs}
-      href={href}
-      target={target}
-      rel={target === '_blank' ? 'noopener' : undefined}
-    >
+    <Link {...attrs} href={href} target={target} rel={rel}>
       {children}
     </Link>
   ) : (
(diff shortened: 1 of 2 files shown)


adriadoz (approved):

components/atom/button/src/Button.js
```diff
@@ -14,13 +14,12 @@ const Button = ({
 }) => {
   if (isSubmit) attrs.type = 'submit'
   if (isButton) attrs.type = 'button'
+
+  const defaultRel = target === '_blank' ? 'noopener' : undefined
+  const rel = attrs.rel || defaultRel
```
jordevo, line 19: wouldn't this achieve the same? 

```suggestion
  const {rel = defaultRel} = attrs
```

tomcask: I like the more traditional way

image

vs

image

Your suggestion does not provide more readability and generates more lines of code.

tomcask (approved):

javiauso (approved):

components/atom/button/src/Button.js
jordevo: gotcha... I'm just so used to assigning while destructuring...

jordevo (approved):

midudev (approved):

midudev:
I've merged master in order to get tests to pass. ✅

PR #1240, opened 2020-10-07
feat(atom/upload): remove atom icon and add sass vars

a-ferrante:
BREAKING CHANGE:
Remove atom icon dep

### Types of changes

Add sass variables to be able to style the component even more from vertical themes.
Also since the svg repos already wrap SVG in AtomIcon we can remove the dependency from this component and use the icon prop instead

### DIFF
--- a/components/atom/upload/src/index.js
+++ b/components/atom/upload/src/index.js
@@ -1,10 +1,6 @@
 import React, {useState, useEffect, lazy, Suspense} from 'react'
 import PropTypes from 'prop-types'
 import cx from 'classnames'
-import AtomIcon, {
-  ATOM_ICON_COLORS,
-  ATOM_ICON_SIZES
-} from '@s-ui/react-atom-icon'
 const Dropzone = lazy(() => import('react-dropzone'))
 
 const STATUSES = {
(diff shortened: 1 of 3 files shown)


components/atom/upload/src/index.js
```diff
@@ -151,9 +140,7 @@ AtomUpload.propTypes = {
   accept: PropTypes.oneOfType([
     PropTypes.string,
     PropTypes.arrayOf(PropTypes.string)
-  ]),
```
andresin87, line 154: loosing this feature

components/atom/upload/src/index.js
```diff
@@ -151,9 +140,7 @@ AtomUpload.propTypes = {
...
-  /** Size of icon */
-  iconSize: PropTypes.oneOf(Object.values(ATOM_ICON_SIZES))
```
andresin87, line 156: loosing this feature

a-ferrante: its not that we are loosing it, the icon component that you pass down to the uploader can have any size, we are just decoupling the size from this component and this can be decided by the parent component.

jcger (approved):

ferransimon (approved):

andresin87 (approved):

PR #1251, opened 2020-10-08
fix(molecule/tabs): prevent render content element if there is empty

rmoralp:
### Types of changes
If we do not set a content for a tab, this component renders an empy div. 
This is not a desired behavior

We are using MoleculeTabs for tab content, but this content is in other position, not jus following tabs. So, we are using tabs without content and manage state in our component.

### Screenshots - Animations
Tabs toggle Vehicle list with other kind of content, but not the form and the header.
image

### DIFF
--- a/components/molecule/tabs/src/index.js
+++ b/components/molecule/tabs/src/index.js
@@ -46,7 +46,9 @@ const MoleculeTabs = ({variant, type, children, onChange}) => {
   return (
     <div className={className}>
       <ul className={CLASS_SCROLLER}>{extendedChildren}</ul>
-      <div className={CLASS_CONTENT}>{activeTabContent}</div>
+      {activeTabContent ? (
+        <div className={CLASS_CONTENT}>{activeTabContent}</div>
+      ) : null}
     </div>
   )
 }
--- a/demo/package.json
+++ b/demo/package.json
@@ -8,8 +8,5 @@
       "packagesFolder": ".",
       "deepLevel": 3
     }
-  },
-  "dependencies": {
-    "undefined": "^0.1.0"
   }
 }
--- a/demo/package.json
+++ b/demo/package.json
@@ -8,5 +8,8 @@
       "packagesFolder": ".",
       "deepLevel": 3
     }
+  },
+  "dependencies": {
+    "undefined": "^0.1.0"
   }
 }


demo/package.json
```diff
@@ -8,5 +8,8 @@
       "packagesFolder": ".",
       "deepLevel": 3
     }
+  },
```
klaufel, line 11: 👀

demo/package.json
```diff
@@ -8,5 +8,8 @@
...
+  "dependencies": {
+    "undefined": "^0.1.0"
```
klaufel: 👀

components/molecule/tabs/src/index.js
```diff
@@ -46,7 +46,9 @@ const MoleculeTabs = ({variant, type, children, onChange}) => {
   return (
     <div className={className}>
       <ul className={CLASS_SCROLLER}>{extendedChildren}</ul>
-      <div className={CLASS_CONTENT}>{activeTabContent}</div>
+      {activeTabContent ? (
+        <div className={CLASS_CONTENT}>{activeTabContent}</div>
+      ) : null}
```
klaufel, line 51:
```suggestion
      {activeTabContent && (
        <div className={CLASS_CONTENT}>{activeTabContent}</div>
      )}
```

no need to return a value, could you remove the ternary?

demo/package.json
rmoralp: Fixed!

components/molecule/tabs/src/index.js
rmoralp: This ternary exactly do that. Use ternaries could cause some undesirable problems because if first operation is not "true" React can render the variable value (for example: 0 or false)

I recommend you an interesting article about that https://kentcdodds.com/blog/use-ternaries-rather-than-and-and-in-jsx

devictoribero: Agree with @rmoralp

devictoribero (approved):

AgonisticKatai (approved):

components/molecule/tabs/src/index.js
klaufel: Thanks @rmoralp!! Excellent article!!  😍 😍

I usually use logical operators a lot and to avoid rendering "0 " them I do the operation with as `!!foo.length`

The ternaries as a `null` return I use them  in the first level of rendering of the component.

Thanks!

klaufel (approved):

jordevo (approved):

andresin87:
me suena que esta asi por temas de SEO
(@midudev)

andresin87 (approved):

andresin87 (approved):

andresin87 (approved):

andresin87 (approved):

andresin87 (approved):

PR #1253, opened 2020-10-09
feat(atom/button): add rounded button support

andresz1:
## atom/button
**TASK**: None

### Types of changes

This PR adds `isRounded` prop to `AtomButton`, if enabled rounded borders variation will be use.

### Screenshots - Animations
Screenshot 2020-10-09 at 14 31 50


turolopezsanabria:
Do you think we could avoid Boolean for this prop? Maybe imagine this one to be used as "shape" or something similar, as if it has multiple options? square, rounded, circular... 

I really like this article @andresin87 shared time ago
https://spicefactory.co/blog/2019/03/26/how-to-avoid-the-boolean-trap-when-designing-react-components/

components/atom/button/src/index.scss
```diff
@@ -345,6 +349,10 @@ $icon-stroke-color: currentColor !default;
     }
   }
 
+  &--rounded {
+    border-radius: $bdrs-atom-button-rounded;
```
jcger: shouldn't this be using the rounded border token and thats it?

andresin87: Maybe is 4 combining both usages integrated

andresz1: I don't think so cuz the width and height of the button are different (see [MDN](https://developer.mozilla.org/en-US/docs/Web/CSS/border-radius))

andresz1:
> Do you think we could avoid Boolean for this prop? Maybe imagine this one to be used as "shape" or something similar, as if it has multiple options? square, rounded, circular...
> 
> I really like this article @andresin87 shared time ago
> https://spicefactory.co/blog/2019/03/26/how-to-avoid-the-boolean-trap-when-designing-react-components/

@turolopezsanabria I thought that maybe for this case was ok due I couldn't think of any other variation related. So what about `shape` prop with possible values to `square` (default) and `rounded` (this case). If no `children` is provided will be displayed as a `IconButton` as it is ATM.

andresin87:
> > Do you think we could avoid Boolean for this prop? Maybe imagine this one to be used as "shape" or something similar, as if it has multiple options? square, rounded, circular...
> > I really like this article @andresin87 shared time ago
> > [spicefactory.co/blog/2019/03/26/how-to-avoid-the-boolean-trap-when-designing-react-components](https://spicefactory.co/blog/2019/03/26/how-to-avoid-the-boolean-trap-when-designing-react-components/)
> 
> @turolopezsanabria I thought that maybe for this case was ok due I couldn't think of any other variation related. So what about `shape` prop with possible values to `square` (default) and `rounded` (this case). If no `children` is provided will be displayed as a `IconButton` as it is ATM.

prefer not to use boolean props 4 that case

components/atom/button/src/index.scss
turolopezsanabria: right, as @andresz1 comments, `rounded` token is actually an oval if the w and h are different

turolopezsanabria:
sure! `shape` for the name of the prop is perfect.

I suggest these values:

- squared (a central button of a group for instance)
- rounded (not circular, but actually rounded, as 8px in Fotocasa for instance) this is the default value is the prop is unchanged
- circular (fully rounded)

andresz1:
Changes made @turolopezsanabria @andresin87

stivaliserna (approved):

javiauso (approved):

andresin87 (approved):

PR #1300, opened 2020-11-18
feat(atom/progressBar): allow children to be displayed inside the com…

ferransimon:
…ponent

We need to allow children to be displayed instead of the percentage to cover some usecases

BREAKING CHANGE:
Now the size of the errorIcon is delagated to the client in order to simplify the code

## Atom/progressBar
**TASK**: 

### Types of changes

We need this change in order to be able to display the following usecases:

image

### DIFF
--- a/components/atom/progressBar/package.json
+++ b/components/atom/progressBar/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@s-ui/react-atom-progress-bar",
-  "version": "1.10.0",
+  "version": "2.0.0",
   "description": "",
   "main": "lib/index.js",
   "scripts": {
--- a/components/atom/progressBar/package.json
+++ b/components/atom/progressBar/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@s-ui/react-atom-progress-bar",
-  "version": "2.0.0",
+  "version": "1.10.0",
   "description": "",
   "main": "lib/index.js",
   "scripts": {
--- a/components/atom/progressBar/src/CircleProgressBar/index.scss
+++ b/components/atom/progressBar/src/CircleProgressBar/index.scss
@@ -1,7 +1,7 @@
 @import '~@s-ui/theme/lib/index';
 @import './settings';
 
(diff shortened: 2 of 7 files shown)


andresin87:
last manual small progress indicator looses its own %indicator text

components/atom/progressBar/src/CircleProgressBar/index.scss
```diff
@@ -3,49 +3,71 @@
 
 .sui-AtomCircleProgressBar {
   background-color: transparent;
-  display: inline-block;
-  position: relative;
```
andresin87, line 7: this might not be removed

andresin87:
If its a broken change the version might be modified in package. (progressBar is not already used by any other sui-components package).

The style naming might colide between versions so please, add an identifier class to prevent this posible library usage bug.

andresin87 (approved):

components/atom/progressBar/package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "@s-ui/react-atom-progress-bar",
-  "version": "1.10.0",
```
jcger: this should be set by the automatic release process, did you create a commit adding something to the breaking changes question?

jcger (approved):

components/atom/progressBar/package.json
ferransimon: Yes, I did, but @andresin87 told me it wasn't automatic so now I do not know what to do xD

jcger: It should be working, tell me if not

andresin87: The automatic release process upgrades the X.[number].0 version. As long as I know, this it the requirement to create a new major

javiauso (approved):

components/atom/progressBar/package.json
jcger: it changes the X if you fill the breaking changes question while running sui-studio commit command

midudev (approved):

PR #1310, opened 2020-11-25
feat(molecule/collapsible): disable max height

juanmendez90:
## [Molecule]/[Collapsible]
https://jira.scmspain.com/browse/SCMI-50632

### Types of changes

**Context**
When you use an imported font (bigger than the default one) and a short text that doesn't need to be collapsible there is a bug caused by a race condition between the component render and the font import, if the component render before the font when the font updates the text will be collapsed and there will be no show button to expand it

**Solution**
Set max height to none when the show button is false.

### Screenshots - Animations
Before
image

After
image


ruben-martin-lozano (approved):

oriolpuig (approved):

emiliovillu (approved):

andresin87 (commented):
👍

PR #1330, opened 2020-12-07
feat(atom/image): use picture tag around img and add sources prop

jcger:

Adds performant picture sources depending on the current device. We use the thumbnail component witch uses this one to show the company logo in our cards, these logos can be resized depending on query params in order to load the lower res images in mobile but we can't take advantage of them as this feature isn't implemented.

Since mdn's docs mention that the picture tag may include zero or more sources and one image and that it's semantically correct to use picture inside figure ([stack overflow](https://stackoverflow.com/a/23941403)), I've added the picture tag as fixed element within the component

### Types of changes

### DIFF
--- a/components/atom/image/src/index.js
+++ b/components/atom/image/src/index.js
@@ -29,7 +29,7 @@ const AtomImage = ({
   errorText,
   onError,
   onLoad,
-  sources,
+  sources = [],
   alt,
   ...imgProps
 }) => {
@@ -146,10 +146,6 @@ AtomImage.propTypes = {
 
   /** <img> props */
   ...htmlImgProps
-}
-
-AtomImage.defaultProps = {
-  sources: []
 }
 
 export default AtomImage
--- a/components/atom/image/src/index.js
+++ b/components/atom/image/src/index.js
@@ -137,7 +137,12 @@ AtomImage.propTypes = {
    * Source tags inside picture element,
    * array of props defined in https://developer.mozilla.org/en-US/docs/Web/HTML/Element/source expected
    */
-  sources: PropTypes.array,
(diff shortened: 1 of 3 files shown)


components/atom/image/src/index.js
```diff
@@ -29,6 +29,8 @@ const AtomImage = ({
   errorText,
   onError,
   onLoad,
+  sources,
```
andresin87: `sources = [],` and remove the default props section. react-docs parses this value too.

andresin87 (approved):

components/atom/image/src/index.js
```diff
@@ -80,13 +82,19 @@ const AtomImage = ({
...
-          {...imgProps}
-        />
+        <picture>
+          {sources.map((source, idx) => (
+            <source key={idx} {...source} />
+          ))}
+          <img
+            className={CLASS_IMAGE}
+            onLoad={handleLoad}
+            onError={handleError}
```
tonyghiani, line 92: I see the `handleLoad` function is memoized with useCallback, while the `handleError` function  is redefined for every render. In this case the handleLoad benefit is vanished by the declaration of handleError, which will be every time a new prop for the image tag, triggering a rerender of it. It's even possible that using useCallback in this case is a de-optimization, since it has compute that extra piece of logic with no real benefit. I may suggest to measure it and remove the useCallback memoization

tonyghiani (approved):

components/atom/image/src/index.js
jcger: I'll remove it

edit: cannot be done as it's executed by an effect

PR #1333, opened 2020-12-07
fix(atom/button): default props type deprecated usage mode

andresin87:
## ATOM/BUTTON
**TASK**: 

### Types of changes

when there is no given design or color prop, the default props callback is failing on the returning COLORS.PRIMARY value cuz it is not an object but a normal array of values


danivm (approved):

emiliovillu (approved):

turolopezsanabria (commented):
oops, now I see the buttons will render a deprecated prop (TYPE) in case no Prop is provided.

PR #1342, opened 2020-12-10
feat(atom/button): add alignText prop

Sabri1209:
## Atom/Button

Add new prop "alignText"  to be able to align text left or center. By default it will be "center".

### Types of changes
- [ ] Documentation Only
- [ ] Refactor

### Screenshots - Animations
Captura de pantalla 2020-12-10 a las 10 25 03

### DIFF
--- a/components/atom/button/src/config.js
+++ b/components/atom/button/src/config.js
@@ -18,6 +18,11 @@ export const DESIGNS = {
   LINK: 'link'
 }
 
+export const ALIGN_TEXT = {
+  CENTER: 'center',
+  LEFT: 'left'
+}
+
 /**
  * Available colors for the button
  */
@@ -82,6 +87,7 @@ export const OWN_PROPS = [
   'design',
   'focused',
   'fullWidth',
+  'alignText',
   'groupPosition',
   'leftIcon',
   'negative',
--- a/components/atom/button/src/config.js
+++ b/components/atom/button/src/config.js
@@ -88,7 +88,7 @@ export const OWN_PROPS = [
   'design',
   'focused',
   'fullWidth',
-  'alignText',
+  'alignment',
   'groupPosition',
   'leftIcon',
   'negative',
--- a/components/atom/button/src/config.js
+++ b/components/atom/button/src/config.js
@@ -18,9 +18,10 @@ export const DESIGNS = {
   LINK: 'link'
 }
 
-export const ALIGN_TEXT = {
+export const ALIGNMENT = {
   CENTER: 'center',
-  LEFT: 'left'
+  LEFT: 'left',
+  RIGHT: 'right'
 }
 
 /**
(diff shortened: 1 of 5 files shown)


components/atom/button/src/index.js
```diff
@@ -213,6 +218,10 @@ AtomButton.propTypes = {
    * Size of button 'small' (default), 'large'
    */
   size: PropTypes.oneOf(Object.values(SIZES)),
+  /**
+   * Align text 'center' (default), 'left'
+   */
+  alignText: PropTypes.bool,
```
arnau-rius: Should be a String?

components/atom/button/src/index.js
```diff
@@ -69,6 +71,7 @@ const getPropsWithDefaultValues = props => {
     color,
     design,
     type,
+    alignText: alignText || 'center',
```
arnau-rius: replace 'center' by `ALIGN_TEXT.CENTER`

components/atom/button/src/index.js
arnau-rius: Even Better, `alignText: PropTypes.oneOf(Object.keys(ALIGN_TEXT))`

Sabri1209: Ups!

components/atom/button/src/index.js
```diff
@@ -213,6 +218,10 @@ AtomButton.propTypes = {
    * Size of button 'small' (default), 'large'
    */
   size: PropTypes.oneOf(Object.values(SIZES)),
+  /**
+   * Align text 'center' (default), 'left'
```
arnau-rius: maybe it is more accurate to say that this prop aligns the content (not only the text) because icons are also affected by this alignment. 👍🏽

arnau-rius (approved):

components/atom/button/src/config.js
```diff
@@ -18,6 +18,11 @@ export const DESIGNS = {
   LINK: 'link'
 }
 
+export const ALIGN_TEXT = {
```
andresin87: why the right alignment is not offered?

andresin87: better call it `ALIGNEMENT`, also for the prop but lower case.

turolopezsanabria: The comment about the RIGHT alignment is valid for consistency, although we don't have right to left languages in MA, so it's hard to imagine a use case fo it

components/atom/button/src/index.js
```diff
@@ -213,6 +218,10 @@ AtomButton.propTypes = {
    * Size of button 'small' (default), 'large'
    */
   size: PropTypes.oneOf(Object.values(SIZES)),
+  /**
+   * Align content 'center' (default), 'left'
```
andresin87: remember you must update this comment

andresin87:
Add some Article section with the diferent alignement options:
| default | left   | center | right  |
| ------- | ------ | ------ | ------ |
| button  | button | button | button |

please

javiauso (approved):

andresin87 (approved):

PR #1343, opened 2020-12-10
feat(molecule/modal): add size and contentless props

andresz1:
## Molecule/Modal
**TASK**: None

### Types of changes

- [ ] Documentation Only
- [x] Refactor

This PR adds footer and size support. My ideal API for the modal would be something like this using compound components pattern:
```jsx
<Modal isOpen={isOpen} onClose={onClose}>
  <ModalHeader>Title</ModalHeader>
  
  <ModalCloseButton />

  <ModalContent>
    <Lorem count={2} />
  </ModalContent>

  <ModalFooter>
    <Button onClick={onClose}>
      Close
    </Button>
  </ModalFooter>
</Modal>
```

These changes could be a first iteration to reach that point opening the API to use composition using `isContentless` prop. I wanted to **avoid breaking changes** that's the reason why used that prop. Let me know if you think that would be better to create a major.

```jsx
<Modal header="Title" isOpen={isOpen} onClose={onClose} size="medium" isContentless>
  <Modal.Content>
    <Lorem count={2} />
  </Modal.Content>

  <Modal.Footer>
    <Button onClick={onClose}>
      Close
    </Button>
  </Modal.Footer>
</Modal>
```

🚨 Is you use the modal as contentless, you will need to set `enableContentScroll` and `withoutIndentation` props in the `Modal.Content` component instead of `Modal` just to keep it simple I didn't want to create a context for the modal yet.

Use `size` prop to choose a max width. If it's not set it will be like it was. 

### Screenshots - Animations

Scroll with footer

Screenshot 2020-12-10 at 11 37 54

### DIFF
--- a/components/molecule/modal/src/index.js
+++ b/components/molecule/modal/src/index.js
@@ -15,8 +15,17 @@ import {SUPPORTED_KEYS} from './config'
 import {suitClass} from './helpers'
 import {Close} from './Close'
 import {HeaderRender} from './HeaderRender'
+import MoleculeModalContent from './Content'
+import MoleculeModalFooter from './Footer'
 import WithAnimation from './HoC/WithAnimation'
 import WithUrlState from './HoC/WithUrlState'
+
+export const MODAL_SIZES = {
+  XSMALL: 'xsmall',
+  SMALL: 'small',
+  MEDIUM: 'medium',
+  LARGE: 'large'
+}
 
 const toggleWindowScroll = disableScroll => {
   window.document.body.classList.toggle('is-MoleculeModal-open', disableScroll)
@@ -30,6 +39,7 @@ const MoleculeModal = ({
   fitContent = false,
   fitWindow = false,
   floatingIconClose = false,
+  size,
   header,
   iconClose = false,
   isClosing,
@@ -38,9 +48,9 @@ const MoleculeModal = ({
   onClose = () => {},
   portalContainerId = 'modal-react-portal',
   usePortal = true,
-  withoutIndentation = false
+  withoutIndentation = false,
+  isContentless
 }) => {
-  const contentRef = useRef()
   const wrapperRef = useRef()
 
   const [isClientReady, setIsClientReady] = useState(false)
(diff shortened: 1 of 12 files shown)


components/molecule/modal/src/index.scss
```diff
@@ -182,31 +193,4 @@ body.is-MoleculeModal-open {
...
-      height: $pb-molecule-modal-content;
-    }
-
-    &:first-child {
-      margin-top: 0;
-    }
-
-    &:last-child {
-      margin-bottom: 0;
-    }
```
andresz1: IMO this is not needed

components/molecule/modal/src/index.scss
```diff
@@ -182,31 +193,4 @@ body.is-MoleculeModal-open {
       width: $w-molecule-modal-icon;
     }
   }
-
-  &-content {
-    -webkit-overflow-scrolling: touch;
-    flex: 1 1 auto;
-    overflow-y: auto;
-    margin: $mt-molecule-modal-content $mr-molecule-modal-content 0
```
andresz1: Changed this to use padding instead because of the scroll. I think that it won't cause any breaking change. Any suggestion @andresin87 ?

package.json
```diff
@@ -37,6 +37,7 @@
     "fs-extra": "9.0.1",
     "globby": "11.0.1",
     "husky": "4.3.0",
+    "react-lorem-component": "^0.13.0",
```
andresin87: "react-lorem-component": "0",

javiauso (approved):

components/molecule/modal/src/index.js
```diff
@@ -125,11 +120,8 @@ const MoleculeModal = ({
     const dialogClassName = cx(suitClass({element: 'dialog'}), {
       [suitClass({element: 'dialog--full'})]: fitWindow,
       [suitClass({element: 'dialog--out'})]: isClosing,
-      [suitClass({element: 'dialog--fit'})]: fitContent
-    })
-
-    const contentClassName = cx(suitClass({element: 'content'}), {
-      [suitClass({element: 'content--without-indentation'})]: withoutIndentation
+      [suitClass({element: 'dialog--fit'})]: fitContent,
+      [suitClass({element: `dialog--${size}`})]: !!size
```
rmoralp, line 125: [Subjective opinion] Could we use a descriptive way? `Boolean(size)`.
It is more semantic

rmoralp (commented):
❤️  compound components

package.json
andresz1: Is not needed IMO since is a external dep

andresin87: remove ^ anyway

andresin87: it might not be in top level dep. Only in the demo needed.

components/molecule/modal/src/index.scss
andresin87: checked

stivaliserna (approved):

components/molecule/modal/src/index.js
```diff
@@ -89,34 +99,19 @@ const MoleculeModal = ({
...
     }
   }
 
-  const extendedChildren = Children.toArray(children).map(child =>
-    cloneElement(child, {
-      onClose: closeModal
-    })
-  )
+  const renderChildren = () => {
+    return Children.toArray(children).map(child =>
```
andresin87: remove brackets. Arrow fn direct

components/molecule/modal/src/index.js
```diff
@@ -242,6 +240,10 @@ MoleculeModal.propTypes = {
    * The callback function to be called when the animation ends, defined by the HoC WithAnimation
    */
   onAnimationEnd: PropTypes.func,
+  /**
+   * If true children won't be wrapped with content
```
andresin87: the explanation might not have details about the component implementation. What does content means? How does it affect (behavior-visually)?

Issue #1352, opened 2020-12-15
ATOM/BUTTON


xaviermariaconejo:
The Atom Button component has a constant COLORS in the config.js file, but this is an array, not like others constants like DESIGNS that is an object, so we can refer to the colors using the CONSTANT and not using "magic strings" like write "primary"

The main problem is that the export of the COLORS constant should be a object too, and right now is an array, so this will be breaking change

Issue #1335, opened 2020-12-07
Default styles of buttons if NO Prop is provided: Should be solid design + primary color


turolopezsanabria:
**Describe the bug**
When no prop is provided the buttons defaults to a deprecated prop (type) causing a deprecated style.
![deprecated type](https://user-images.githubusercontent.com/23620759/101387643-03e23f00-38bf-11eb-906d-35e06fd89b39.png)

**Steps to reproduce**
Go to Buttons
Navigate to MA theme
Scroll down to the bottom of the page

**Expected behaviour**
Ir should never default-back to the deprecated prop "type". Instead it should be `design: solid` + `color:primary`

**Screenshots**
In Milanuncios theme is easy to spot how this affects the styles.
![broken style](https://user-images.githubusercontent.com/23620759/101387711-21170d80-38bf-11eb-9ea3-9158dbd9248c.png)

**Additional info**
Latest PR related with this issue is https://github.com/SUI-Components/sui-components/pull/1333/

Issue #1312, opened 2020-11-25
Button Component - color default


xaviermariaconejo:
**Is your iteration request related to a problem?**
In the component [Button](https://sui-components.now.sh/workbench/atom/button/documentation/api) is detailed that the default color value is "primary". But, if i choose the design="link" the default color is not primary, in fact i don't know from where is picking up that color, but with the theme of Habitaclia it pains it blue. Then when I choose in specific the color="primary" is orange

**Describe the solution you'd like**
I would like by default the color is the primary of the theme

PR #1353, opened 2020-12-15
feat(atom/button): fix default color & design value

xaviermariaconejo:
## ATOM/BUTTON
**ISSUE**: [1335](https://github.com/SUI-Components/sui-components/issues/1335)
**ISSUE**: [1312](https://github.com/SUI-Components/sui-components/issues/1312)
**ISSUE**: [1352](https://github.com/SUI-Components/sui-components/issues/1352)

### Types of changes

- [X] Breaking change (fix or feature that would cause existing functionality to not work as expected)
- [ ] Documentation Only
- [ ] Refactor

Fix default values of color & design
Fix COLORS variables is a object

### Screenshots - Animations
image
image


adriadoz (approved):
👏

andresin87:
In one hand, I personally agree with the default color in case of not providing any color, but I think the default design as solid might only work if the link boolean prop is not activated and 4 that case, the L&F might be link the design link

On the other hand... are we creating a major only 4 the array to constants object changes? I suggest to postpose this change and include many other features like the deprecated type prop removal.

andresin87:
[https://sui-components-kcsot6jy5.vercel.app/workbench/atom/button/demo](https://sui-components-kcsot6jy5.vercel.app/workbench/atom/button/demo)

danivm (approved):

nucliweb (approved):

midudev:
> In one hand, I personally agree with the default color in case of not providing any color, but I think the default design as solid might only work if the link boolean prop is not activated and 4 that case, the L&F might be link the design link
> 
> On the other hand... are we creating a major only 4 the array to constants object changes? I suggest to postpose this change and include many other features like the deprecated type prop removal.

If we could avoid doing a major for the Atom Button while fixing any issue, then we should avoid it. Atom Button is a component used over lots of verticals and doing this will produce several steps on different levels.

components/atom/button/src/index.js
```diff
@@ -55,52 +52,116 @@ const getModifiers = props => {
...
-    loader = <ButtonSpinnerIcon />
-  } = getPropsWithDefaultValues(props)
+    title
+  } = getPropsWithDefaultValues(typeConversion(props))
 
   const classNames = cx(
     CLASS,
-    !type && COLOR_CLASSES[color],
-    !type && CLASSES[design],
-    type && CLASSES[type],
```
andresin87, line 103: `type` classes definitely removed! 🥳

We are using an adapter to transform it to the `design` and `color` combo in
`typeConversion` fn

andresin87:
[https://sui-components-1xp71ugv8.vercel.app/workbench/atom/button/demo](https://sui-components-1xp71ugv8.vercel.app/workbench/atom/button/demo)

components/atom/button/src/index.js
```diff
@@ -55,52 +52,116 @@ const getModifiers = props => {
   )
 }
```
andresin87, line 53: I think the `deprecated` fn might be located on another place (other library or other file in different folder structure maybe)

andresin87:
Major is no longer needed until new features

components/atom/button/src/index.js
```diff
@@ -55,52 +52,116 @@ const getModifiers = props => {
...
-  } = getPropsWithDefaultValues(props)
+    title
+  } = getPropsWithDefaultValues(typeConversion(props))
 
   const classNames = cx(
     CLASS,
-    !type && COLOR_CLASSES[color],
-    !type && CLASSES[design],
-    type && CLASSES[type],
+    COLOR_CLASSES[color],
```
xaviermariaconejo: Oooone little question, COLOR_CLASSES is till needed then? or the color classes names could be placed in CLASSES directly now since type has disappeared?

andresin87: yup, i'll make that iteration tomorrow 👍

andresin87: DONE

Issue #1205, opened 2020-08-20
Can not apply ellipsis to button link


xaviermariaconejo:
**Is your iteration request related to a problem?**
I need to add ellipsis styles, the typical three dots, but the current component doesn't allows me to do that

**Describe the solution you'd like**
When it is a button with link design without href, it renders a button and inside a span, I can apply the correct styles to apply the ellipsis on the button, but the span breaks that styles. I would chose to remove that span, or add the style: display: inline-block, to fix it

turolopezsanabria:
hm.. yup I see this :S 
we definitely have to find a solution (without a major release) for buttons rendered and styled as `a` links.

xaviermariaconejo:
En la reunión de SUI-Components hemos comentado que este tipo de cambios, que son bastante particulares dentro del vertical, mejor que se queden dentro del vertical, con el approach que el propio vertical elija, desde componentes comunes dentro del vertical, clases de css, etc.

Y si hay una parte de estilos que se quieran aprovechar se pueden utilizar los estilos que te da SUI-Theme, desde clases hasta mixins.

Así que se cierra este ISSUE ✌️

PR #1362, opened 2020-12-22
feat(atom/button): add styles to be possible to add ellipsis to the b…

xaviermariaconejo:
…utton with link design

## ATOM/BUTTON
**ISSUE**: [1205](https://github.com/SUI-Components/sui-components/issues/1205)

### Types of changes

- [ ] Documentation Only
- [ ] Refactor

Developing I found that with the current styles is not possible to add ellipsis styles to the possible Atom Button, I think that add ellipsis styles should happens only with design="link", so I add the styles needed so with a className to the button is possible now to add custom styles.

I'm not 100% sure if use the className prop is the best solution, I'm looking forward to your feedback

### Screenshots - Animations
**SOLUTION**
image
image

**CODE EXAMPLE**
I will do a copy paste from the explanation in slack:

Vale, he hecho una prop ellipsis que añade los estilos de ellipsis. Y luego, el tamaño del ellipsis de por si mismo, lo he dejado para que sea en funcion del tamaño del wrapper que tenga el div.

 Es decir, que para utilizarlo sea en plan:
```
<div className="wrapper">
    <AtomButton
         link
         ellipsis
         href="www.tosomewhere.com">
     Hola soy un link con ellipsis
    </AtomButton>
</div>
```

Y es este div, que puede aplicar como estilo:
```
.wrapper {
    width: 50px; // or whatever
}
```

Basicamente porque el width a poner sera muy custom para cada caso, y de esta manera pues cada developer podrá poner la que quiera y como quiera :+1:

### DIFF
--- a/components/atom/button/src/index.scss
+++ b/components/atom/button/src/index.scss
@@ -203,13 +203,10 @@ $icon-stroke-color: currentColor !default;
     display: inline-flex;
     height: 100%;
     pointer-events: none;
+    justify-content: center;
   }
 
   &--loading {
-    #{$base-class}-inner {
-      justify-content: center;
-    }
-
     #{$base-class}-text {
       opacity: 0;
     }
@@ -221,7 +218,7 @@ $icon-stroke-color: currentColor !default;
   &-centerIcon {
     display: inline-flex;
 
-    // add compatibility with SuiAtomIcon with correct alignment
+    // Add compatibility with SuiAtomIcon with correct alignment
     .sui-AtomIcon {
       display: inline-flex;
     }
@@ -234,17 +231,25 @@ $icon-stroke-color: currentColor !default;
 
   &-centerIcon {
     position: absolute;
+    top: 50%;
+    left: 50%;
+    transform: translate(-50%, -50%);
   }
 
   &#{$base-class}--link:not(#{$base-class}--solid):not(#{$base-class}--outline):not(#{$base-class}--flat) {
-    background: transparent;
-
-    border: none;
     @include link();
-    padding: 0;
+    background: transparent;
+    border: none;
+    height: auto;
     justify-content: center;
+    padding: 0;
     text-align: center;
     text-decoration: none;
+
+    #{$base-class}-inner {
+      display: inline;
+    }
+
     @include button-focused {
       background: transparent;
     }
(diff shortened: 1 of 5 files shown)


demo/atom/button/demo/index.scss
```diff
@@ -0,0 +1,6 @@
+.AtomButtonLinkEllipsis {
+  width: 60px;
```
adriadoz: Why 60px? It should be a variable or maybe prop to be able to adjust it depending on the need?

components/atom/button/src/index.scss
```diff
@@ -234,17 +231,25 @@ $icon-stroke-color: currentColor !default;
...
-    @include link();
-    padding: 0;
+    height: auto;
     justify-content: center;
+    padding: 0;
     text-align: center;
     text-decoration: none;
+
+    #{$base-class}-inner {
+      display: inline;
```
PablitoGS: ellipsis should work without this modification

xaviermariaconejo: Really? How? Exactly this is what I found why isn't working before

andresin87: I personally think this is too much specific and we should not increase that danger button for that purposes. Need time to think about it

xaviermariaconejo: I've changed to this `display: inline` will be changed only when the ellipsis prop is set to true

PablitoGS: I don't know if it's because there is a different structure but it should work
Captura de pantalla 2020-12-22 a las 15 47 21

xaviermariaconejo: Tell me how you did it, I'm still trying to figure it out to do that with the link design atom button, please

andresin87:
I think thats a better way of obtaining this behavior
https://primer.style/components/Truncate

xaviermariaconejo:
En la reunión de SUI-Components hemos comentado que este tipo de cambios, que son bastante particulares dentro del vertical, mejor que se queden dentro del vertical, con el approach que el propio vertical elija, desde componentes comunes dentro del vertical, clases de css, etc.

Y si hay una parte de estilos que se quieran aprovechar se pueden utilizar los estilos que te da SUI-Theme, desde clases hasta mixins.

Así que se cierra esta PR ✌️

PR #1370, opened 2021-01-08
feat(layout/grid): add gapless prop

danivm:
### Types of changes

- [ ] Documentation Only
- [ ] Refactor

Add `isGapless` prop to remove items padding.

### Screenshots - Animations
Captura de pantalla 2021-01-08 a las 12 35 24

### DIFF
--- a/components/layout/grid/src/index.js
+++ b/components/layout/grid/src/index.js
@@ -4,11 +4,12 @@ import cx from 'classnames'
 
 import {ALIGN_ITEMS, BASE_CLASS, JUSTIFY_CONTENT} from './settings'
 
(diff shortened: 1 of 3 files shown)


AitorRodriguez990 (approved):

emiliovillu (approved):

andresin87:
I think the right naming 4 those purposes is gutter

turolopezsanabria:
I suggest NOT using Boolean props if we can. Can we have a Prop with 4 possible spacings and only one valid value? What do you think? 
i.e. Gutter none, small, medium, large?

PR #1372, opened 2021-01-11
Feature/drawer

ferransimon:
## [MOLECULE]/[DRAWER]
**TASK**: 

### Types of changes

- [ ] Documentation Only
- [ ] Refactor

Created a new molecule Drawer

### Screenshots - Animations

drawer_demo


components/molecule/drawer/README.md
```diff
@@ -0,0 +1,34 @@
+# MoleculeDrawer
+
+> Description
+
```
xaviermariaconejo, line 4: Add Description

components/molecule/drawer/README.md
```diff
@@ -0,0 +1,34 @@
...
+#### Import the styles (Sass)
+
+```css
+@import '~@s-ui/theme/lib/index';
+// @import 'your theme';
+@import '~@s-ui/react-molecule-drawer/lib/index';
+```
+
+
+> **Find full description and more examples in the [demo page](#).**
```
xaviermariaconejo: Add the link or remove this line

components/molecule/drawer/src/index.js
```diff
@@ -0,0 +1,112 @@
...
+
+const Overlay = forwardRef((props, ref) => {
+  return <div ref={ref} className="react-MoleculeDrawer-overlay" {...props} />
+})
+
+const Body = props => {
+  return <div className="react-MoleculeDrawer-body" {...props} />
+}
+
+const Content = forwardRef((props, ref) => {
```
xaviermariaconejo: For each one of this little components you should add the displayName & propTypes

ferransimon: They are just to give semantics, do you think that addind displayName & propTypes will improve anything at all?

components/molecule/drawer/src/index.js
```diff
@@ -0,0 +1,112 @@
...
+        }}
+      >
+        <Content
+          ref={contentRef}
+          onAnimationEnd={handlerOnAnimationEnd}
+          className={cx(
+            'react-MoleculeDrawer-content',
+            `react-MoleculeDrawer-content--${placement}`
+          )}
+          style={{
```
xaviermariaconejo, line 70: there is no other way to approach this? with classNames? I don't think use the style prop is the best approach. But may be I'm wrong

components/molecule/drawer/src/index.js
```diff
@@ -0,0 +1,112 @@
...
+  )
+
+  return isClientReady ? createPortal(drawer, getContainer()) : null
+}
+
+MoleculeDrawer.displayName = 'MoleculeDrawer'
+MoleculeDrawer.propTypes = {
+  isOpen: PropTypes.bool,
+  onClose: PropTypes.func.isRequired,
+  placement: PropTypes.oneOf(['right', 'left', 'bottom', 'top'])
```
xaviermariaconejo: Instead of hardcoded strings "right, left, bottom, top", they should be places in a constant object and export it, so the developer who use this components could avoid use hardcoded string too

xaviermariaconejo (approved):

xaviermariaconejo:
Awesome work mate!

components/molecule/drawer/src/index.js
xaviermariaconejo: Hmmmm I don't, and I'm ok this way, BUT, the linter we have mark it as an error if a React Component doesn't have displayName & propTypes. So, up to you 👍

tonyghiani: If the main purpose is the semantic, i think creating 3 new functions, forwarding ref and adding display names is a bit overengineered, i'd suggest to pass directly the className to the element in the return statement of the component and create renamed tags to give a semantic to the main component:
```
const Body = 'div'
const Overlay = 'div'
const Content = 'div'
```

The pass directly the className where required. I see it was duplicated for the **react-MoleculeDrawer-content**

components/molecule/drawer/src/index.js
ferransimon: Since it is dynamic based on 2 props I have not found any other way, maybe I am missing something since I am not quite good in animations.

components/molecule/drawer/src/index.js
```diff
@@ -0,0 +1,117 @@
...
+
+export default function MoleculeDrawer({
+  portalContainerId = 'drawer-react-portal',
+  isOpen = false,
+  placement = 'left',
+  onClose,
+  children
+}) {
+  const overlayRef = useRef(null)
+  const contentRef = useRef(null)
```
andresz1: Is `contentRef` being used ?

components/molecule/drawer/src/index.js
```diff
@@ -0,0 +1,117 @@
...
+      if (ev.key === 'Escape') {
+        onClose(ev)
+        ev.preventDefault()
+      }
+    },
+    [isOpen, onClose]
+  )
+
+  useEffect(() => {
+    document.removeEventListener('keydown', onKeyDown)
```
andresz1: Is this truly needed ?

components/molecule/drawer/src/index.js
```diff
@@ -0,0 +1,117 @@
...
+  TOP: 'top',
+  RIGHT: 'right',
+  BOTTOM: 'bottom',
+  LEFT: 'left'
+}
+
+export default function MoleculeDrawer({
+  portalContainerId = 'drawer-react-portal',
+  isOpen = false,
+  placement = 'left',
```
andresz1: `placement = PLACEMENTS.LEFT` ?

components/molecule/drawer/src/index.js
ferransimon: Good point, thanks!

components/molecule/drawer/src/index.js
ferransimon: I guess not, we are already cleaning up everything in the unmount callback

tonyghiani (approved):

javiauso (approved):

PR #1374, opened 2021-01-11
feat(molecule/pagination): add show edges + size

sergiqm94:
## molecule/pagination
**TASK**: https://jira.scmspain.com/browse/SCMI-52258

### Types of changes

- [ ] Documentation Only
- [ ] Refactor

We have created the prop **size** to control the size of the buttons using the "size" variable of the atomButton.

We have also created the prop **showEdges** to show always the first page and the last one as you have in the picture.

### Screenshots - Animations

**Size**
image

**Show edges**
image
image
image

### DIFF
--- a/components/molecule/pagination/src/helpers/pagination/lowRange.js
+++ b/components/molecule/pagination/src/helpers/pagination/lowRange.js
@@ -10,6 +10,7 @@
 const lowRange = ({page, showEdges, showPages, totalPages}) => {
   const lowRangeFixer = Math.floor(showPages / 2)
   let lowRange
+
   if (page % showPages) lowRange = page - (page % showPages)
   else lowRange = page - showPages
 
@@ -18,8 +19,7 @@ const lowRange = ({page, showEdges, showPages, totalPages}) => {
 
     if (lowRange + showPages === page) {
       lowRange = page - lowRangeFixer
-    }
-    if (lowRange === page - 1) {
+    } else if (lowRange === page - 1) {
       lowRange = lowRange - lowRangeFixer
     }
   }
--- a/components/molecule/pagination/src/helpers/pagination/lowRange.js
+++ b/components/molecule/pagination/src/helpers/pagination/lowRange.js
@@ -6,9 +6,25 @@
  * @param {number} params.showPages - Number of pages to display in the range of pages displayed
  * @return {number} lowest number of the assigned range for the page
  */
-const lowRange = ({page, showPages}) => {
-  if (page % showPages) return page - (page % showPages)
-  return page - showPages
+
+const lowRange = ({page, showEdges, showPages, totalPages}) => {
+  const lowRangeFixer = Math.floor(showPages / 2)
+  let lowRange
+  if (page % showPages) lowRange = page - (page % showPages)
+  else lowRange = page - showPages
+
+  if (showEdges) {
+    if (page <= showPages) return 1
+
+    if (lowRange + showPages === page) {
+      lowRange = page - lowRangeFixer
+    }
+    if (lowRange === page - 1) {
+      lowRange = lowRange - lowRangeFixer
+    }
+  }
+
+  return Math.min(lowRange, totalPages - showPages)
 }
 
 export default lowRange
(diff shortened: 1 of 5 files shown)


adriadoz (approved):
Nice!

components/molecule/pagination/src/index.scss
```diff
@@ -15,6 +15,13 @@ $c-pagination-button-icon-path: inherit !default;
     margin-left: $m-s;
   }
 
+  &-divider {
+    margin-left: $m-s;
+    display: flex;
+    justify-content: center;
+    align-items: center;
```
adriadoz: Alphabetic order please :D

components/molecule/pagination/src/helpers/pagination/lowRange.js
```diff
@@ -6,9 +6,25 @@
...
+  if (page % showPages) lowRange = page - (page % showPages)
+  else lowRange = page - showPages
+
+  if (showEdges) {
+    if (page <= showPages) return 1
+
+    if (lowRange + showPages === page) {
+      lowRange = page - lowRangeFixer
+    }
+    if (lowRange === page - 1) {
```
xaviermariaconejo: why don't use else if?

components/molecule/pagination/src/index.js
```diff
@@ -11,6 +11,7 @@ const BASE_CLASS = 'sui-MoleculePagination'
 const CLASS_PREV_BUTTON_ICON = 'sui-MoleculePagination-prevButtonIcon'
 const CLASS_NEXT_BUTTON_ICON = 'sui-MoleculePagination-nextButtonIcon'
 const PAGE_NUMBER_HOLDER = '%{pageNumber}'
+const divider = '···'
```
xaviermariaconejo: If it's a constant you should name it in mayus: DIVIDER

components/molecule/pagination/src/index.js
```diff
@@ -127,31 +132,87 @@ const MoleculePagination = ({
...
+                }
+                onSelectPage={onSelectPage}
+                size={size}
+                {...linkProps(totalPages)}
+              >
+                {totalPages}
+              </PageButton>
+            </>
+          )}
+        </>
```
xaviermariaconejo, line 215: I think that file is too large, it makes it hard to read and maintenance, do you think we could split it somehow? Anyway this is something that happens in a lot of other components so I'm fine to merge it like this, but I think would be nice to think ways to split components this large

demo/molecule/pagination/demo/index.js
```diff
@@ -39,6 +39,14 @@ const Demo = () => {
...
       <div className="sui-StudioPreview-content sui-StudioDemo-preview">
         <h1>Pagination</h1>
+        <h2>Size</h2>
+        <div className={CLASS_DEMO_SECTION}>
+          <h3>Small</h3>
+          <MoleculePagination totalPages={25} page={17} size="small" />
+          <h3>Large</h3>
+          <MoleculePagination totalPages={25} page={17} />
+        </div>
+        {/* ------------------------------------------------------------------------------------------------------------- */}
```
xaviermariaconejo: I think this is a kind of comment to separate the code? You could write something instead of only "----" so other developers will know why and what are you indicating

xaviermariaconejo (approved):

demo/molecule/pagination/demo/index.js
sergiqm94: This comment was already applied in different parts of the file, I guess they put it to differentiate the different parts of the demo.

I think it's good to remove them. Instead I leave a separation line in the code.

components/molecule/pagination/src/index.js
sergiqm94: I don't think that's necessary. We could create a component for the edges but I don't think that makes sense because it's only going to be used for pagination.

components/molecule/pagination/src/helpers/pagination/lowRange.js
sergiqm94: Yes we can use it

midudev (approved):

PR #1373, opened 2021-01-11
feat(molecule/notification): add outline style to notification molecule

xaviermariaconejo:
## [MOLECULE]/[NOTIFICATION]
**TASK**: [SUIC-249](https://jira.scmspain.com/browse/SUIC-249)
**ISSUE**: [1270](https://github.com/SUI-Components/sui-components/issues/1270)

### Types of changes

- [ ] Bug fix (non-breaking change which fixes an issue)
- [X] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to not work as expected)
- [ ] Documentation Only
- [ ] Refactor

### Description, Motivation and Context

Add new style: Outline

Fix alignment of text to centered

Export different possible values for props of that component

### Screenshots - Animations

image

PR #1378, opened 2021-01-14
fix(molecule/notification): fix useEffect infinte bucle

xaviermariaconejo:
### Types of changes

- [ ] Documentation Only
- [ ] Refactor

Habia un bucle infinito en el useEffect causado por la dependencia: triggerAutoClose. Para mi no tiene mucho sentido poner una dependencia en el useEffect que sea una funcion que es llamada solo si se cumple una condicion booleana (como en este caso lo es triggerAutoClose que tiene un if), en ese caso, la dependencia que sea la condicion booleana y ya esta, porque puede ocasionar bucles infinitos que llamen todo el rato a ese useEffect, en este caso, parece que pasaba eso

### DIFF
--- a/components/molecule/notification/src/index.js
+++ b/components/molecule/notification/src/index.js
@@ -75,7 +75,7 @@ const MoleculeNotification = ({
       clearTimeout(autoCloseTimeout.current)
       clearTimeout(transitionTimeout.current)
     }
-  }, [show, triggerAutoClose, effect, autoCloseTimeInSeconds])
+  }, [show, effect, autoCloseTimeInSeconds])
 
   const getButtons = () =>
     buttons


jordevo (approved):

jordevo:
I'm afraid we might be introducing a new bug or some unexpected behaviour here...

Instead of removing `triggerAutoClose` from the `useEffect` deps, to me it seems like this:
https://github.com/SUI-Components/sui-components/pull/1373/files#diff-eba1b23725e14c510bfd9518504556bbdaa4170e8408c39c474d85874528f22fR20

introduced the bug in the first place, so if anything needs to go, to me it'd be the default value set for `autoCloseTiming`. What do you think @xaviermariaconejo @kikoruiz @midudev ?

kikoruiz (approved):

jordevo (approved):

PR #1408, opened 2021-02-22
feat(layout/grid): v2

andresin87:
## LAYOUT/GRID
**TASK**: 

### Types of changes

- [ ] Documentation Only
- [x] Refactor

This major version adds the possibility to set different gutters even per breakpoint. Ables the transition to substitute the `isGapless` boolean prop for `gutter={0}`, makes different sizes do not collide per breakpoint, removes unnecessary paddings in order to become nested friendly grid and lots of etc.


beatrizip (approved):

components/layout/grid/src/gridItem/index.js
```diff
@@ -1,10 +1,30 @@
 import PropTypes from 'prop-types'
 import cx from 'classnames'
 
-import {BASE_CLASS, CELL_NUMBERS} from '../settings'
+import {BASE_CLASS, CELL_NUMBERS, BREAKPOINTS} from '../settings'
+
+const getColSpanClassNamesTransform = ({colSpan, ...otherProps}) => {
```
jcger: what about unit testing this? seems to have enough complexity to justify it

components/layout/grid/src/gridItem/index.js
```diff
@@ -1,10 +1,30 @@
...
+  const className = []
+  if (CELL_NUMBERS.includes(colSpan) && otherProps.xxs === undefined) {
+    className.push(`${BASE_CLASS}-item--xxs-${colSpan}`)
+  } else if (typeof colSpan === 'object') {
+    Object.values(BREAKPOINTS).forEach(breakpoint => {
+      if (
```
jcger: maybe you could filter this items out instead of using an if

components/layout/grid/src/gridItem/index.js
```diff
@@ -1,10 +1,30 @@
 import PropTypes from 'prop-types'
 import cx from 'classnames'
 
-import {BASE_CLASS, CELL_NUMBERS} from '../settings'
+import {BASE_CLASS, CELL_NUMBERS, BREAKPOINTS} from '../settings'
+
+const getColSpanClassNamesTransform = ({colSpan, ...otherProps}) => {
+  const className = []
+  if (CELL_NUMBERS.includes(colSpan) && otherProps.xxs === undefined) {
+    className.push(`${BASE_CLASS}-item--xxs-${colSpan}`)
```
jcger: as this is only one key and nothing else will be added, you could just return the string and avoid the push/join cost

components/layout/grid/src/gridItem/index.js
```diff
@@ -22,6 +42,7 @@ export default function LayoutGridItem({
 }) {
   const classNames = cx(
```
jcger, line 67: would this be worth to be cached?

components/layout/grid/src/gridItem/index.js
```diff
@@ -1,10 +1,30 @@
 import PropTypes from 'prop-types'
 import cx from 'classnames'
 
-import {BASE_CLASS, CELL_NUMBERS} from '../settings'
+import {BASE_CLASS, CELL_NUMBERS, BREAKPOINTS} from '../settings'
+
+const getColSpanClassNamesTransform = ({colSpan, ...otherProps}) => {
+  const className = []
+  if (CELL_NUMBERS.includes(colSpan) && otherProps.xxs === undefined) {
```
jcger: what's the point of xxs being undefined? Maybe a descriptive fn or a comment would help to understand what's happening here

davidmartin84 (approved):

components/layout/grid/src/gridItem/index.js
andresin87: I'll use reduce

components/layout/grid/package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "@s-ui/react-layout-grid",
-  "version": "1.6.0",
```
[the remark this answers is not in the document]
andresin87: ok

components/layout/grid/src/gridItem/index.js
andresin87: done in the reduce

components/layout/grid/src/gridItem/index.js
andresin87: cached

andresin87:
all changes applied. Thanks @jcger 4 your suggestions 😉 👍

PR #1416, opened 2021-03-02
feat(atom/popover): add prop disableToggle

sergiqm94:
### Types of changes

- [ ] Documentation Only
- [ ] Refactor

**Problem**
We want to keep the popover visible until the user forces it to close (by clicking the "X").

**Solution**
We disable the native toggle so that it remains visible even if we perform actions both inside and outside the children by forcing the "X" to be pressed to close it.

### DIFF
--- a/components/atom/popover/src/index.js
+++ b/components/atom/popover/src/index.js
@@ -22,6 +22,7 @@ const Popover = lazy(() => import('reactstrap/lib/Popover'))
 function AtomPopover({
   children,
   closeIcon,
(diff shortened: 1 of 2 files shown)


danivm (approved):

ruben-martin-lozano (approved):

andresin87:
I think the right way might be tu have it as a controlled component. It means have a prop isOpen, where the user can control under onClose handler the value from external component

PR #1420, opened 2021-03-04
Feature/forward modal ref

tonyghiani:
### Types of changes

- [ ] Documentation Only
- [ ] Refactor

Forward the reference received in the modal component.
It is necessary in order to refer to the main wrapper and use it, for example, to measure the modal size.

### DIFF
--- a/components/molecule/modal/src/index.js
+++ b/components/molecule/modal/src/index.js
@@ -12,6 +12,7 @@ import {
 } from 'react'
 import {createPortal} from 'react-dom'
 import cx from 'classnames'
+import {useMergeRefs} from '@s-ui/react-hooks'
 import {SUPPORTED_KEYS} from './config'
 import {suitClass} from './helpers'
 import {Close} from './Close'
@@ -30,16 +31,6 @@ export const MODAL_SIZES = {
 
 const toggleWindowScroll = disableScroll => {
   window.document.body.classList.toggle('is-MoleculeModal-open', disableScroll)
-}
-
-const combineRefs = (...refs) => value => {
-  refs.forEach(ref => {
-    if (typeof ref === 'function') {
-      ref(value)
-    } else if (ref != null) {
-      ref.current = value
-    }
-  })
 }
 
 const MoleculeModal = forwardRef(
@@ -67,7 +58,7 @@ const MoleculeModal = forwardRef(
     forwardedRef
   ) => {
     const wrapperRef = useRef()
-    const ref = combineRefs(wrapperRef, forwardedRef)
+    const ref = useMergeRefs(wrapperRef, forwardedRef)
 
     const [isClientReady, setIsClientReady] = useState(false)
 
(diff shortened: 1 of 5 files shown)


demo/molecule/modal/demo/SizeModal.js
```diff
@@ -18,7 +18,7 @@ const SizeModal = () => {
   return (
     <>
       {Object.values(MODAL_SIZES).map(size => (
-        <button key="size" onClick={handleClick(size)}>
```
andresin87, line 21: 👀

components/molecule/modal/src/index.js
```diff
@@ -31,152 +32,168 @@ const toggleWindowScroll = disableScroll => {
...
-
-  const [isClientReady, setIsClientReady] = useState(false)
-
-  const getContainer = () => {
-    let containerDOMEl = document.getElementById(portalContainerId)
-    if (!containerDOMEl) {
-      containerDOMEl = document.createElement('div')
-      containerDOMEl.id = portalContainerId
-      document.body.appendChild(containerDOMEl)
+const combineRefs = (...refs) => value => {
```
andresin87: wtf?

tonyghiani: What do you mean? Is about the function logic or the implementation?

andresin87: 📞?

andresin87: in review

andresin87: Here I made an example indicating the formal proposed way 4 coding this using [that lib](https://github.com/thebuilder/react-intersection-observer) as described in the first FAQ of its readme. [https://codesandbox.io/s/condescending-pascal-kj3ff](https://codesandbox.io/s/condescending-pascal-kj3ff)

andresin87: I personally think its a clean solution to do that obtaining the same

tonyghiani: Mmm I have a couple of doubts fo the proposed implementation in the sandbox:
- It creates a ref **conditionally** with useRef, and it breaks a  rule of hook about using them conditionally. That's why I prefer the merging callback to check if the passed references exist or not.
- In case we'll need to use a similar approach for other components (which could happen often in a components library), we should duplicate this implementation if there are more references to merge, whether the ideal option could be to abstract this logic in a custom hook, something like this
```js
function combineRefs(...refs) {
  return function setReferences(node) {
    refs.forEach((ref) => {
      if (typeof ref === "function") ref(node);
      else if (ref !== null) ref.current = node;
    });
  };
}

export function useRefs(...refs) {
  return useCallback(combineRefs(...refs), refs);
}
```
so we'll have a utility hook that returns a memorized reference function to pass directly to the DOM element.

andresin87: take a look on actual implementation

tonyghiani: I've checked it, it's ok and should work for forwardedRef when it is not a function ref, in that case we still need to check for it's type to make it working in any possible scenario

andresin87 (approved):

ferransimon (approved):

PR #1446, opened 2021-03-29
feat(atom/input): add password right element

andresz1:
## Atom/Input
**TASK**: None

### Types of changes
Add `rightElement` prop to `Password` component

- [ ] Documentation Only
- [ ] Refactor

I need to be able to add `aria-label` to the button and a custom icon

### Screenshots - Animations

Mar-19-2021 13-54-02

### DIFF
--- a/components/atom/input/src/Password/index.js
+++ b/components/atom/input/src/Password/index.js
@@ -1,4 +1,4 @@
-import {useState} from 'react'
+import {useState, isValidElement, cloneElement} from 'react'
 import PropTypes from 'prop-types'
 import Input from '../Input'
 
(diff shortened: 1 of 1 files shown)


andresin87:
The fact is that nowadays we have a sui-password type already located in
https://github.com/SUI-Components/sui-components/blob/master/components/atom/input/src/Password/index.js
which is not adding the possibility to set the icon. Iterating it can get the same effect.

andresin87:
Try to use the ...
- leftIcon
- rightIcon
- onClickLeftIcon
- onClickRightIcon
...props of the prop of the [HoC Input icon](https://github.com/SUI-Components/sui-components/blob/master/components/atom/input/src/Input/Features/Icon/index.js) in order to re-use code (DRY)

PR #1454, opened 2021-04-01
Feat/changes input tags

a-ferrante:
### Types of changes

- [ ] Documentation Only
- [ ] Refactor

Adding 3 new props to the inputTags component

- disabled: To indicate if the field is disabled, this will apply a special class and the input inside the field will not be render
- maxTags: Number to indicate the number of maximum tags that the field can contain, if its specified when the number of tags reach its limit the component will be disabled
- placeholder: placeholder text to be displayed when there is no tags and the input is empty

### Screenshots - Animations
Captura de pantalla 2021-04-01 a las 13 11 58

### DIFF
--- a/components/molecule/inputTags/src/index.js
+++ b/components/molecule/inputTags/src/index.js
@@ -9,6 +9,7 @@ const CLASS_TAGS = `${BASE_CLASS}--withTags`
 const CLASS_TAGS_FOCUS = `${CLASS_TAGS}--focus`
 const CLASS_TAGS_ERROR = `${CLASS_TAGS}--error`
 const CLASS_TAGS_SUCCESS = `${CLASS_TAGS}--success`
+const CLASS_TAGS_DISABLED = `${CLASS_TAGS}--disabled`
 
 // eslint-disable-next-line react/prop-types
 const AtomTagItem = ({onClose = () => {}, id, ...restProps}) => {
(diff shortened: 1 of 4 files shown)


components/molecule/inputTags/src/index.scss
```diff
@@ -20,6 +20,10 @@ $sizes-atom-input-paddings: m $p-m, s $p-s !default;
     @extend %sui-atom-input-input-focus;
   }
 
+  &--disabled {
+    background-color: $c-gray-light-4;
```
nucliweb: IMO, I would define a component variable, to have the option to override via theme

```
$bg-input-tag-disable: $c-gray-light-4 !default;
```

```suggestion
    background-color: $bg-input-tag-disable;
```

nucliweb (approved):
👏🏼

components/molecule/inputTags/src/index.js
```diff
@@ -120,15 +134,26 @@ MoleculeInputTags.propTypes = {
...
 MoleculeInputTags.defaultProps = {
   size: inputSizes.MEDIUM,
   value: '',
   tags: [],
   onChangeTags: () => {},
-  onChange: () => {}
+  onChange: () => {},
+  placeholder: '',
+  disabled: false
 }
```
AitorRodriguez990, line 157: Can we remote this? `defaultProps` are not necessary because they can be added on creating the function (line 21).

AitorRodriguez990 (approved):

adriadoz (approved):

andresin87 (approved):

PR #1460, opened 2021-04-12
feat(molecule/inputTags): add a prop to avoid entering the same tag more than once

a-ferrante:
### Types of changes

- [ ] Documentation Only
- [ ] Refactor

We have the requirement to use this component in a way that we could ensure that the user is not able to enter the same value more than once.
Added a prop with default value true to avoid messing up with current implementation and add a check that if the new value is not duplicate its pushed into the tags array, if not the event its sent with the previous values
Not pretty sure about the naming on the prop/func so any input is welcome

### DIFF
--- a/components/molecule/inputTags/src/index.js
+++ b/components/molecule/inputTags/src/index.js
@@ -70,7 +70,7 @@ const MoleculeInputTags = ({
     ev.preventDefault()
     if (value) {
       const tags = [...tagsFromProps]
-      if (allowDuplicates || (!allowDuplicates && !isDuplicate(tags, value))) {
+      if (allowDuplicates || !isDuplicate(tags, value)) {
         tags.push(value)
       }
       onChangeTags(ev, {tags, name, value: ''})
--- a/components/molecule/inputTags/src/index.js
+++ b/components/molecule/inputTags/src/index.js
@@ -18,6 +18,11 @@ const AtomTagItem = ({onClose = () => {}, id, ...restProps}) => {
   return <AtomTag onClose={handleClose} {...restProps} />
 }
 
+const isDuplicate = (values, newValue) => {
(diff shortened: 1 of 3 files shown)


adriadoz (approved):
IMHO the name is okey!

components/molecule/inputTags/src/index.js
```diff
@@ -63,7 +69,10 @@ const MoleculeInputTags = ({
     const {name} = ev.target
     ev.preventDefault()
     if (value) {
-      const tags = [...tagsFromProps, value]
+      const tags = [...tagsFromProps]
+      if (allowDuplicates || (!allowDuplicates && !isDuplicate(tags, value))) {
```
tonyghiani: I think you can change this condition to 
```js
if (allowDuplicates || !isDuplicate(tags, value))) {
```
it should work anyway

a-ferrante: @tonyghiani yes, done 😄

manjarcode (approved):

andresin87:
Add the allowDuplicates to the readme plz

ferransimon (commented):
@a-ferrante What about passing a filter func prop that gets the value and tags that by default always returns true but you can override it so it filters out what you need, some cases can be duplicates but some not, I think it gives more flexibility than just a boolean prop. What do you think?

a-ferrante:
@ferransimon i think is a good idea, but i also think that not allowing entering duplicates its a common case and if by default the function provided always returns true all verticals that want to filter by duplicate will have to define the filter function to do the same thing. I think that if in the future another filter requirment come up we can iterate over this

ferransimon (approved):

andresin87 (approved):

PR #1464, opened 2021-04-15
Feat/atom tag value

a-ferrante:
## atom/tag
**TASK**: 
https://jira.scmspain.com/browse/PI-37376

### Types of changes

- [ ] Documentation Only
- [ ] Refactor

At Infojobs we have the requierment to display some values in tags. 
We need to handle this values of actionable tags to display the info correctly.

### DIFF
--- a/components/atom/tag/README.md
+++ b/components/atom/tag/README.md
@@ -113,6 +113,20 @@ $atom-tag-types: (
 );
 ```
 
+
+### Value
+
+Use the prop value if you want to use this the value on you onClick handler
+
+```js
+<AtomTag
+  label="Tag with value"
+  onClick={(_, {value}) => alert('click! this is my value', value)}
+  value={'test}
+/>
+```
+
+
 **Your high order component**
 
 ```js
--- a/components/atom/tag/src/Actionable/index.js
+++ b/components/atom/tag/src/Actionable/index.js
@@ -21,13 +21,14 @@ const ActionableTag = function({
   target,
   rel,
   linkFactory,
-  className
+  className,
+  value = null
 }) {
   return (
     <ActionableTagContainer
       className={getClassNames({className})}
       Link={linkFactory}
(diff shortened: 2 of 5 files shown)


adriadoz (approved):

components/atom/tag/README.md
```diff
@@ -113,6 +113,20 @@ $atom-tag-types: (
 );
 ```
 
+
+### Value
+
+Use the prop value if you want to use this the value on you onClick handler
```
ferransimon, line 119: "if you want to use the value on you**r** onClick"

components/atom/tag/src/Actionable/index.js
```diff
@@ -54,7 +55,8 @@ ActionableTag.propTypes = {
   onClick: PropTypes.func,
   target: PropTypes.oneOf(['_self', '_blank', '_parent', '_top']),
   rel: PropTypes.arrayOf(PropTypes.oneOf(Object.values(LINK_TYPES))),
-  linkFactory: PropTypes.func
+  linkFactory: PropTypes.func,
+  value: PropTypes.oneOfType([PropTypes.string, PropTypes.number])
```
ferransimon, line 59: That could be anything, maybe someone needs to store and object instead of just a string or number.

ferransimon (approved):

components/atom/tag/src/Actionable/index.js
a-ferrante: i see this similar to an optionList component where the value must be a string, i added number to add some other scenarios but i dont know if we should store objects as value in this type of components

manjarcode (approved):

components/atom/tag/src/Actionable/index.js
andresin87: Interesting. Maybe when its needed would not break anything and we will iterate it.

andresin87 (approved):

PR #1477, opened 2021-04-23
feat(atom/tag): disable prop in atom tag

manjarcode:
## ATOM/TAG
https://jira.scmspain.com/browse/PI-38501

### Types of changes

- [ ] Documentation Only
- [ ] Refactor

Added disabled behaviour to the actionable AtomTag, since this component can work as button, then makes sense to been able to disabled it.

### Screenshots - Animations

image

### DIFF
--- a/components/atom/tag/src/Actionable/Container.js
+++ b/components/atom/tag/src/Actionable/Container.js
@@ -9,15 +9,25 @@ const ActionableTagContainer = ({
   href,
   target,
   rel,
+  disabled,
   children,
   ...props
 }) => {
-  return href ? (
-    <Link href={href} target={target} rel={rel} {...props}>
+  return href && !disabled ? (
+    <Link
+      role="button"
+      href={href}
+      target={target}
+      rel={rel}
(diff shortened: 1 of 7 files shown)


andresin87 (approved):

ChrisDesignOps:
Its OK for now...  
...but also we have to improve Tag actionable solid colours. I created [this issue](https://github.mpi-internal.com/scmspain/frontend-ij--uilib-theme/issues/161) for doing that. 
I saw you created the disabled with an opacity of .3 (i will update this on design because i created disabled with sui variant colours and it's more ease with opacity yes!)

a-ferrante (approved):

adriadoz (approved):

manjarcode:
@ChrisDesignOps I talked with Dani about this. We agreed to try to apply it to the solid tags (which are not actionables) if it was not very difficult, but when I tried to apply I was having an issue with the hover of the disabled tags, so since this use case is not needed yet I decided to not implement it right now

PR #1513, opened 2021-05-17
feat(molecule/drawer): added fullscreen and slow animation props

ferransimon:
## MOLECULE/DRAWER
**TASK**: 

### Types of changes

- [ ] Documentation Only
- [ ] Refactor

image

We need the drawer to be able to display at fullscreen to fit the usecase of the IJ filters.

### Screenshots - Animations

Fullscreen:
image

Medium size: 

image


components/molecule/drawer/src/index.scss
```diff
@@ -90,6 +90,78 @@ $bgc-molecule-drawer-overlay-finish: rgba(0, 0, 0, 0.6) !default;
   }
 }
 
+@keyframes open-drawer-left--fullscreen {
+  0% {
+    left: -100%;
+  }
+  100% {
+    left: 0px;
```
nucliweb: We don't need to define a unit to the value zero.
Same at the rest of the code.

```suggestion
    left: 0;
```

PR #1555, opened 2021-06-08
Feature/drawer full screen2

ferransimon:
## MOLECULE/DRAWER
**TASK**: 

### Types of changes

- [ ] Documentation Only
- [x] Refactor

image

We need the drawer to be able to display at fullscreen to fit the usecase of the IJ filters.

### Screenshots - Animations

Fullscreen:
image

Medium size: 

image

### DIFF
--- a/components/molecule/drawer/src/Drawer.js
+++ b/components/molecule/drawer/src/Drawer.js
@@ -1,0 +1,66 @@
+import PropTypes from 'prop-types'
+import cx from 'classnames'
+import useEventListener from '@s-ui/react-hooks/lib/useEventListener'
+import {
+  ANIMATION_DURATION,
+  PLACEMENTS,
+  SIZES,
+  CONTENT_ELEMENT_TYPE,
+  BODY_ELEMENT_TYPE
+} from './settings'
+
+const MoleculeDrawer = ({
+  isOpen = false,
+  placement = PLACEMENTS.LEFT,
+  size = SIZES.AUTO,
+  animationDuration = ANIMATION_DURATION.FAST,
+  onClose,
+  children
+}) => {
+  useEventListener('keydown', event => {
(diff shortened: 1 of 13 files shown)


andresin87:
- [x] triggers the sui-theme animation duration tokens.
- [x] default auto size
- [x] transitionX/Y (removes keyframes)
- [x] classname anticolide declarations
- [x] compound components (overlay decouplation)
- [x] structure settings for js and scss.

andresin87 (approved):

components/molecule/drawer/src/Overlay.js
```diff
@@ -0,0 +1,39 @@
...
+      if (target !== undefined) {
+        target.current.style.position = 'relative'
+        target.current.style.overflow = 'hidden'
+      }
+    }, [target])
+    return isVisible ? (
+      <div
+        ref={forwardedRef}
+        className="react-MoleculeDrawer-overlay"
+        onClick={event => {
```
tonyghiani: Why do you need this? can onClick be something different than a function? otherwise you don't need to do this check and you can directly spread it with the other props.

components/molecule/drawer/src/Drawer.js
```diff
@@ -0,0 +1,90 @@
...
+      if (isOpenState === false) return
+      event.preventDefault()
+      if (event.key === 'Escape') {
+        setIsOpenState(false, true)
+      }
+    })
+
+    return (
+      <div
+        onTransitionEnd={event => {
```
tonyghiani: Could you extract this in a function and then pass it to prop? It's just to keep the return statement cleaner

andresin87: also done

tonyghiani (approved):

PR #1586, opened 2021-06-28
Feat/enable secondary color for buttons

nachoAdevinta:
### Types of changes

- [ ] Documentation Only
- [ ] Refactor

Some verticals use a secondary color for buttons (in addition of the accent one) and most of the times the style is overwritten to achieve a 'secondary' color style instead using the Theme Variables.

### Screenshots - Animations

We can find a quick example in the Fotocasa home page:

image

### DIFF
--- a/components/atom/button/demo/index.js
+++ b/components/atom/button/demo/index.js
@@ -64,9 +64,15 @@ const starIcon = (
 
 const atomButtonColorsIterator = atomButtonColors
   .filter(color =>
-    ['primary', 'accent', 'neutral', 'success', 'alert', 'error'].includes(
-      color
-    )
+    [
+      'primary',
+      'secondary',
+      'accent',
+      'neutral',
+      'success',
+      'alert',
+      'error'
+    ].includes(color)
   )
   .map((color, index) => [{color}, index])
 const atomButtonSocialColorsIterator = atomButtonColors
@@ -354,7 +360,7 @@ const Demo = () => {
             are solid by default for each <Code>design</Code>.
           </Paragraph>
           <Article>
-            <Grid cols={7} gutter={10}>
+            <Grid cols={8} gutter={10}>
               <Cell />
               {atomButtonColorsIterator.map(([{color}], index) => (
                 <Cell key={index} style={flexCenteredStyle}>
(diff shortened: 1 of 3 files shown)


andresin87:
atomButtonColorsIterator appears multiple times in the demo. not only in first example

andresin87:
I think the secondary usage can be the same as accent? Is it really needed?

nachoAdevinta:
> atomButtonColorsIterator appears multiple times in the demo. not only in first example

@andresin87 yep, but it only breaks grid layout in the first two grids:

image

the others are rows but cols, so no need to update cols count:

image

nachoAdevinta:
> I think the secondary usage can be the same as accent? Is it really needed?

@andresin87 that has been a topic with UX, to update accent to use it, but in addition to the primary color #303ab2 fotocasa is using 2 more:

1. #e3165b for accent (for example in fav button or some banners):

image

image

2. #1edbcf for secondary (for example in home search button and others to come in the near future):

image

oriolpuig:
Hi @nachoAdevinta,

This discussion to add a $c-secondary color to SUI is recurrent.

When I started at the company (2 years ago), I thought to add it and at the end, after discuss it with SUI team, we agreed to not do it.

The current solution that your are trying to do, is partially solved at Fotocasa theme, you can check it [here](https://github.mpi-internal.com/scmspain/frontend-fc--web-server/blob/master/theme/src/components/_components.scss#L5).

I hope to help you

nachoAdevinta:
> Hi @nachoAdevinta,
> 
> This discussion to add a $c-secondary color to SUI is recurrent.
> 
> When I started at the company (2 years ago), I thought to add it and at the end, after discuss it with SUI team, we agreed to not do it.
> 
> The current solution that your are trying to do, is partially solved at Fotocasa theme, you can check it [here](https://github.mpi-internal.com/scmspain/frontend-fc--web-server/blob/master/theme/src/components/_components.scss#L5).
> 
> I hope to help you

Hi @oriolpuig, thx!!

That is already part of another PR that I have prepared to customize the hypothetical secondary color. At the end, it's not enough because customizing the `$c-atom-button-colors` map values is indeed overwriting one of current existing ones. If there is no secondary, which one should I overwrite?

PR #1587, opened 2021-06-29
feat(components/molecule/photoUploader): add handled images prop

oriolpuig:
### Types of changes

We upload images to server on everey image uploaded on SUI component. The status of image at SUI component doesn't update from outside.

With this new prop `handledFiles`, we can pass the new SUI component status from outside.

### DIFF
--- a/components/molecule/photoUploader/src/index.js
+++ b/components/molecule/photoUploader/src/index.js
@@ -1,4 +1,4 @@
-import {useState, forwardRef} from 'react'
+import {useEffect, useState, forwardRef} from 'react'
 import {useDropzone} from 'react-dropzone'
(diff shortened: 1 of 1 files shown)


adriadoz (approved):

danivm (approved):

components/molecule/photoUploader/src/index.js
```diff
@@ -87,6 +88,10 @@ const MoleculePhotoUploader = forwardRef(
       DEFAULT_NOTIFICATION_ERROR
     )
 
+    useEffect(() => {
+      setFiles(handledFiles)
```
andresz1, line 92: Beware though, as sync props and state is usually an anti-pattern. I'd rather make it controlled/uncontrolled

PR #1605, opened 2021-07-13
Feat/scmi 60031/validation code

danivm:
## atom/validationCode
**TASK**:  [SCMI-60031](https://jira.scmspain.com/browse/SCMI-60031) 

### Types of changes

- [ ] Documentation Only
- [ ] Refactor

Create new `ValidationCode` component

### Screenshots - Animations
validationcode


components/atom/validationCode/src/index.js
```diff
@@ -0,0 +1,67 @@
+import PropTypes from 'prop-types'
+import {useState, useRef, createRef} from 'react'
+
+const DEFAULT_LENGTH = 6
+const BASE_CLASS = 'react-AtomValidationCode'
```
arnau-rius: All components have 'sui-ComponentName' as baseClass. Should we follow the convention?

components/atom/validationCode/src/index.js
```diff
@@ -0,0 +1,67 @@
...
+}) {
+  const [numbers, setNumbers] = useState(new Array(length).fill(null))
+
+  const numberRefs = useRef([])
+  numberRefs.current = numbers.map(
+    (_, i) => numberRefs.current[i] ?? createRef()
+  )
+
+  const handleChange = ({i, value}) => {
+    if (value < 10) {
```
arnau-rius: Maybe we can assign this `10` into a variable so we can give it some meaning. Is it the `MAX_NUMBER_FOR_EACH_VALUE` maybe?

components/atom/validationCode/src/index.js
```diff
@@ -0,0 +1,67 @@
...
+
+      nextInput
+        ? nextInput.current.focus()
+        : numberRefs.current[i].current.blur()
+    }
+  }
+
+  return (
+    <div className={BASE_CLASS}>
+      {label && <h3 className={`${BASE_CLASS}-label`}>{label}</h3>}
```
arnau-rius: IMHO this should not be inside the component. There is a special component `atom/label` that does this job.. Maybe we can keep it separated and let the developers compose it.

arnau-rius (approved):
Nice component 👏🏽

components/atom/validationCode/src/index.js
```diff
@@ -0,0 +1,65 @@
...
+  numberRefs.current = numbers.map(
+    (_, i) => numberRefs.current[i] ?? createRef()
+  )
+
+  const handleChange = ({i, value}) => {
+    if (value < INPUT_MAX_NUMBER) {
+      const numberArray = [...numbers]
+      numberArray[i] = value
+      setNumbers(numberArray)
+      onChange(numberArray.join(''))
```
andresin87: event handlers must return as an argument, the raw event and an object with the desired returning values:
```js
onChange(ev, {value: numberArray.join('')})
```
this means you also need to send the raw event to the declarated innerHandler. You can also wrap it into a useCallback in order not to redeclare it each time per render.

andresin87 (commented):
What do we want to do if the length is modified by prop (update lyfecycle)? It's going to work?

andresin87:
What do we do if we want to mask it (*) for pin usages? Is it needed?

danivm:
> What do we do if we want to mask it (*) for pin usages? Is it needed?

This is not in the specs:
https://www.figma.com/file/KEuHkTBkjjb1VX9D2dd1fn/Validation-Code?node-id=522%3A478

oriolpuig (approved):

oriolpuig:
🔥🔥🔥🔥🔥

nistadev (approved):
🔝 🔥

AitorRodriguez990 (approved):

PR #1610, opened 2021-07-13
feat(components/molecule/select): limit the number of posible selected items

arnau-rius:
### Types of changes

- [ ] Documentation Only
- [ ] Refactor

🆕  New prop **`maxTags`** that limits the numbers of selected items.
🔦  It follows the same API defined for the `MoleculeInputTags` component.

### Screenshots - Animations

In the next example, `maxTags` is 3.

maxtags

### DIFF
--- a/components/molecule/select/README.md
+++ b/components/molecule/select/README.md
@@ -18,21 +18,22 @@ import MoleculeSelect, {
 } from '@s-ui/react-molecule-select'
 import MoleculeSelectOption from '@s-ui/react-molecule-dropdown-option'
 
-const IconCloseTag = () => <span>x</span>  
-const IconArrowDown = () => <span>▼</span>  
+const IconCloseTag = () => <span>x</span>
+const IconArrowDown = () => <span>▼</span>
 
-const options = ['John','Paul','George','Ringo']
+const options = ['John', 'Paul', 'George', 'Ringo']
 ```
 
 ### Single Selection
 
 #### Basic usage
+
 ```js
 <MoleculeSelect
-      placeholder="Select a Country..."
-      onChange={(_, {value}) => console.log(value)}
-      iconArrowDown={<IconArrowDown />}
-    >
+  placeholder="Select a Country..."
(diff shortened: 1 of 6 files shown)


components/molecule/select/src/index.js
```diff
@@ -197,6 +197,9 @@ MoleculeSelect.propTypes = {
   /** if select accept single value or multiple values */
   multiselection: PropTypes.bool,
 
+  /** if multiselection, limit the number of selected values */
+  maxTags: PropTypes.oneOfType([PropTypes.string, PropTypes.number]),
```
arnau-rius: 💡 What do you think about this? Should we allow numbers only❓

components/molecule/select/README.md
```diff
@@ -18,21 +18,22 @@ import MoleculeSelect, {
 } from '@s-ui/react-molecule-select'
 import MoleculeSelectOption from '@s-ui/react-molecule-dropdown-option'
```
arnau-rius, line 19: Some auto-formatting ... ⬇️ 😕

components/molecule/select/src/index.js
jordevo: what could be a string for, here?

jmanrumartinez: IMO it's not needed to allow the user to pass the prop as a string since it's only a number representation

jmanrumartinez (approved):

jordevo (approved):

javiauso (approved):

tomasmax (approved):

turolopezsanabria:
Awesome thnx
as discussed in our weekly
- Can we make sure a [helpText](https://sui-components.vercel.app/workbench/atom/helpText/demo) is always used in combination with the maxTags allowed?
- I also suggest showing an alert to the user once the max number allowed has been reached (not final solution, just a proposal)

screenshot_2021-07-14_at_11 56 03

andresin87 (approved):

PR #1616, opened 2021-07-21
feat(components/atom/tooltip): functional component

andresin87:
## ATOM/TOOLTIP
**TASK**: 

### Types of changes

- [x] Documentation
- [x] Refactor

- Functional Components
- title & content --> content
- HoC removed:
  - withIntersectionObserver
  - withOpenToogle
- Hooks usage
  - [useControlledState](https://github.com/SUI-Components/adevinta-spain-components/tree/master/components/react/hooks#usecontrolledstate) (adevinta-spain-components) (@s-ui/react-hooks)
  - [useIntersection](https://github.com/streamich/react-use/blob/master/docs/useIntersection.md) (react-use)
- removes the resulting color classes cached.

### Inspiration
- ref target
  - [reactstrap 8.9.0](https://reactstrap.github.io/components/tooltips/#app)
- wrapping
  - [carbon](https://www.carbondesignsystem.com/components/tooltip/usage)
  - [chakra](https://chakra-ui.com/docs/overlay/tooltip)
  - [material-ui](https://material-ui.com/components/tooltips/#tooltip)
  - [tooltip](https://ant.design/components/tooltip/#header)


components/atom/tooltip/src/index.scss
```diff
@@ -221,6 +200,20 @@ $c-atom-tooltip-colors: (
     }
   }
 
+  &[data-placement='left-start'],
+  &[data-placement='right-start'] {
+    #{$class-inner} {
+      transform: translateY(-8px);
```
nistadev: It is better to use variables for this `translateY` value.

components/atom/tooltip/src/index.scss
```diff
@@ -221,6 +200,20 @@ $c-atom-tooltip-colors: (
...
+    }
+  }
+
+  &[data-placement='left-end'],
+  &[data-placement='right-end'] {
+    #{$class-inner} {
+      transform: translateY(8px);
```
nucliweb: Same here

components/atom/tooltip/src/index.scss
andresin87: removed – deprecated selectors

components/atom/tooltip/src/index.scss
andresin87: same (removed – deprecated selectors)

components/atom/tooltip/src/AtomTooltip.js
```diff
@@ -0,0 +1,135 @@
...
+const AtomTooltip = forwardRef(
+  (
+    {
+      children,
+      color,
+      content,
+      defaultIsVisible,
+      delay,
+      isArrowed = true,
+      isVisible,
```
nucliweb, line 34: As you were commented in the SUI Components Weekly today, we would define the default value to `true`, right?

```suggestion
      isVisible = true,
```

andresin87: default naming:
- if bolean prop --> prefix is (no with)
- semantic affirmative (visible – no hidden)

resulting attribute name: `isVisible`

the default value depends in the behaviour desired for the component: in this case `undefined`

another example:
- hidearrow --> isArrowed
- withLabel --> isLabeled (this example opens the next point -with is a reserved prefix used for HoC and use is reserved for Hooks)

nucliweb (approved):

components/atom/tooltip/src/index.scss
```diff
@@ -82,7 +88,7 @@ $c-atom-tooltip-colors: (
 }
 
 #{$base-class} {
-  display: block;
+  display: contents;
```
nucliweb, line 91: 👀  The `display: contents` don't have support for IE 11, and has partial support for Safari, Safari iOS, and Samsung Internet, take a look at the [support](https://caniuse.com/css-display-contents)

andresin87: https://github.com/SUI-Components/sui-components/blob/master/components/atom/popover/src/index.scss#L33

2 months ago (nobody has reported any issue)

andresin87: 🤷

nistadev (approved):

PR #1631, opened 2021-08-03
feat(components/atom/icon): add graylight color to colors

jmanrumartinez:
### Types of changes

- [ ] Documentation Only
- [ ] Refactor

Add grayLight color into the exported `ATOM_ICON_COLORS` constant

### DIFF
--- a/components/atom/icon/src/index.scss
+++ b/components/atom/icon/src/index.scss
@@ -16,7 +16,8 @@ $atom-icon-colors: (
   'error': $c-error,
   'primary': $c-primary,
   'success': $c-success,
-  'gray': $c-gray
+  'gray': $c-gray,
+  'grayLight': $c-gray-light-1
 );
 
 .sui-AtomIcon {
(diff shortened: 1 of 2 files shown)


components/atom/icon/src/index.scss
```diff
@@ -16,7 +16,8 @@ $atom-icon-colors: (
   'error': $c-error,
   'primary': $c-primary,
   'success': $c-success,
-  'gray': $c-gray
+  'gray': $c-gray,
+  'grayLight': $c-gray-light-1
```
jmanrumartinez, line 20: Do you think we should have a detailed key for gray light 1 variation? Like `grayLight1` instead of have `grayLight` with gray light 1 variation.

rmoralp (commented):
IMHO, even though the compiled code is small, it could be dangerous to start adding all color variations. Could you check if `currentColor` covers your needs and UX requirements?

jmanrumartinez:
> IMHO, even though the compiled code is small, it could be dangerous to start adding all color variations. Could you check if `currentColor` covers your needs and UX requirements?

That's right, I had a misconception about `currentColor`. Thanks @rmoralp for explaining it to me. I'm gonna close the PR.

PR #1660, opened 2021-09-01
feat(components/molecule/checkboxField): Add color tkn to checkboxField

fredyzz:
### Types of changes

Add color token to molecule checkboxField.
This token will allow us to keep the default color of the label and modify it when we need it, as it is required in the design of the component or the page.


midudev (approved):

turolopezsanabria (approved):
Approved, but just to have more info in case this is something to take into consideration for other themes as well

Can you add more info as of why you need this?
thnx!

PR #1661, opened 2021-09-01
feat(components/atom/switch): Add prop to disable the padding in the container

xaviermariaconejo:
### Types of changes

- New feature (non-breaking change which adds functionality)

Add prop to disable the padding in the switch atom container

### DIFF
--- a/components/atom/switch/src/SwitchType/single.js
+++ b/components/atom/switch/src/SwitchType/single.js
@@ -54,7 +54,7 @@ export const SingleSwitchTypeRender = forwardRef(
         <div
           className={cx(suitClass({element: 'container'}), {
             [suitClass({
-              element: 'container--isisDisabledPadding'
+              element: 'container--isDisabledPadding'
             })]: isDisabledPadding
           })}
           tabIndex="0"
--- a/components/atom/switch/src/SwitchType/single.js
+++ b/components/atom/switch/src/SwitchType/single.js
@@ -10,6 +10,7 @@ export const SingleSwitchTypeRender = forwardRef(
   (
     {
       disabled,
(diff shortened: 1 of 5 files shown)


adriadoz (approved):

nistadev (approved):

components/atom/switch/src/SwitchType/single.js
```diff
@@ -51,7 +52,11 @@ export const SingleSwitchTypeRender = forwardRef(
         onClick={() => onToggle()}
       >
         <div
-          className={suitClass({element: 'container'})}
+          className={cx(suitClass({element: 'container'}), {
+            [suitClass({
+              element: 'container--isisDisabledPadding'
```
arturovm2:
```suggestion
              element: 'container--isDisabledPadding'
```

arturovm2 (approved):

components/atom/switch/src/SwitchType/single.js
```diff
@@ -121,6 +126,10 @@ SingleSwitchTypeRender.propTypes = {
    * Is Input disabled?
    */
   disabled: PropTypes.bool,
+  /**
+   * The padding of the container is set to 0
+   */
+  isDisabledPadding: PropTypes.bool,
```
andresin87, line 132: In some components problem appears with the  padding and other components is margin. We are gonna tackle this like an epic and we might porpose a common boolean prop for all of them. My proposal is `isFitted` (default undefined) because there is no other component which is currently using this prop and it doesn't seems to be adapted by any other for diferent purposes

fit: caber, encajar, ajustar, encajar en, adaptar, entrar

PR #1677, opened 2021-09-09
Feature/add disable status field

tonyghiani:
### Types of changes

- [ ] Documentation Only
- [ ] Refactor

- Update the Field component to accept a `disabled` prop and render the label with a proper style.
- Allow InputField and TextField to forward the `disabled` prop to the internal field
- Refactored InputField demo


components/atom/label/src/index.scss
```diff
@@ -4,9 +4,10 @@
 $c-atom-label: inherit !default;
 $c-atom-label-optional: $c-gray-light !default;
 $c-atom-label-contrast: $c-white !default;
+$c-atom-label-disabled: inherit !default;
```
andresin87, line 7: why disabled color is inherited?

tonyghiani: I passed this as default color to not break the behaviour of those fields that maybe are already forwarding the disabled property when used, but that do not expect to change the label color. Using inherit as the $c-atom-label I want to assure no field label will get the gray color or any other until is not explicitly specified in the vertical theme.

andresin87: could you add an extra column to the [label type demo](https://sui-components-6un4no4kq-schibstedspain.vercel.app/workbench/atom/label/demo) in order to give them aligned?

jcger (approved):

andresin87 (approved):

PR #1681, opened 2021-09-10
feat(components/atom/icon): Add title prop for accessibility

stivaliserna:
### Types of changes

- [ ] Documentation Only
- [ ] Refactor

This PR is adding the `title` prop for accessibility and adds conditionally the `role="img"` and an `aria-label` when `title` is present. 

Before the `role="img"` was always rendered causing screen readers to read "Unlabelled image" for each icon. With this approach, if the title is not provided it means the icon is purely decorative and it should be hidden from the screen reader.

### DIFF
--- a/components/atom/icon/src/Icon.js
+++ b/components/atom/icon/src/Icon.js
@@ -2,9 +2,21 @@ import PropTypes from 'prop-types'
 
 const IMG_ROLE = 'img'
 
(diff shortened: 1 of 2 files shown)


beatrizip (approved):
👏

andresin87 (approved):

adriadoz (approved):

andresz1 (approved):

PR #1684, opened 2021-09-14
fix(components/molecule/select): remove select border when molecule is focused

oriolpuig:
## MOLECULE/SELECT
**TASK**: NOJIRA

### Types of changes

Remove border of inside select input when select is focused.

### 🔙 Before
image

### 🔜 After
image

### DIFF
--- a/components/molecule/select/src/index.scss
+++ b/components/molecule/select/src/index.scss
@@ -95,6 +95,11 @@ $c-molecule-select-disabled: inherit !default;
       &-container {
         @extend %sui-atom-input-select-focus;
         border: $bdw-s solid $c-primary;
+
+        .sui-AtomInput-input,
+        .sui-AtomInput--withTags {
+          border-color: transparent;
+        }
       }
     }
   }


components/molecule/select/src/index.scss
```diff
@@ -95,6 +95,11 @@ $c-molecule-select-disabled: inherit !default;
       &-container {
         @extend %sui-atom-input-select-focus;
         border: $bdw-s solid $c-primary;
+
+        .sui-AtomInput-input,
+        .sui-AtomInput--withTags {
+          border-color: transparent;
```
andresin87, line 101: border-width: 0 to remove additional borders and not only make it transparent?

arturovm2 (approved):

components/molecule/select/src/index.scss
oriolpuig: If you change the width of the border, it can cause a side effect (little movement) of the layout, isn't it?

andresin87: true

andresin87 (approved):

PR #1685, opened 2021-09-14
feat(components/molecule/select): Remove HOC withSelectUI and create …

codingwithdani:
## MOLECULE/SELECT
**TASK**: 

### Types of changes

- [ ] Documentation Only
- [x] Refactor

Remove HOC and create new component MoleculeInputSelect


components/molecule/select/src/components/MoleculeInputSelect.js
```diff
@@ -0,0 +1,50 @@
...
+
+  const classNames = cx(CLASS_ARROW, {
+    [CLASS_ARROW_DOWN]: !isOpen,
+    [CLASS_ARROW_UP]: isOpen
+  })
+
+  const extendedChildren = () =>
+    Children.toArray(children)
+      .filter(Boolean)
+      .map((child, index) => cloneElement(child, {...props}))
```
midudev: Small suggestion to avoid recreating a new props object as they should be inmutable.

```suggestion
      .map((child, index) => cloneElement(child, props))
```

midudev (approved):

components/molecule/select/src/components/MoleculeInputSelect.js
codingwithdani: If I remove the dots, we break the "select" component. 

Check this out, this is how the component looks with the spread on the props. We have some options previously selected and the text "Select a ..."
Captura de pantalla 2021-09-14 a las 12 22 17

And when we remove the spread, the component stops working and we can't select an option. 
Captura de pantalla 2021-09-14 a las 12 22 44

codingwithdani: Fixed! Thanks

andresin87 (approved):

PR #1714, opened 2021-09-29
feat(components/molecule/drawer): Drawer content vertical grow

manjarcode:
### Types of changes

- [ ] Documentation Only
- [ ] Refactor

Allow Drawer content to grow vertically in order to fill the space

### Screenshots - Animations
Before:
drawer before

After:
drawer after

### DIFF
--- a/components/molecule/drawer/src/Drawer.js
+++ b/components/molecule/drawer/src/Drawer.js
@@ -70,7 +70,7 @@ const MoleculeDrawer = forwardRef(
         )}
         ref={forwardedRef}
       >
-        <div className="react-MoleculeDrawer-content-body">{children}</div>
+        {children}
       </div>
     )
   }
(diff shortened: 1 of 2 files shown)


andresin87:
- react-MoleculeDrawer-content: is fullHeight
- react-MoleculeDrawer-content-body: is inline height

What do you need?

manjarcode:
@andresin87 I need the background to be grey, but I don't want to modify it for the whole vertical in the theme just for this case, so since I can't edit the background for the instance I need the content to fill the space in order to use the background of the child

andresin87:
create a scss token to make the content background color editable and replace that content in your vertical.

manjarcode:
@andresin87 there is a token already for the background color, but I don't want to define the default color just for this case, I'm trying to implement this:

image

andresin87:
Wait. Let me see

the `<div className="react-MoleculeDrawer-content-body">{children}</div>` wrapper is unnecesary.

add `display: block;` to the content and render directly the {children} inside the content (remove the body wrapper)

manjarcode:
@andresin87  I just applied the changes you requested!

andresin87 (approved):

PR #1767, opened 2021-10-01
fix(components/molecule/modal): fix scroll

sergiqm94:
### Types of changes

- [ ] Documentation Only
- [ ] Refactor

Force enable scroll on dismount or isOpen false

### DIFF
--- a/components/molecule/modal/src/MoleculeModal.js
+++ b/components/molecule/modal/src/MoleculeModal.js
@@ -45,6 +45,9 @@ const MoleculeModal = forwardRef(
     const wrapperRef = useRef()
     const ref = useMergeRefs(wrapperRef, forwardedRef)
 
+    const blockScrollPage = () => (document.body.style.overflow = 'hidden')
+    const enableScrollPage = () => (document.body.style.overflow = 'auto')
+
     const [isClientReady, setIsClientReady] = useState(false)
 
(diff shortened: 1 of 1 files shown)


components/molecule/modal/src/MoleculeModal.js
```diff
@@ -45,6 +45,9 @@ const MoleculeModal = forwardRef(
     const wrapperRef = useRef()
     const ref = useMergeRefs(wrapperRef, forwardedRef)
 
+    const blockScrollPage = () => (document.body.style.overflow = 'hidden')
+    const enableScrollPage = () => (document.body.style.overflow = 'auto')
+
     const [isClientReady, setIsClientReady] = useState(false)
 
     const getContainer = () => {
```
andresin87, line 53:
```js
 const getContainer = () => {
      let containerDOMEl = document.getElementById(portalContainerId)
      if (!containerDOMEl) {
        containerDOMEl = document.createElement('div')
        containerDOMEl.id = portalContainerId
        document.body.appendChild(containerDOMEl)
      }
      return containerDOMEl
    }
```
old version

it had change because b4 your last changes, the modal could be loaded in the document body or triggering and element provided.

Fix it plz

Pato2790 (approved):

andresin87 (approved):

PR #1787, opened 2021-10-07
feat(components/molecule/accordion): add custom title prop to add custom components as tab titles

quinwacca:
### Types of changes

Add custom title prop to add custom components as tab titles
We need tab titles to be h3 as a SEO strategy. 

### Screenshots - Animations
Captura de pantalla 2021-10-07 a las 12 49 58


andresin87:
make title admit custom node prop:
```js
PropTypes.oneOfType([PropTypes.string, PropTypes.node]).isRequired
```

or even better:
```js
PropTypes.node.isRequired
```

...directly which also admits string.

this prevents adoption confusion with the behavior combination of both props (title and customTitle)

andresin87 (approved):

PR #1803, opened 2021-10-08
fix(components/molecule/autosuggest): fix the disabled state

### LINKED ISSUE #1792: Autosuggest - Multiple Selection > disabled effect not working
Component [Autosuggest](https://sui-components.vercel.app/workbench/molecule/autosuggest/demo)

**Describe the bug**
Multiple Selection > disabled: component is not truly disabled as one can interact with the tags inside by removing them.

**Steps to reproduce**
1. On the Demo page go to Multiple Selection > disabled
2. Click on the X (close button) of any of the tags - this should not be clickable
2. You can click on the X(close button) on the right and everything will be removed -> this should not be clickable

**Expected behavior**
For a disabled component, it should not be possible to interact with its content. In this case, it is possible to remove the tags.

**Screenshots**

Screen Recording 2021-10-07 at 18 30 23

mslourens:
hide the close buttons of input and tag components

ISSUES CLOSED: #1792

## molecule/autosuggest
**TASK**: #1792

### Types of changes

- [ ] Documentation Only
- [ ] Refactor

Fix the disabled state of the autosuggest by passing the disabled prop to the tag input component and hiding the close button of the autosuggest. Therefore this PR also touches the atom/tag component.

**Note**: by checking out the branch, you won't see a change in the tags, because the inputTag molecule import the atomTag from `@s-ui/react-atom-tag` instead of the locally changed `../../../atom/tag/src`. When you change the import you'll see the updated atomTag as well.

### Screenshots - Animations
image

### DIFF
--- a/components/molecule/autosuggest/src/components/MultipleSelection.js
+++ b/components/molecule/autosuggest/src/components/MultipleSelection.js
@@ -1,7 +1,7 @@
 /* eslint-disable react/prop-types */
 import {useRef} from 'react'
 import MoleculeDropdownList from '@s-ui/react-molecule-dropdown-list'
-import MoleculeInputTags from '@s-ui/react-molecule-input-tags'
+import MoleculeInputTags from '../../../inputTags/src'
 import useMergeRefs from '@s-ui/react-hooks/lib/useMergeRefs'
 import isEqual from 'lodash.isequal'
 import {InputWithClearUI as MoleculeInputTagsWithClearUI} from './InputWithClearUI'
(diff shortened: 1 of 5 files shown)


components/molecule/autosuggest/src/components/MultipleSelection.js
```diff
@@ -1,7 +1,7 @@
 /* eslint-disable react/prop-types */
 import {useRef} from 'react'
 import MoleculeDropdownList from '@s-ui/react-molecule-dropdown-list'
-import MoleculeInputTags from '@s-ui/react-molecule-input-tags'
+import MoleculeInputTags from '../../../inputTags/src'
```
andresin87, line 4: this line might be restored for the CI

andresin87:
In order to successfully deploy with valid versions you need to make 3 different feat/fix commits adding:
- atom/tag
- molecule/autosuggest
- input/tags

if you do not specify it in each commit, our action will not publish the new changes on each pkg until new changes occurs in each package

mslourens:
closed in favour of #1806

PR #1819, opened 2021-10-13
fix(components/atom/label): Changed demo label and src

delalama:
Added bottom margin on label and optional text on demo

ISSUES CLOSED: 1752

### Types of changes

- [ ] Documentation Only
- [ ] Refactor

Added (Optiona) instead of *
Added 4px bottom margin to label class
Input is being imported from sui library, so maybe the requirement is to use "AtomInput" instead, anyway i tried to import the AtomInput but i have troubles with the css.

### Screenshots - Animations
Selection_005
uppercased label name

### DIFF
--- a/components/atom/label/demo/index.js
+++ b/components/atom/label/demo/index.js
@@ -46,8 +46,8 @@ const Demo = () => {
                 <AtomLabel
                   name={`atomLabelName-${key}`}
                   for={`labelName-${key}`}
-                  text={`label ${value}`}
-                  optionalText="*"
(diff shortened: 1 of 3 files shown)


andresin87:
I think this is a combinarion of #1821 and other feature. I temporaly close this issue unless @delalama  consider it is not

delalama:
> I think this is a combinarion of #1821 and other feature. I temporaly close this issue unless @delalama consider it is not

Yes, i've pushed on 1821 a PR with same code but from a logic branch, sorry

PR #1837, opened 2021-10-18
Feat/add highlight query value for custom children content

klaufel:
### Types of changes

- [ ] Documentation Only
- [ ] Refactor

- Add the custom content with the component children and higlihtQuery funcionality.

Problem when adding content in the children of the component, in this way, through the `higlihtValue` prop, it matches to apply the `<mark>` through the `higlightQuery` prop and we can also have custom content within the component.

We make the component backward compatible with the rest of functionalities.

### Screenshots - Animations

CleanShot 2021-10-18 at 13 12 08

### DIFF
--- a/components/molecule/dropdownOption/src/index.js
+++ b/components/molecule/dropdownOption/src/index.js
@@ -147,7 +147,7 @@ MoleculeDropdownOption.propTypes = {
   selected: PropTypes.bool,
   /** Text to be highlighted in the option text if found */
   highlightQuery: PropTypes.string,
-  /** Text to be highlighted in the option text if found */
+  /** Text to be display if used with highlight query with custom content */
   highlightValue: PropTypes.string,
   /* key to provoke the onClick callback. Valid any value defined here → https://www.w3.org/TR/uievents-key/#named-key-attribute-values */
   onSelectKey: PropTypes.oneOfType([PropTypes.string, PropTypes.array]),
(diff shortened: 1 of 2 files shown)


nistadev (approved):

adriadoz (approved):

oriolpuig (approved):

components/molecule/dropdownOption/src/index.js
```diff
@@ -73,6 +75,20 @@ const MoleculeDropdownOption = ({
...
+        <div className={CLASS_ITEM}>
+          <span
+            onFocus={handleInnerFocus}
+            dangerouslySetInnerHTML={{__html: mark}}
+            className={innerClassName}
+          />
+          {children}
+        </div>
+      )
+    }
```
jordevo: would this do the job for you? so that you can remove the return on L92...

I try to avoid early returns where possible...

```suggestion
      return (
        <div className={CLASS_ITEM}>
          <span
            onFocus={handleInnerFocus}
            dangerouslySetInnerHTML={{__html: mark}}
            className={innerClassName}
          />
          {highlightValue ? children : null}
        </div>
      )
```

jordevo (approved):

components/molecule/dropdownOption/src/index.js
klaufel: I have adapted the return so that it complies with the same contract and not have a wrapper so that it is currently compatible with the current use.

Agree with what you comment and in our iteration we stopped using that class to be able to do it this form, thanks

kikoruiz (approved):

PR #1841, opened 2021-10-19
Avoid installing packages on forks if NPM_TOKEN is not available

midudev:
We check if NPM_TOKEN is available to avoid GitHub Actions to try installing private packages.

### Types of changes
- [ ] Documentation Only
- [ ] Refactor

People that are creating PRs from forks are getting errors on GitHub Actions because it's missing the needed token to install private packages. As they're optional, we could safely skip this step if token is not available.

### Screenshots
CleanShot 2021-10-19 at 16 26 57@2x


package.json
```diff
@@ -5,7 +5,7 @@
   "private": true,
   "engines": {
     "node": "16",
-    "npm": "7"
+    "npm": "7 || 8"
```
andresin87, line 8: 🎉🚀

andresin87 (approved):

tomasmax (approved):

Issue #1740, opened 2021-09-30
Drawer - Allow closing it by clicking outside the drawer


turolopezsanabria:
**Describe the problem**
If there's no button in the drawer, there's no way to close it in mobile.

**Steps to reproduce**
1. Go to the [Drawer Demo](https://sui-components.vercel.app/workbench/molecule/drawer/demo)
2. Scroll down to the placements examples, enable the overlay
3. Open one of them
4. See the error (the only way to close the drawer is hitting the key ESC, and that is imposible on a mobile device)

**Suggested improvement:**
Add a prop to enable and disable closing the drawer by clicking outside the drawer (click on the overlay)

**Screenshots**

https://user-images.githubusercontent.com/23620759/135414335-f5843c4d-eb6f-4f70-b55c-cec4a9b9ef7b.mov

PR #1845, opened 2021-10-21
Fix/drawer internal state

ferransimon:
## MOLECULE/DRAWER
**TASK**: 

### Types of changes

- [ ] Documentation Only
- [ ] Refactor

When pressing ESCAPE key we were not calling the onClose callback so the outer state turnes unsync with the inner state, in the current demo is not correctly appreciated because the outer state is managed as setIsOpen(!isOpen) but you have to press the button 2 times in order to be able to open the drawer again.

### DIFF
--- a/components/molecule/drawer/src/Drawer.js
+++ b/components/molecule/drawer/src/Drawer.js
@@ -33,6 +33,7 @@ const MoleculeDrawer = forwardRef(
       event.preventDefault()
       if (event.key === 'Escape') {
         setIsOpenState(false, true)
+        onClose(event, {isOpen: false})
       }
     })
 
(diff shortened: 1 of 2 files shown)


turolopezsanabria:
great PR!
check issue #1740 please, very related to this one
thnx

components/molecule/drawer/src/Drawer.js
```diff
@@ -20,7 +20,6 @@ const MoleculeDrawer = forwardRef(
     forwardedRef
   ) => {
     const [isOpenState, setIsOpenState] = useControlledState(isOpen) // inner state
-    const [isOpenedState, setIsOpenedState] = useState(isOpen) // transition delayed state
```
andresin87, line 23: I think you have removed the transition dalayed behavior

ferransimon: How can I reproduce that? I do not see any strange behavior in the demo

turolopezsanabria (commented):
tested in safari and chrome and works perfect
Did you have the chance to address @andresin87 comments?

andresin87 (approved):

andresin87:
this issue will not solve #1740

turolopezsanabria:
> this issue will not solve #1740

ok, fine for me

PR #1848, opened 2021-10-21
fix(components/molecule/accordion): fix autoHeight borders behaviour

### LINKED ISSUE #1711: Accordion: The division lines disappear, but have double width sometimes
There are two (probably related) issues with the lines of the Accordion

1. The border disappears for a moment (safari)

https://user-images.githubusercontent.com/23620759/135114196-07b6d20c-16e4-4b4e-b3e9-594c800e81f5.mov

2. The border is duplicated for a moment (chrome)

https://user-images.githubusercontent.com/23620759/135114186-86c27d4a-8c63-4a0f-8463-551079e7eed5.mov

**Steps to reproduce**
1. Go to https://sui-components.vercel.app/workbench/molecule/accordion/demo
2. Scroll down to `"Accordion with auto height"`
4. See the defect

**Expected behaviour**
Lines should always be visible, never double

nistadev:
ISSUES CLOSED: #1711

## MOLECULE/ACCORDION
**TASK**: N/A

### Types of changes

- [ ] Documentation Only
- [ ] Refactor

This PR solves the problems described in issue #1711 with an elegant & simple solution. 

Since css transition property does not work without specifying an element height (that can be done with JS, though the requirements of the issue doesn't ask for it) we just check withTransition prop if autoHeight is not enabled.

### Screenshots - Animations
1. Border disappearing on safari (solved)
border-disappearing-safari

2. Double border on close (solved)
double-border-at-close

### DIFF
--- a/components/molecule/accordion/src/Tab/index.js
+++ b/components/molecule/accordion/src/Tab/index.js
@@ -37,7 +37,8 @@ const Tab = ({
     [OPEN_CLASS]: isOpen
   })
   const contentClassName = cx(CONTENT_CONTAINER_CLASS, {
-    [`${CONTENT_CONTAINER_CLASS}--withTransition`]: withTransition,
+    [`${CONTENT_CONTAINER_CLASS}--withTransition`]:
+      !autoHeight && withTransition,
     [`${CONTENT_CONTAINER_CLASS}--withScrollVisible`]: withScrollVisible
   })
 
--- a/components/molecule/accordion/src/Tab/index.js
+++ b/components/molecule/accordion/src/Tab/index.js
@@ -34,6 +34,8 @@ const Tab = ({
     [OPEN_CLASS]: isOpen
   })
   const containerClassName = cx(CONTAINER_BUTTON_CLASS, {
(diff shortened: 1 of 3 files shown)


turolopezsanabria:
oh, the visible / invisible gap is still happening in Safari

https://user-images.githubusercontent.com/23620759/138421730-37754ba4-83be-408a-b2c2-59183652ef94.mov

nistadev:
> oh, the visible / invisible gap is still happening in Safari

So this is not related to `autoHeight` behaviour? I will check it now, my bad assuming it was only about the `autoHeight`.

nistadev:
safari-borders

Here is my second approach. I tried different stuff but this is the only way I got borders without being glitchy. Safari renders it good even without transition delays, but other browsers (chrome, brave, etc) have 10ms glitch when rendering the css changes. I also had to implement withTransition in the title container to be able to filter autoHeight approach, and not apply any transition in that specific case. 

I am open to feedback! 👍

turolopezsanabria (approved):
nice!
I've checked all themes in Desktop and Mobile, Safari and Chrome and works much better than before
Thnx 👏🏼

andresin87 (approved):

PR #1856, opened 2021-10-27
Feature/molecule validation code optional resend

AitorRodriguez990:
### Types of changes

- [ ] Documentation Only
- [ ] Refactor

Resend button is now optional. Also, fix the delete button when the prop is not informed.

### DIFF
--- a/components/molecule/validationCode/src/validationCode.js
+++ b/components/molecule/validationCode/src/validationCode.js
@@ -89,9 +89,14 @@ const MoleculeValidationCode = forwardRef(
       <div className={baseClass}>
         <div className={`${baseClass}-header`}>
           <p className={`${baseClass}-header-labelTitle`}>{labelText}</p>
-          <AtomButton onClick={onClearHandler} design={atomButtonDesigns.FLAT}>
-            {deleteButtonTextLabel}
-          </AtomButton>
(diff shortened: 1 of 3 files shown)


components/molecule/validationCode/src/validationCode.js
```diff
@@ -132,11 +139,11 @@ const MoleculeValidationCode = forwardRef(
...
   /** text on the label */
-  labelText: PropTypes.string,
+  labelText: PropTypes.string.isRequired,
   /** text on delete button */
   deleteButtonTextLabel: PropTypes.string,
   /** text on send button  */
-  sendButtonTextLabel: PropTypes.string,
+  sendButtonTextLabel: PropTypes.string.isRequired,
   /** text on resend button */
   resendButtonTextLabel: PropTypes.string,
```
andresin87, line 148: ¿isRequired too?

AitorRodriguez990: Why not? I can do it optional too, and only works with the `onChange` method, but I'm not sure about it.

andresin87:
My opinion, The actions area might be full customizable and we might offer what it is by default. In case of any kind of customization, use the callback for giving new ReactNode.

PR #1289, opened 2020-10-30
feat(atom/input): fix readonly

andresz1:
## Atom/Input
**TASK**: None

### Types of changes

- [ ] Bug fix (non-breaking change which fixes an issue)
- [ ] New feature (non-breaking change which adds functionality)
- [x] Breaking change (fix or feature that would cause existing functionality to not work as expected)

### Description, Motivation and Context
I'd like to use the `readonly` attribute as it is not mixed with `disabled`

### Screenshots - Animations
None

PR #1875, opened 2021-11-10
Upgrade/react atom input@5

andresin87:
# ATOM/INPUT
**TASK**: 

### Types of changes

- [ ] Documentation Only
- [ ] Refactor
- [x] Component Version Upgrade

related to:

#1873 

and also 
https://github.com/SUI-Components/sui/pull/1289

### DIFF
--- a/components/atom/textarea/package.json
+++ b/components/atom/textarea/package.json
@@ -9,6 +9,7 @@
     "build:styles": "cpx './src/**/*.scss' ./lib"
   },
   "dependencies": {
+    "@s-ui/react-atom-input": "5",
     "@s-ui/react-atom-help-text": "1"
   },
   "peerDependencies": {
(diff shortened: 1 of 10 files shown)


manjarcode (approved):

components/atom/textarea/package.json
```diff
@@ -9,6 +9,7 @@
     "build:styles": "cpx './src/**/*.scss' ./lib"
   },
   "dependencies": {
+    "@s-ui/react-atom-input": "5",
```
jcger, line 12: wont't there be css collision in those projects who have already this package installed but with an old major version?

andresin87: The only changes is the disabled addition for read-only prop

andresin87: #1289

PR #1899, opened 2021-11-24
feat(components/atom/button): pass custom state to react router

AgonisticKatai:
## ATOM/BUTTON
**TASK**: [MTR-44872](https://jira.scmspain.com/browse/MTR-44872)

### Types of changes

Setting up the component to receive a routeState property that will allow us to pass a state via react routes

### Screenshots - Animations
Kapture 2021-11-24 at 10 19 16

### DIFF
--- a/components/atom/button/src/Button.js
+++ b/components/atom/button/src/Button.js
@@ -10,6 +10,7 @@ const Button = ({
   isButton, // eslint-disable-line react/prop-types
   link,
   linkFactory: Link,
+  routeState,
   forwardingRef, // eslint-disable-line react/prop-types
   ...attrs
 }) => {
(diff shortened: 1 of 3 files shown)


components/atom/button/src/Button.js
```diff
@@ -20,7 +21,14 @@ const Button = ({
...
 
   return link ? (
-    <Link {...attrs} href={href} target={target} rel={rel} ref={forwardingRef}>
+    <Link
+      {...attrs}
+      href={href}
+      target={target}
+      rel={rel}
+      ref={forwardingRef}
+      state={routeState}
```
jordevo, line 30: 🤔  can't this be accomplished already via `attrs`? I'm not sure adding a non-standard `html` attribute here (`state`) makes sense when `linkFactory` is providing a regular anchor element... what do you think?

andresin87: The routing state management it's not a responsibility of button component.

andresin87: you should better use linkFactory prop abstraction

PR #1917, opened 2021-12-01
fix(components/molecule/validationCode): avoid rendering empty delete button

AitorRodriguez990:
### Types of changes

- [ ] Documentation Only
- [ ] Refactor

Avoid rendering the Delete button when it has no text.

### Screenshots - Animations

image

### DIFF
--- a/components/molecule/validationCode/src/validationCode.js
+++ b/components/molecule/validationCode/src/validationCode.js
@@ -22,7 +22,7 @@ const MoleculeValidationCode = forwardRef(
   (
     {
       defaultValue = '',
-      deleteButtonTextLabel = '',
+      deleteButtonTextLabel,
       disabled,
       isPassword,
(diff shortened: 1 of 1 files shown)


Alverd04 (approved):

components/molecule/validationCode/src/validationCode.js
```diff
@@ -22,7 +22,7 @@ const MoleculeValidationCode = forwardRef(
   (
     {
       defaultValue = '',
-      deleteButtonTextLabel,
+      deleteButtonTextLabel = '',
```
andresin87: you dont need that defaut value 4 this

andresin87: `''` and `undefined` also are going to fe falsy

andresin87 (approved):

PR #1938, opened 2021-12-17
Fix/molecule/photouploader error when save image endpoint fails

fredyzz:
## MOLECULE/PHOTOUPLOADER
**TASK**: 

### Types of changes

- [ ] Documentation Only
- [ ] Refactor

Actually, when the user uploads a picture and the endpoint returns an error, this component doesn't show any error to the user.
With this change, the user will be noticed about the error, and the picture wouldn't be shown in the photo grid.

### Screenshots - Animations
Captura de pantalla 2021-12-17 a las 12 51 30

### DIFF
--- a/components/molecule/photoUploader/src/fileTools.js
+++ b/components/molecule/photoUploader/src/fileTools.js
@@ -110,6 +110,7 @@ export const prepareFiles = ({
   currentFiles,
   defaultFormatToBase64Options,
   errorCorruptedPhotoUploadedText,
+  errorSaveImageEndpoint,
   handlePhotosRejected,
   newFiles,
   setCorruptedFileError,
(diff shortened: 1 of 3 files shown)


oriolpuig (approved):

components/molecule/photoUploader/src/index.js
```diff
@@ -434,6 +436,9 @@ MoleculePhotoUploader.propTypes = {
   /** Text showed at error notification when some file of the initialPhotos fails */
   errorInitialPhotoDownloadErrorText: PropTypes.string.isRequired,
 
+  /** Text showed at error notification when the saveImages endpoint returns an error  */
+  errorSaveImageEndpoint: PropTypes.string.isRequired,
```
AitorRodriguez990: Are you sure you want to do this prop required? It breaks retro compatibility. I think it is not needed.

AitorRodriguez990 (approved):

components/molecule/photoUploader/src/index.js
fredyzz: Good observation Aitor! I took out the required one and added a default message to it.

nucliweb (approved):

components/molecule/photoUploader/src/fileTools.js
```diff
@@ -110,6 +110,7 @@ export const prepareFiles = ({
   currentFiles,
   defaultFormatToBase64Options,
   errorCorruptedPhotoUploadedText,
+  errorSaveImageEndpoint,
```
arnau-rius, line 113: Maybe `errorSaveImageText` it is similar to other props like: `errorCorruptedPhotoUploadedText`

arnau-rius (approved):

andresin87 (approved):

PR #1939, opened 2021-12-20
fix(components/molecule/photoUploader): Add condition to prepareFiles…

fredyzz:
### Types of changes

- [ ] Documentation Only
- [ ] Refactor

Add condition to know in prepareFiles function when callbackUploadPhoto is used.
This is necessary to know when to trigger an error in the event of a problem with the image upload.


jmanrumartinez (approved):

jmanrumartinez:
As we talked about, in MA this fix works as expected, but if you use `callBackUploadPhoto` without passing an URL, will trigger the error too, so that's not a good approach, maybe could be good to manage this kind of rule (to check if exists the URL or not) into the own component than the PhotoUploader 👍

oriolpuig (approved):

andresin87 (approved):

jmanrumartinez (approved):

quinwacca (approved):

quinwacca:
🚀

PR #1946, opened 2022-01-12
feat(components/atom/tooltip): Separate settings file and improve css output

midudev:
- [x] Add settings file to separate variables definitions.
- [x] Improve css output by removing not needed selector.

### DIFF
--- a/components/atom/tooltip/src/index.scss
+++ b/components/atom/tooltip/src/index.scss
@@ -1,5 +1,4 @@
 @import './settings.scss';
-@import '~@s-ui/theme/lib/index';
 
 #{$class-target} {
   color: $c-atom-tooltip-target;
--- a/components/atom/tooltip/src/index.scss
+++ b/components/atom/tooltip/src/index.scss
@@ -1,76 +1,5 @@
+@import './settings.scss';
 @import '~@s-ui/theme/lib/index';
-
-$fz-atom-tooltip: $fz-m !default;
-$lh-atom-tooltip: $lh-m !default;
-$wlc-atom-tooltip: 4 !default; // -webkit-line-clamp → lines to show
-
-$maw-atom-tooltip: 250px !default;
-$mah-atom-tooltip: 100px !default;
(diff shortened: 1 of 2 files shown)


nucliweb (approved):

components/atom/tooltip/src/index.scss
```diff
@@ -1,76 +1,4 @@
...
-  'alert': (
-    $c-alert,
-    $c-white
-  ),
-  'error': (
-    $c-error,
-    $c-white
-  )
-) !default;
+@import './settings.scss';
```
jcger, line 2: is it a must to include the sui-theme in the settings file? to import the theme and then the settings sounds more appropriate to me as there should be sui-theme tokens being used in the file and having it imported in the settings file makes it harder to know where those tokens come from

midudev: But you don't see any token usage of the sui-theme here, all the tokens used in the file are from the `settings`. To be honest, I see more appropiate to import the sui-theme where their tokens are being used and that's inside the `settings` file.

rmoralp (approved):

components/atom/tooltip/src/index.scss
jcger: but in those components where it is used, its not a must that its removed from there and/or only included in the settings file, right?

midudev: But here was used only because the settings where using them.

If I put the theme here and remove it from the settings, it would happen what you're mentioning in the first comment. You won't know where the tokens from the settings come from.

I'm going to add again the line here to keep importing the @s-ui/theme as it was before but that line would do nothing.

midudev: I've added the line importing again the @s-ui/theme in the `index.scss`: https://github.com/SUI-Components/sui-components/pull/1946/commits/8f84f42aba6c8827b0d283a522ead529062d9ec3

jcger (approved):

PR #1948, opened 2022-01-12
feat(components/molecule/tabs): Add a SCSS variable to be able to define the Molecule/Tabs font weight

oegea:
## [molecule]/[tabs]
**TASK**: [#MTR-49188](https://jira.scmspain.com/browse/MTR-49188)

### Types of changes

In coches.net, we need to be able to define the default font-weight for tab items.
Currently, it's only possible to define the font-weight attribute only for the active item, and non-active items are inheriting the font-weight from parent elements.

This change is placed in the context of coches.net rebranding, in which we are required to change the app typography and make some other adjustments.

### Screenshots - Animations

image

Note that the image is exposed with demo purposes only and doesn't represent the final result.


xmurcia (approved):

andresin87:
can you please move all variable tokens to a _settings.scss file and include it in the index¿?

andresin87 (approved):

oegea:
> can you please move all variable tokens to a _settings.scss file and include it in the index¿?

Sure!

jordevo (approved):

PR #1976, opened 2022-01-28
fix(components/atom/input): Fix atom input focus box shadow when has state

xaviermariaconejo:
## [ATOM]/[INPUT]
**TASK**: [SCMI-68934](https://jira.scmspain.com/browse/SCMI-68934)

### Types of changes

- Bug fix (non-breaking change which fixes an issue)

When the atom input is focused has a box shadow with the color $c-primary, when it has a state (like, error or success), the box-shadow doesn't match the same color as the border

### Screenshots - Animations

NOW/BUGG
Captura de pantalla 2022-01-28 a las 1 32 05

FIX
Captura de pantalla 2022-01-28 a las 1 31 49


andresin87 (approved):

components/atom/input/src/Input/Component/index.scss
```diff
@@ -65,6 +65,10 @@ $class-read-only: '#{$base-class}--readOnly';
   @each $state, $color in $states-atom-input {
     &.sui-AtomInput-input--#{$state} {
       border-color: $color;
+
+      &:focus {
+        box-shadow: 0 0 3px 0 $color;
```
jordevo: would it make sense to define this as a variable so that it can be overridden by themes?

xaviermariaconejo: The part of the pixels yes, but the color don't really think so, because we want that the border-color match with the box-shadow color

AitorRodriguez990: Or another approach: If a vertical needs to override this value by its team, can add this feature. It might never happen.

AitorRodriguez990 (approved):

PR #1999, opened 2022-02-02
feat(components/atom/actionButton): Add a SCSS token to customize border radius of atom/actionButton

oegea:
## atom/actionButton
**TASK**: [#MTR-50017](https://jira.scmspain.com/browse/MTR-50017)

### Types of changes

A SCSS token has been added to acionButton, to make the border-radius parameter customizable.
This change is required to comply with coches.net rebranding requirements.

**Important:** Someone has already created the token used to define action button's border radius, but the token wasn't referenced in any file of `sui-components`, seems that it wasn't being used yet. Additionally, I've changed the default value to initial, because until now actionButtons haven't had any border-radius. 

### Screenshots - Animations

**Before**

image

**After**

image

Notice that before the change, there weren't border-radius, and after we define 8px of border radius.

### DIFF
--- a/components/atom/actionButton/src/styles/index.scss
+++ b/components/atom/actionButton/src/styles/index.scss
@@ -7,6 +7,7 @@ $base-class: '.sui-AtomActionButton';
   align-items: center;
   background: none;
   border: none;
+  border-radius: $bdrs-atom-action-button;
   box-sizing: border-box;
   display: flex;
   flex-direction: column;
--- a/components/atom/actionButton/src/styles/settings.scss
+++ b/components/atom/actionButton/src/styles/settings.scss
@@ -61,7 +61,7 @@ $c-atom-action-button-colors: (
   )
 ) !default;
 
-$bdrs-atom-action-button: $bdrs-m !default;
+$bdrs-atom-action-button: initial !default;
 $bdrs-atom-action-button-icon: $bdrs-rounded !default;
 
 $ff-atom-action-button: $ff-sans-serif !default;


andresin87:
This is not adding a new scss token but also changing its current value. Are you @oegea  sure you want to change it for all verticals?

maybe it is declared but not used?

oegea:
> This is not adding a new scss token but also changing its current value. Are you @oegea sure you want to change it for all verticals?
> 
> maybe it is declared but not used?

Yep, the token already existed (has been recently added) but wasn't in use.

Note that the token has been recently added, don't know by who and when. But some days ago it didn't exist.

oegea:
Btw, let me check when it was created. I haven't seen it using git blame because the recent css split into multiple files. But I'm gonna manually check it.

oegea:
@andresin87 

First of all, apologies. The token has existed since 2020, and not a few days as I previously said. I thought it has been recently added, but after checking it again seems that I was wrong. (😢 apologies again)

However, I've been looking for references to this token in `sui-components` repo, and haven't found any usage. So, my personal opinion is that we could safely set its value to `initial` (ensuring then that it doesn't affect to other verticals), and use the same token to define the component border radius.

Additionally, I've search the token name in all organization (with the github web search), and seems also that is not in use...

turolopezsanabria:
Hello, after the catchup through Meet, I agree with you, I think it's a good idea to leave it as `initial`
Thanks!

turolopezsanabria (approved):
thnx!

andresin87 (approved):

jordevo (approved):

PR #2000, opened 2022-02-02
fix(components/atom/tooltip): Fix zindex in atom tooltip

xaviermariaconejo:
### Types of changes

-  Bug fix (non-breaking change which fixes an issue)

I've noticed that the z-index placed in the class sui-AtomTooltip, in fact, wasn't doing anything. Since the inner div was placing the position absolute, and should be this div the ones to set the z-index.

I've fixed in this PR, I've my doubts if I've resolved in the best way.

This fix a bug we have in FotocasaPro when there is other div with some z-index, so the AtomTooltip is hidden by this div, even if the z-index is lower than $z-atom-tooltip

### DIFF
--- a/components/atom/tooltip/src/styles/index.scss
+++ b/components/atom/tooltip/src/styles/index.scss
@@ -25,10 +25,8 @@ $class-target: '#{$base-class}-target';
   // Avoid flick on hover
   pointer-events: none;
 
-  position: absolute;
   // Allow breaking very long words so they don't overflow the tooltip's bounds
   word-wrap: break-word;
-  z-index: $z-atom-tooltip;
 
   // Wrapper for the tooltip content
   #{$class-inner} {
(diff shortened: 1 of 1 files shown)


andresin87 (approved):

components/atom/tooltip/src/styles/index.scss
```diff
@@ -114,6 +112,8 @@ $class-target: '#{$base-class}-target';
   }
 
   &--auto {
+    z-index: $z-atom-tooltip;
```
andresin87, line 115: i think you can remove it also

andresin87, line 115: why you have changed this rule inclusion?

andresin87: L21 #{$base-class} { // is for all the tooltip elements, which might have this z-index rule and not only the position-auto initially rendered

andresin87 (approved):

PR #2041, opened 2022-03-06
Feat molecule/tabs new tokens

jordevo:
## molecule/tabs
**TASK**: 

### Types of changes
- [ ] Documentation Only
- [ ] Refactor

It may be helpful sometimes to be able to customize the border radius on the `molecule/tabs` component.
The `border-radius` property for both the container element and the items themselves has no way of being styled via theming / overriding properties. Hence the motivation of this PR.
This PR only addresses the issue for classic (ie horizontal) layout for the tabs.
Also: unifying Sass settings files.

### Screenshots - Animations
N/A


components/molecule/tabs/src/styles/settings.scss
```diff
@@ -1 +1,33 @@
-@import '../_settings.scss';
```
andresin87: this removal is not retrocompatible cuz there are someone importing ancient file for tokens.

andresin87: Keep old token declaration file path until we make a major version

andresin87 (changes requested):

components/molecule/tabs/src/styles/settings.scss
jordevo: my bad! I didn't think it was going to break!

andresin87 (approved):

AgonisticKatai (approved):

quinwacca (approved):

PR #2059, opened 2022-03-11
feat(components/atom/panel): switch to PolymorphicElement

tonyghiani:
### Types of changes

- [ ] Documentation Only
- [ ] Refactor

In order to be able to semantically render the panel component with different HTML tags, the base element of the panel need to be replaced with the PolymorphicElement.

### DIFF
--- a/components/atom/panel/src/ColorPanel.js
+++ b/components/atom/panel/src/ColorPanel.js
@@ -21,7 +21,7 @@ const ColorPanel = function ({as = 'div', children, ...props}) {
 ColorPanel.displayName = 'ColorPanel'
 
 ColorPanel.propTypes = {
-  ...PolymorphicElement.propTypes,
+  as: PropTypes.elementType,
   children: PropTypes.node,
   color: PropTypes.oneOf(Object.values(COLORS)),
   alpha: PropTypes.oneOf(Object.values(ALPHA)),
--- a/components/atom/panel/src/ColorPanel.js
+++ b/components/atom/panel/src/ColorPanel.js
@@ -1,5 +1,7 @@
 import PropTypes from 'prop-types'
 
+import PolymorphicElement from '@s-ui/react-atom-polymorphic-element'
+
 import {
   ALPHA,
   COLORS,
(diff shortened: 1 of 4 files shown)


manjarcode (approved):

javiauso (approved):

components/atom/panel/src/ColorPanel.js
```diff
@@ -8,13 +10,18 @@ import {
...
+    <PolymorphicElement as={as} className={getClassNames(props)}>
+      {children}
+    </PolymorphicElement>
+  )
 }
 
 ColorPanel.displayName = 'ColorPanel'
 
 ColorPanel.propTypes = {
+  ...PolymorphicElement.propTypes,
```
andresin87: do not 🙅‍♀️🙅‍♂️ because react-docgen does not make the docu with this meth

components/atom/panel/src/ImagePanel.js
```diff
@@ -44,17 +46,22 @@ const getStyles = function ({src}) {
...
       {children}
-    </div>
+    </PolymorphicElement>
   )
 }
 
 ImagePanel.displayName = 'ImagePanel'
 
 ImagePanel.propTypes = {
+  ...PolymorphicElement.propTypes,
```
andresin87: same

andresin87 (changes requested):
some changes

andresin87 (approved):

andresin87 (approved):
Looks like ImagePanel and ColorPanel are AtomPanel resulting component, and this is not providing the 'as' prop method documented to them. Add the 'as' prop to the AtomPanel and document it also 🙏.

andresin87 (changes requested):
Looks like ImagePanel and ColorPanel are AtomPanel resulting component, and this is not providing the 'as' prop method documented to them. Add the 'as' prop to the AtomPanel and document it also 🙏.

tonyghiani:
> Looks like ImagePanel and ColorPanel are AtomPanel resulting component, and this is not providing the 'as' prop method documented to them. Add the 'as' prop to the AtomPanel and document it also 🙏.

@andresin87  I'll definitely add this to the documentation! About passing this prop inside the AtomPanel, properties are already spread and the `as` property will be passed to the other two, would you prefer to explicitly restructure the as property and set it in both possible scenarios in the index.js file?

andresin87:
> > Looks like ImagePanel and ColorPanel are AtomPanel resulting component, and this is not providing the 'as' prop method documented to them. Add the 'as' prop to the AtomPanel and document it also 🙏.
> 
> @andresin87 I'll definitely add this to the documentation! About passing this prop inside the AtomPanel, properties are already spread and the `as` property will be passed to the other two, would you prefer to explicitly restructure the as property and set it in both possible scenarios in the index.js file?

declare it in the propTypes if you consider it must be documented.

andresin87 (approved):

PR #2087, opened 2022-03-24
feat(components/atom/switch): a11y n performance

andresin87:
## ATOM/SWITCH
**TASK**: 

### Types of changes

- [ ] Documentation Only
- [x] Refactor
- [x] A11y
- [x] Performance
- [x] Naming 

- Current component does not have a valid role. This PR adds the "switch" role to the component (a11y)
- The focus (css only) behavior was made using classes (react (click n focus mixed) states  strategy) generating a lot of re-renders. It has been refactored using a combination of focus and focus-visible pseudo-class.
- The naming for the selected (checked) status are called active. This is not right cuz this word is not correct. You can take a look in : [mdn:active](https://developer.mozilla.org/en-US/docs/Web/CSS/:active)
- Remove a lot on JS code lines
- Fix a bug when keeping pressed the tab key just using the browser's native focus behavior
- checked class --> aria-checked attribute and css selector
- disabled class --> disable attribute and css selector

##### 🚨 FOCUS, ACTIVE, HOVER:
- [ ] `JS react classes 🙅‍♂️🙅‍♀️`
- [x] `CSS pseudo-clases ❤️`

--> [WTF is :focus-visible? 👀](https://css-tricks.com/almanac/selectors/f/focus-visible/)

### Screenshots - Animations

[check here to see the differences between focus-hover-active-and-focus-visible](https://codesandbox.io/s/focus-focus-visible-hover-active-checkbox-qdrcyr?file=/src/sass/layout/checkbox.scss)

### DIFF
--- a/components/atom/switch/src/styles/index.scss
+++ b/components/atom/switch/src/styles/index.scss
@@ -13,11 +13,11 @@ $base-class: '.sui-AtomSwitch';
 
   &-container {
     align-items: center;
+    box-shadow: none;
     display: flex;
     height: $h-atom-switch-container-default;
     justify-content: center;
     outline: none;
-    box-shadow: none;
 
     &--fullWidth {
       justify-content: space-between;
--- a/components/atom/switch/src/styles/index.scss
+++ b/components/atom/switch/src/styles/index.scss
@@ -13,39 +13,33 @@ $base-class: '.sui-AtomSwitch';
 
   &-container {
     align-items: center;
-    box-shadow: none;
     display: flex;
     height: $h-atom-switch-container-default;
     justify-content: center;
     outline: none;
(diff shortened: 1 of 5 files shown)


components/atom/switch/src/styles/index.scss
```diff
@@ -13,39 +13,33 @@ $base-class: '.sui-AtomSwitch';
 
   &-container {
     align-items: center;
-    box-shadow: none;
     display: flex;
     height: $h-atom-switch-container-default;
     justify-content: center;
     outline: none;
+    box-shadow: none;
```
nucliweb: 🤔  We use alphabetic order, right?

nucliweb (approved):

PR #2089, opened 2022-03-25
feat(components/molecule/accordion): new prop forceClose

jordevo:
Allow for closing of all opened tabs on a molecule/accordion component

## MOLECULE/ACCORDION
**TASK**: N/A

### Types of changes

- [ ] Documentation Only
- [ ] Refactor

### DIFF
--- a/components/molecule/accordion/src/index.js
+++ b/components/molecule/accordion/src/index.js
@@ -1,4 +1,4 @@
-import {useState} from 'react'
+import {useEffect, useState} from 'react'
 import PropTypes from 'prop-types'
 
 import Tab from './Tab/index.js'
@@ -7,6 +7,7 @@ import {BASE_CLASS} from './settings.js'
 const MoleculeAccordion = ({
   children,
   defaultOpenedTabs = [],
+  forceClose = false,
   onToggleTab = () => {},
   withAutoClose,
   ...tabProps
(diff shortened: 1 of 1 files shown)


components/molecule/accordion/src/index.js
```diff
@@ -57,6 +62,7 @@ MoleculeAccordion.propTypes = {
    * Array with tab indexes to be opened by default
    */
   defaultOpenedTabs: PropTypes.arrayOf(PropTypes.Number),
+  forceClose: PropTypes.bool,
```
quinwacca, line 65: Maybe you can add some doc
```suggestion
  /**
   * To set the status open/closed from outside the component
   */
  forceClose: PropTypes.bool,
```

quinwacca (approved):

components/molecule/accordion/src/index.js
andresin87: as long as i know, it'll be a problem because if yo start navigating and use the forceClose boolean prop (true), then yo continue navigating and if you want to force it again you need to change twice (false, and true) to re-forceClose all tabs.

The right strategy might be to be a controlled-uncontrolled component combining `defaultOpenedTabs` and`openedTabs` also. I'll take it.

andresin87 (changes requested):
As long as i know, it'll be a problem because if you start navigating and use the forceClose boolean prop (true), then you continue navigating and if you want to force it again you need to change twice (false, and true) to re-forceClose all tabs. No demo, neither tests. It's so much hard to test the purposed behavior.

The right strategy might be to be a controlled-uncontrolled component combining `defaultOpenedTabs` and`openedTabs` also.

I'll take it.

jordevo:
No longer relevant, a new approach has been taken.

PR #2096, opened 2022-03-31
Fix/atom switch prevent form submit

oriolpuig:
## ATOM/SWITCH
**TASK**: NO_TICKET

### Types of changes

Allow fire switch click after label is clicked in Single Atom Switch

### DIFF
--- a/components/atom/switch/src/SwitchType/single.js
+++ b/components/atom/switch/src/SwitchType/single.js
@@ -68,6 +68,7 @@ export const SingleSwitchTypeRender = forwardRef(
             />
           )}
           <button
+            type="button"
             className={cx(suitClass({element: 'inputContainer'}), {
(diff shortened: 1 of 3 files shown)


components/atom/switch/src/SwitchType/single.js
```diff
@@ -81,6 +80,7 @@ export const SingleSwitchTypeRender = forwardRef(
             aria-disabled={disabled}
             disabled={disabled}
             {...(!disabled && {tabIndex: 0})}
+            {...(name && {id: name})}
```
andresin87, line 83: `id={name}` is equal

andresin87:
```suggestion
            id={name}
```

andresin87 (approved):

PR #2160, opened 2022-05-24
feat(components/atom/input): use polymorphic input

andresz1:
### Types of changes

- [ ] 🧾 Documentation
- [ ] 📷 Demo
- [ ] 🧪 Test
- [ ] 🧠 Refactor
- [ ] 💄 Styles

This PR makes input work wit `as` prop. The main reason is to use it as a `select`. The look and feel should be the same but it works as a native select. We are planning to release a `NativeSelect` component that uses the input (we will discuss this tomorrow cc @andresin87 @turolopezsanabria).

**How can it be used ?**

```
<Input as="select">
  <option value="1">1</option>
  <option value="2">2</option>
  <option value="3">3</option>
</Input>
```

### DIFF
--- a/components/atom/input/src/Input/Component/index.js
+++ b/components/atom/input/src/Input/Component/index.js
@@ -111,7 +111,7 @@ const Input = forwardRef(
         required={required}
         pattern={pattern}
         inputMode={inputMode}
-        children={children}
+        children={as === 'input' ? undefined : children}
       />
     )
   }
--- a/components/atom/input/src/Input/Component/index.js
+++ b/components/atom/input/src/Input/Component/index.js
@@ -1,6 +1,7 @@
 import {forwardRef} from 'react'
 import PropTypes from 'prop-types'
 
+import PolymorphicElement from '@s-ui/react-primitive-polymorphic-element'
 import useMergeRefs from '@s-ui/react-hooks/lib/useMergeRefs'
 
 import {
@@ -14,6 +15,7 @@ import {
 const Input = forwardRef(
   (
     {
+      as = 'input',
       disabled,
       readOnly,
       hideInput,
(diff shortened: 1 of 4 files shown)


jmanrumartinez (approved):

javiauso (approved):

jordevo (approved):

kikoruiz (approved):

tomasmax (approved):

components/atom/input/src/Input/index.js
```diff
@@ -37,7 +37,12 @@ const BaseInput = forwardRef(
             onClickLeftIcon={onClickLeftIcon}
             onClickRightIcon={onClickRightIcon}
           >
-            <Input ref={forwardedRef} {...inputProps} size={size} />
+            <Input
+              ref={forwardedRef}
+              children={children}
+              {...inputProps}
+              size={size}
+            />
```
andresin87: better use the clasical way
```js
<Input
  ref={forwardedRef}
  {...inputProps}
  size={size}
>
  {children}
</Input>
```

andresin87 (changes requested):

components/atom/input/src/Input/Component/index.js
```diff
@@ -107,12 +111,15 @@ const Input = forwardRef(
         required={required}
         pattern={pattern}
         inputMode={inputMode}
+        children={as === 'input' ? undefined : children}
```
andresin87, line 114: let children always become added but normally not used to simplify the logic and add it not as a prop but as a nested element

andresin87 (changes requested):

components/atom/input/src/Input/Component/index.js
andresz1: I needed to do it this way because there was an error related to the input (it can not have children)

andresin87 (approved):

PR #2171, opened 2022-05-24
Feature/atom spinner add negative prop

quinwacca:
### Types of changes
Add a negative prop, to have an alternate overlay color.

- [x] 💄 Styles

In motor we have the need to have an alternate overlay background color. We currently have a light one, but in some places a dark one fits better. By theme, currently we only can have one background color defined.
With this PR we can have two background colors defined.

### Screenshots 
Current: image
Negative: image


components/atom/spinner/src/styles/settings.scss
```diff
@@ -1,4 +1,5 @@
 $bgc-atom-spinner: rgba($c-white, 0.6) !default;
+$bgc-atom-spinner-negative: rgba($c-black, 0.6) !default;
```
jordevo, line 2: how about using CSS' own filter function taking `$bgc-atom-spinner` as input?

PR #2176, opened 2022-05-26
feat(components/atom/spinner): new version of atom spinner

quinwacca:
### Types of changes
New version of atom spinner!

### BREAKING CHANGES:

Add different overlay types, replacing old `noBackground` boolean prop.
Add children with injected props from `AtomSpinner`, allowing full customization of the component.
Removed `noBackground` prop. You can use now prop `overlayType: transparent`

- [x] 💥 Breaking change (fix or feature that would cause existing functionality to not work as expected)
- [x] 🧾 Documentation
- [x] 📷 Demo
- [x] 🧪 Test
- [x] 🧠 Refactor
- [x] 💄 Styles

By theme, only one overlay color could be set. We have the necessity to have two different overlays in motor at the same time. Now we can select the desired overlay by prop, also a transparent one, and we can define these colors on our theme, of course.

image

Also, `AtomSpinner` now can receive a `children` prop that will be injected with `AtomSpinner` props and the `AtomSpinner`'s loader, to fully customize our spinner.

May-26-2022 14-50-35

### DIFF
--- a/components/atom/spinner/src/DefaultSpinner.js
+++ b/components/atom/spinner/src/DefaultSpinner.js
@@ -1,0 +1,15 @@
+import PropTypes from 'prop-types'
+import {forwardRef} from 'react'
+
+const DefaultSpinner = forwardRef(({delayed, loader}, forwardedRef) => {
+  return <span ref={forwardedRef}>{!delayed ? loader : <noscript />}</span>
+})
+
+DefaultSpinner.propTypes = {
+  delayed: PropTypes.bool,
+  loader: PropTypes.object
+}
+
+DefaultSpinner.displayName = 'DefaultSpinner'
+
+export default DefaultSpinner
--- a/components/atom/spinner/src/DefaultSpinner.js
+++ b/components/atom/spinner/src/DefaultSpinner.js
@@ -1,12 +1,12 @@
 import PropTypes from 'prop-types'
 import {forwardRef} from 'react'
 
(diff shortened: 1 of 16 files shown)


andresin87 (approved):

jordevo (approved):

components/atom/spinner/src/DefaultSpinner.js
```diff
@@ -0,0 +1,15 @@
+import PropTypes from 'prop-types'
+import {forwardRef} from 'react'
+
+const DefaultSpinner = forwardRef(({delayed, loader}, forwardedRef) => {
+  return <span ref={forwardedRef}>{!delayed ? loader : <noscript />}</span>
+})
+
+DefaultSpinner.propTypes = {
+  delayed: PropTypes.bool,
```
jordevo: since we're releasing a major... 

```suggestion
  isDelayed: PropTypes.bool,
```

what do you think?

andresin87: boolean props might have the prefix:
✅  `is[whatever]`
✅  `has[whatever]`

❌  `with[whatever]` because in react, with prefix is used for HoC
❌  `use[whatever]` because in react, use prefix is used for hooks

quinwacca: Good idea! Thanks!

PR #2219, opened 2022-06-17
Fix/export colors dictionary correctly

codingwithdani:
## Category/Component

**TASK**: 

### Types of changes

- [ ] 🧾 Documentation
- [x] 📷 Demo
- [ ] 🧪 Test
- [ ] 🧠 Refactor
- [ ] 💄 Styles

On the Button component, the array "atomButtonColors" it is not working correctly and in no vertical is it being used. 

It's only implemented in some cases and it's working because it's taking the default color "primary":
https://github.mpi-internal.com/search?q=atomButtonColors&type=code

To fix this issue I changed the array "atomButtonColors" for an Object.

We are going to be able to use the atomButtonColors variable on all verticals and remove magic strings.

### DIFF
--- a/components/atom/button/test/index.test.js
+++ b/components/atom/button/test/index.test.js
@@ -7,12 +7,12 @@ import ReactDOM from 'react-dom'
 import chai, {expect} from 'chai'
 import chaiDOM from 'chai-dom'
 import sinon from 'sinon'
+
 import {fireEvent} from '@testing-library/react'
 
-import * as pkg from '../src/index.js'
-import {createClasses} from '../src/config.js'
-
 import json from '../package.json'
+import {createClasses} from '../src/config.js'
+import * as pkg from '../src/index.js'
 
 chai.use(chaiDOM)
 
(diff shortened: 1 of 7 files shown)


quinwacca (approved):

andresin87:
🚨 This is a breaking change. Can we discuss why we need to make this change or maybe e can find a soft approach

components/atom/button/test/index.test.js
```diff
@@ -542,15 +542,16 @@ describe(json.name, () => {
   })
 
   describe('atomButtonColors', () => {
```
andresin87, line 544: check the other exported object tests and copy them

codingwithdani: solved

components/atom/button/test/index.test.js
```diff
@@ -542,41 +542,45 @@ describe(json.name, () => {
...
-    it('value must be an array', () => {
+    it('value must be a dictionary', () => {
```
andresin87:
```suggestion
    it('value must be an object enum', () => {
```

components/atom/button/test/index.test.js
```diff
@@ -542,41 +542,45 @@ describe(json.name, () => {
   })
 
   describe('atomButtonColors', () => {
-    it('value must be an array', () => {
+    it('value must be an object enum, () => {
```
andresin87:
```suggestion
    it('value must be an object enum', () => {
```

andresin87 (approved):

PR #2222, opened 2022-06-21
feat(components/molecule/tabs): Allow to disable scroll into view

oriolpuig:
### Types of changes

- [ ] 🧾 Documentation
- [ ] 📷 Demo
- [ ] 🧪 Test
- [ ] 🧠 Refactor
- [ ] 💄 Styles

The current MoleculeTabs scrollToView is not working fine. When tabs scroll into view, all container is moving rarely and is not working as expected. 

As discussed with @andresin87, we will add a prop `autoScrollIntoView` (true by default to maintain the compatibility) to disable this funcionality in case you want, and in the near future we will improve this funcionality to fix the bug correctly. Probably the fix will cause a new major version.

### DIFF
--- a/components/molecule/tabs/src/components/MoleculeTab.js
+++ b/components/molecule/tabs/src/components/MoleculeTab.js
@@ -1,19 +1,31 @@
-import {useEffect, useRef, forwardRef} from 'react'
+import {forwardRef, useEffect, useRef} from 'react'
+
 import cx from 'classnames'
 import PropTypes from 'prop-types'
+
 import useMergeRefs from '@s-ui/react-hooks/lib/useMergeRefs'
 
 import {
   CLASS_TAB,
-  CLASS_TAB_ICON,
-  CLASS_TAB_COUNT,
   CLASS_TAB_ACTIVE,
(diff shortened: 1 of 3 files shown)


components/molecule/tabs/src/components/MoleculeTab.js
```diff
@@ -1,19 +1,31 @@
...
+    {
+      autoScrollIntoView = true,
+      active,
+      onChange,
+      disabled,
+      icon,
+      count,
+      label,
+      numTab,
+      isIntersecting
```
jordevo:
```suggestion
      active,
      autoScrollIntoView = true,
      count,
      disabled,
      icon,
      isIntersecting,
      label,
      numTab,
      onChange
```

sorting props in ascending order always helps 😉

jordevo (approved):

components/molecule/tabs/src/components/MoleculeTab.js
oriolpuig: Totally agree, thanks!

andresin87 (approved):

PR #2240, opened 2022-07-13
Feature/cd 55842/restyling readonly and disabled sui input

IsabelGomez87:
## Category/Component

`🔍 Show`

**TASK**: https://jira.scmspain.com/browse/MTR-55842

### Types of changes

- [ ] 🧾 Documentation
- [ ] 📷 Demo
- [ ] 🧪 Test
- [ ] 🧠 Refactor
- [ ] 💄 Styles

Styles have been added to the readonly class to be able to change them in the theme of each vertical

### DIFF
--- a/components/atom/input/src/Input/Component/index.scss
+++ b/components/atom/input/src/Input/Component/index.scss
@@ -18,6 +18,12 @@ $class-read-only: '#{$base-class-input}--readOnly';
     display: none;
   }
 
+  &--readOnly {
+    background: $bgc-atom-input-read-only;
+    border: $bd-atom-input-read-only;
+    color: $c-atom-input-read-only;
+  }
+
   @include atom-input-shape-wrapper(
     $base-class,
     $base-class-input,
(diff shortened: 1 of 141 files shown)


components/atom/input/src/Input/Component/index.scss
```diff
@@ -18,6 +18,12 @@ $class-read-only: '#{$base-class-input}--readOnly';
     display: none;
   }
 
+  &--readOnly {
```
andresin87, line 21: always use native selector over classes
https://www.w3schools.com/cssref/sel_read-only.asp

andresin87: I know its not [yours](https://github.com/SUI-Components/sui-components/commit/a1d6a0e39979bd8e958fd6afea2f80636c2b2e75) but can you remove that class and handle that using the `:read-only` pseudo-selector?

turolopezsanabria: +1 to this

IsabelGomez87: But the `:read-only` pseudo-selector is not supported in IE11. We could control it with the HTML attribute or with the class, as it is now (in fact, we have used it because it was already created [here](https://github.com/SUI-Components/sui-components/blob/master/components/atom/input/src/Input/Component/index.scss#L2))

andresin87 (approved):

PR #2279, opened 2022-08-16
feat(components/molecule/tabs): make tabs accesibles

oriolpuig:
### Types of changes

- [ ] 🧾 Documentation
- [ ] 📷 Demo
- [x] 🧪 Test
- [ ] 🧠 Refactor
- [ ] 💄 Styles

Currently our component is not Accessibility friendly, with this PR we want to make it possible. 
We use the specific role for each element to improve User Experience and our Test Experience with Testing Library.

#### 📃 DOCS: https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Roles/tab_role

### DIFF
--- a/components/molecule/tabs/src/components/MoleculeTabs.js
+++ b/components/molecule/tabs/src/components/MoleculeTabs.js
@@ -42,20 +42,29 @@ const MoleculeTabs = ({
 
   const activeTabContent = childrenArray.reduce((activeContent, child) => {
     if (child) {
-      const {children: childrenChild, active} = child.props
-      return active ? childrenChild : activeContent
+      const {children: childrenChild, active, label, numTab} = child.props
+      return active ? (
+        <div
+          aria-labelledby={label}
+          className={CLASS_CONTENT}
+          id={`molecule-tab-content-${numTab}`}
+          role="tabpanel"
+        >
+          {childrenChild}
+        </div>
+      ) : (
+        <div className={CLASS_CONTENT}>{activeContent}</div>
+      )
     }
-    return activeContent
+    return <div className={CLASS_CONTENT}>{activeContent}</div>
   }, null)
 
   return (
     <div className={className}>
-      <ul ref={outerRef} className={CLASS_SCROLLER}>
+      <ul ref={outerRef} className={CLASS_SCROLLER} role="tablist">
         {extendedChildren}
       </ul>
-      {activeTabContent ? (
-        <div className={CLASS_CONTENT}>{activeTabContent}</div>
-      ) : null}
+      {activeTabContent || null}
     </div>
   )
 }
--- a/components/molecule/tabs/src/components/MoleculeTabs.js
+++ b/components/molecule/tabs/src/components/MoleculeTabs.js
@@ -47,13 +47,13 @@ const MoleculeTabs = ({
     if (child) {
       const {children: childrenChild, active, numTab} = child.props
 
-      return active ? (
-        <div className={CLASS_CONTENT} id={`${id}-${numTab}`} role="tabpanel">
-          {childrenChild}
-        </div>
-      ) : (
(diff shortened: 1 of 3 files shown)


andresz1 (approved):

andresz1:
Nice PR 👏🏻👏🏻👏🏻 Should we also add [aria-orientation](https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Attributes/aria-orientation) when the `type` is set to `vertical` ? https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Roles/tablist_role

components/molecule/tabs/src/components/MoleculeTabs.js
```diff
@@ -42,20 +42,28 @@ const MoleculeTabs = ({
 
   const activeTabContent = childrenArray.reduce((activeContent, child) => {
     if (child) {
-      const {children: childrenChild, active} = child.props
-      return active ? childrenChild : activeContent
+      const {children: childrenChild, active, numTab} = child.props
+      return active ? (
+        <div
+          className={CLASS_CONTENT}
+          id={`molecule-tab-content-${numTab}`}
```
andresz1: If we add two tabs in the same page. Is this id going to be repeated ?

oriolpuig: For sure, yes. Good catch. 
I've added this to help the `aria-controls` on tab elements 🤔. 

In order to do it unique, we must pass a `id` to the MoleculeTab and use it to build it dynamically, isn't it? What do you think?

andresin87: Pass an id to `MoleculeTabs`, prop drill it to the MoleculeTab and concat the numtab to get the id of each.

components/molecule/tabs/src/components/MoleculeTabs.js
```diff
@@ -36,26 +38,39 @@ const MoleculeTabs = ({
...
+        aria-orientation={
+          isVerticalOrientation ? TYPES.VERTICAL : TYPES.HORIZONTAL
+        }
+      >
         {extendedChildren}
       </ul>
-      {activeTabContent ? (
-        <div className={CLASS_CONTENT}>{activeTabContent}</div>
-      ) : null}
+      {activeTabContent || null}
```
andresin87:
```suggestion
      {activeTabContent}
```

components/molecule/tabs/src/components/MoleculeTabs.js
```diff
@@ -36,26 +38,39 @@ const MoleculeTabs = ({
...
-      return active ? childrenChild : activeContent
+      const {children: childrenChild, active, numTab} = child.props
+
+      return active ? (
+        <div className={CLASS_CONTENT} id={`${id}-${numTab}`} role="tabpanel">
+          {childrenChild}
+        </div>
+      ) : (
+        activeContent
+      )
```
andresin87:
```js
return (<div className={CLASS_CONTENT} id={`${id}-${numTab}`} role="tabpanel">{active ? childrenChild : activeContent}</div>)
```
retrocompatibility

andresin87: review plz, maybe its wrong

oriolpuig: Nope, it's not well. Your solution add the active tab content inside another active tab content

image

andresin87:
```js
if(active) {
  return (
    <div className={CLASS_CONTENT} id={`${id}-${numTab}`} role="tabpanel">
      {childrenChild}
    </div>
  )
}
```

andresin87 (approved):

PR #2287, opened 2022-08-22
Feat/image editor improvement

oegea:
#### `❓ Ask`

**TASK**: [MTR-56617](https://jira.scmspain.com/browse/MTR-56617)

### Types of changes

**Motivation**
In coches.net pro we are working to implement a page from which users can edit their account details. One of these details is the user profile image.

Following the UX specifications, we need a way to not only select a new user image, but also crop and edit that image before submitting it. We've selected `molecule/imageEditor` to achieve this goal.

However, every time an image is edited on `molecule/imageEditor` a blob url is returned by default, and it would be easier for us to directly receive the blob object instead of a url, in order to be able to submit the result as part of a `FormData` object. 

**Technical observations**
Technically, it should be possible to reverse the url-generation process by performing an http request to retrieve the blob url content, and initialize a blob object with that content, however we think that this approach adds too much complexity to our domain, and doesn't seem to be a performance-oriented approach (internally, the blob object is duplicated, first in the `imageEditor` component, and later from our side).

**Changes**
This PR adds the possibility of directly retrieving a blob object when using the `molecule/imageEditor` component.

### DIFF
--- a/components/molecule/imageEditor/README.md
+++ b/components/molecule/imageEditor/README.md
@@ -25,7 +25,6 @@ return (
     onChange={handleChange}
     rotateLabelIcon={<RotateIcon>}
     rotateLabelText="Rotate"
-    returnBlobObject={true}
   />
 )
 ```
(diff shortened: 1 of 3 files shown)


andresin87:
You do not need a prop to do that. just return it by default. its retrocompatible

andresin87 (approved):

PR #2290, opened 2022-08-22
feat(components/molecule/imageEditor): Add a preview mode to molecule/imageEditor to optimize image

oegea:
### Types of changes

ImageEditor is currently calculating a new image every time the user makes zoom or rotates an image. Which, with a normal use, could happen between 50 and 100 times just by sliding the zoom or rotation slide.

This is causing a bad performance. Additionally, as we don't know which is the final calculated image, we're not sure when the  `save` button of our web-app can be enabled to finish the process.

To solve this problematic, this proposal debounces the execution of some critical functions, to ensure that image is calculate just when needed.

Co-authored with @andresin87 

### Screenshots - Animations

There aren't visual changes.

### DIFF
--- a/components/molecule/imageEditor/src/index.js
+++ b/components/molecule/imageEditor/src/index.js
@@ -1,4 +1,4 @@
-import {useCallback, useState} from 'react'
+import {useCallback, useEffect, useRef, useState} from 'react'
 import Cropper from 'react-easy-crop'
 
 import PropTypes from 'prop-types'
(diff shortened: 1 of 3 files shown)


components/molecule/imageEditor/src/index.js
```diff
@@ -99,6 +113,7 @@ const MoleculeImageEditor = ({
 MoleculeImageEditor.displayName = 'MoleculeImageEditor'
 MoleculeImageEditor.propTypes = {
   aspect: PropTypes.number,
+  previewOnlyMode: PropTypes.bool,
```
jordevo:
```suggestion
  isPreviewMode: PropTypes.bool,
```

or even

```suggestion
  isPreview: PropTypes.bool,
```

given they're boolean props. What do you think...?

jordevo (approved):
good stuff 👏 

as per our conversation, though, perhaps it's a bit too hacky and we may find a good use case for a `debounce` here...

AgonisticKatai (approved):

andresin87 (approved):

PR #2303, opened 2022-09-05
feat(components/molecule/imageEditor): Add debouncingTime as a customizable property

oegea:
### Types of changes

It would be great for us in CNET PRO, to have the possibility of customizing the debouncingTime of  `imageEditor`.

Thanks to the new extensibility capabilities, we have been able to implement an approach to improve the component performance without requiring modifications on the `sui` side.  Basically, this new approach avoids calculating the image every time the user performs a modification on the crop or rotation parameters, and exposes a function to calculate it once, only when the edition is finished, using the `useImperativeHandle` hook.

However, as the interface runs smoothly now, debouncing is not longer necessary from our side, and it's slowing the image editor. (It increases performance when image is calculated on every edition, but slows the interface if there are not performance issues and image calculation is performed only at the end of the edition process).

With this PR we suggest an approach to make the debounce time customizable via props so we can disable it only when needed.

Type checking and code flow improvements suggested by @andresin87

### DIFF
--- a/components/molecule/imageEditor/src/index.js
+++ b/components/molecule/imageEditor/src/index.js
@@ -10,7 +10,7 @@ import {baseClass, DEFAULT_ASPECT, getRotationDegrees, noop} from './config.js'
 import ImageEditorCropper from './ImageEditorCropper.js'
 import ImageEditorDefault from './ImageEditorDefault.js'
 import ImageEditorSliders from './ImageEditorSliders.js'
-import {debouncingTime} from './prop-types.js'
+import {debouncingTimePropType} from './prop-types.js'
 
 const MoleculeImageEditor = ({
   aspect = DEFAULT_ASPECT,
--- a/components/molecule/imageEditor/src/index.js
+++ b/components/molecule/imageEditor/src/index.js
@@ -6,22 +6,17 @@ import {debounce} from '@s-ui/js/lib/function/debounce.js'
 import Injector from '@s-ui/react-primitive-injector'
 
 import getCroppedImg from './utils/cropImage.js'
-import {
-  baseClass,
-  DEBOUNCING_TIME,
-  DEFAULT_ASPECT,
-  getRotationDegrees,
-  noop
-} from './config.js'
+import {baseClass, DEFAULT_ASPECT, getRotationDegrees, noop} from './config.js'
 import ImageEditorCropper from './ImageEditorCropper.js'
 import ImageEditorDefault from './ImageEditorDefault.js'
 import ImageEditorSliders from './ImageEditorSliders.js'
+import {debouncingTime} from './prop-types.js'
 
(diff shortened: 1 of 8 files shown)


davidmartin84 (approved):

components/molecule/imageEditor/src/index.js
```diff
@@ -6,21 +6,17 @@ import {debounce} from '@s-ui/js/lib/function/debounce.js'
...
-  noop,
-  getRotationDegrees
-} from './config.js'
+import {debouncingTime} from './prop-types.js'
 
 const MoleculeImageEditor = ({
   aspect = DEFAULT_ASPECT,
   cropLabelIcon,
   cropLabelText,
+  debouncingTime,
```
jordevo, line 19: given this is clearly a performance improvement, shouldn't we set a default value for this prop?

components/molecule/imageEditor/src/prop-types.js
```diff
@@ -0,0 +1,13 @@
+export const debouncingTime = function (props, propName, componentName) {
```
andresin87:
```suggestion
export const debouncingTimePropType = function (props, propName, componentName) {
```

components/molecule/imageEditor/src/index.js
```diff
@@ -6,21 +6,17 @@ import {debounce} from '@s-ui/js/lib/function/debounce.js'
...
+import ImageEditorDefault from './ImageEditorDefault.js'
 import ImageEditorSliders from './ImageEditorSliders.js'
-import {
-  baseClass,
-  DEBOUNCING_TIME,
-  DEFAULT_ASPECT,
-  noop,
-  getRotationDegrees
-} from './config.js'
+import {debouncingTime} from './prop-types.js'
```
andresin87:
```suggestion
import {debouncingTimePropType} from './prop-types.js'
```

components/molecule/imageEditor/src/index.js
```diff
@@ -83,6 +92,7 @@ MoleculeImageEditor.propTypes = {
   aspect: PropTypes.number,
   cropLabelIcon: PropTypes.node,
   cropLabelText: PropTypes.string,
+  debouncingTime,
```
andresin87:
```suggestion
  debouncingTime: debouncingTimePropType,
```

andresin87 (approved):

components/molecule/imageEditor/src/index.js
andresin87: not needed, if it's not declares we will recover the previous version. debounced on demand

jordevo: I'm more concerned about it from a functional standpoint. If the component without a debounce time is not performing correctly, shouldn't we provide a default value that makes the component work _by default_?

PR #2309, opened 2022-09-20
feat(components/molecule/tabs): Change tabs to be able to modify current active tabs by updating props

oegea:
### Types of changes

Current situation:

Currently, it is not possible to modify the current active tab by modifying the props of a `molecule/tabs` instance. The component maintains the current selected tab on its internal state, so even if props are modified to indicate which tab is the active one, the component doesn't reflect it on its state.

Introduced change:

This PR introduces a small change to detect when the active tab has been modified externally, and properly reflect it on its internal state.


jordevo (approved):
LGTM 🤩

andresin87 (approved):

quinwacca (approved):

PR #2337, opened 2022-09-29
feat(components/molecule/avatar): handle background color for avatar

tonyghiani:
## molecule/avatar
🔍 Show

### Types of changes

- [ ] 🧾 Documentation
- [ ] 📷 Demo
- [ ] 🧪 Test
- [ ] 🧠 Refactor
- [ ] 💄 Styles

When a passed src image was erroring and the `name` property was passed, the background-color for the fallback was not applied correctly.

### DIFF
--- a/components/molecule/avatar/test/index.test.js
+++ b/components/molecule/avatar/test/index.test.js
@@ -192,7 +192,7 @@ describe(json.name, () => {
       const {getByText} = setup(props)
 
       // Then
-      expect(getByText('JS')).to.be.visible
+      expect(getByText('js')).to.be.visible
     })
 
     describe('forwardRef', () => {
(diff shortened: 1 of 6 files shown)


components/molecule/avatar/test/index.test.js
```diff
@@ -192,7 +192,7 @@ describe(json.name, () => {
       const {getByText} = setup(props)
 
       // Then
-      expect(getByText('JS')).to.be.visible
+      expect(getByText('js')).to.be.visible
```
jordevo, line 195: cool initials 😉 

so the test is no longer `should render fallback name initials uppercase`?

jordevo (approved):

components/molecule/avatar/test/index.test.js
tonyghiani: The thing is I've changed the way initials are "uppercased", using the CSS transform property instead of the `.toUpperCase()`  js method. The final visual result is the same :)

andresin87 (approved):

ferransimon (approved):

components/molecule/avatar/test/index.test.js
jordevo: oh gotcha!

PR #2354, opened 2022-10-02
Fix/1795 input number safari

### LINKED ISSUE #1795: Input - "Number only" actually accepts other characters in Safari
**Describe the bug**
Input - "Number only" actually accepts other characters in Safari

**Steps to reproduce**
1. Go to [Input demo](https://sui-components.vercel.app/workbench/atom/input/demo) in Safari
2. Scroll down to the Number example
3. See error

https://user-images.githubusercontent.com/23620759/136557581-799a31a8-b253-4483-b505-04925d61b76b.mov

**Expected behaviour**
This option should only allow numbers

**Environment / Context**
- Desktop or Mobile
- Browser and version Safari 14.1.2
- Device: Macbook

stefi23:
## Category/Component

**TASK**:  #1795 

### Types of changes

- [ ] 🧾 Documentation
- [ ] 📷 Demo
- [ ] 🧪 Test
- [ ] 🧠 Refactor
- [ ] 💄 Styles

- Input type number accepts other characters in the Safari and Firefox browser
- To fix the issue an additional validator function was added to the input type number

### DIFF
--- a/components/atom/input/src/index.js
+++ b/components/atom/input/src/index.js
@@ -17,7 +17,7 @@ const AtomInput = forwardRef(({type, ...props}, ref) => {
     case 'number':
       return (
         <Input
-          onKeyDown={e => checkIfValidNumberInput(e)}
+          onKeyDown={checkIfValidNumberInput}
           ref={ref}
           type={type}
           {...props}
--- a/components/atom/input/src/index.js
+++ b/components/atom/input/src/index.js
@@ -8,11 +8,29 @@ import Password from './Password/index.js'
 import {INPUT_SHAPES, TYPES} from './config.js'
 
 const AtomInput = forwardRef(({type, ...props}, ref) => {
+  const checkIfValidNumberInput = event => {
+    // Check if input type number is valid as input type number doesn't currently work in browsers like Safari and Firefox
+    // Allowing: Integers | Backspace | Tab | Delete | Left & Right arrow keys
+    const allowedCharacter =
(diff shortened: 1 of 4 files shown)


components/atom/input/src/index.js
```diff
@@ -8,11 +8,29 @@ import Password from './Password/index.js'
...
+  }
   switch (type) {
     case 'sui-password':
       return <Password ref={ref} {...props} />
     case 'mask':
       return <Mask ref={ref} {...props} />
+    case 'number':
+      return (
+        <Input
+          onKeyDown={e => checkIfValidNumberInput(e)}
```
andresin87:
```suggestion
          onKeyDown={checkIfValidNumberInput}
```

components/atom/input/src/index.js
```diff
@@ -8,11 +8,29 @@ import Password from './Password/index.js'
 import {INPUT_SHAPES, TYPES} from './config.js'
 
 const AtomInput = forwardRef(({type, ...props}, ref) => {
+  const checkIfValidNumberInput = event => {
```
andresin87: move it to settings and easy import it to not redeclare the function on every declaration hoisting and simply get the outside defined

stefi23: Thank you! I moved the function to a helper.js file. Would it be preferred to have it in the config file or in a setting named file?

andresin87: =)

components/atom/input/src/index.js
andresin87: simply direct set the fn to not declare a new fn handler on every render

PR #2370, opened 2022-10-04
docs(Root): change onSelectKey

andresin87:
## molecule/dropdownOption

`🔍 Show`

**TASK**: 

### Types of changes
`onSelectKey` is used for store the key string codes wich selects options
`onWatever` prop naming convention is reserver for handlers in React
`selectKey`must be a proper naming

This iteration fixes the arrow navigation between options in select and autosuggest

- [x] 💥 Breaking change (fix or feature that would cause existing functionality to not work as expected)
- [ ] 🧾 Documentation
- [ ] 📷 Demo
- [ ] 🧪 Test
- [ ] 🧠 Refactor
- [ ] 💄 Styles

### DIFF
--- a/components/molecule/autosuggest/src/index.js
+++ b/components/molecule/autosuggest/src/index.js
@@ -81,7 +81,7 @@ const MoleculeAutosuggest = ({
       refsMoleculeAutosuggestOptions.current[index] = createRef()
       return cloneElement(child, {
         innerRef: refsMoleculeAutosuggestOptions.current[index],
-        onSelectKey: keysSelection
+        selectKey: keysSelection
       })
     })
 
@@ -117,7 +117,6 @@ const MoleculeAutosuggest = ({
     ev.persist()
     const {current: domInnerInput} = refMoleculeAutosuggestInput
     const {current: optionsFromRef} = refsMoleculeAutosuggestOptions
-    const {current: innerRefInput} = innerRefMoleculeAutosuggestInput
     const {key} = ev
     const options = optionsFromRef.map(getTarget)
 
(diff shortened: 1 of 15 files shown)


components/molecule/autosuggest/src/index.js
```diff
@@ -180,8 +179,22 @@ const MoleculeAutosuggest = ({
   }
 
   const handleClick = () => {
+    handleFocusInput()
```
andresin87, line 182: it could be changed to a direct assignation:
`const handleClick = handleFocusInput`
but i think its better leave it on this way just for not affecting to future click code iterations

Pato2790 (approved):

andresin87:
# Take care about names you give to variables:

- onSomething: handler (onSelectKey is not a handler? --> mustnt have on prefix)

Here a short guide which can help anyone to determine proper names to variables:

## reserved prefixes
- useSomething: hook
- withSomething: HoC
- onSomething: handler

## Is it a function? Use a verb!
const handleThat = () => {}

### Is it async? Use gerund -ing!
const handlingThat = async () => {}


## Is it a boolean —> better use a semantic affirmative name
❌ isInvisible=true
✅ isVisible=false

❌ hasRestrictions=false
✅ hasPermisions=true

## Is it a number or string?
Use singular
const fruit, car, person, user, number

## Is it an Array, Set, Map, …?
Use plural
const fruits, cars, persons, users, numbers

## Is it a class or a react-component?
Use initial capital letter
❌ class car {}
✅ class Car {}

❌ const component = () => null
✅ const Component = ()=>null

## And also try to avoid acronyms…
key pressing has no cost, trust me, its free and maintainers gonna apreciate every single meaningfull character you wrote. Try to be empathic with those who will maintain your code.
❌WYSIWYG
✅ whatYouSeeIsWhatYouGet

❌ DRY
✅ dontRepeatYourself

PR #2397, opened 2022-10-13
test(components/molecule/select): fix flaky onblur test

alfdocimo:
## Category/Component

**TASK**: 

### Types of changes

- [ ] 🧾 Documentation
- [ ] 📷 Demo
- [x] 🧪 Test
- [ ] 🧠 Refactor
- [ ] 💄 Styles

Avoid weird race condition with `userEvent.tab()` by firing `focusIn` and `focusOut` events

### DIFF
--- a/components/molecule/select/test/index.test.js
+++ b/components/molecule/select/test/index.test.js
@@ -11,7 +11,7 @@ import chai, {expect} from 'chai'
 import chaiDOM from 'chai-dom'
 import sinon from 'sinon'
 
-import userEvents from '@testing-library/user-event'
+import {fireEvent} from '@testing-library/react'
 
 import json from '../package.json'
 import * as pkg from '../src/index.js'
@@ -100,11 +100,11 @@ describe(json.name, () => {
       const {getByRole} = setup(props)
       const textBox = getByRole('textbox')
 
-      userEvents.click(textBox)
-      userEvents.tab()
+      fireEvent.focusIn(textBox)
+      fireEvent.focusOut(textBox)
 
       // Then
-      sinon.assert.calledOnce(spy)
+      sinon.assert.called(spy)
     })
   })
 
--- a/components/molecule/select/test/index.test.js
+++ b/components/molecule/select/test/index.test.js
@@ -100,11 +100,11 @@ describe(json.name, () => {
       const {getByRole} = setup(props)
       const textBox = getByRole('textbox')
 
-      fireEvent.focusIn(textBox)
-      fireEvent.focusOut(textBox)
+      textBox.focus()
+      fireEvent.blur(textBox)
 
       // Then
-      sinon.assert.called(spy)
+      sinon.assert.calledOnce(spy)
     })
   })
 


AgonisticKatai (approved):

andresin87 (approved):

components/molecule/select/test/index.test.js
```diff
@@ -100,11 +100,11 @@ describe(json.name, () => {
...
       const textBox = getByRole('textbox')
 
-      userEvents.click(textBox)
-      userEvents.tab()
+      fireEvent.focusIn(textBox)
+      fireEvent.focusOut(textBox)
 
       // Then
-      sinon.assert.calledOnce(spy)
+      sinon.assert.called(spy)
```
jordevo: isn't `calledOnce` a better choice for this kind of test? being it more specific, the test is more precise, IMHO. What do you think?

jordevo (approved):

components/molecule/select/test/index.test.js
```diff
@@ -100,11 +100,11 @@ describe(json.name, () => {
       const {getByRole} = setup(props)
       const textBox = getByRole('textbox')
 
-      userEvents.click(textBox)
-      userEvents.tab()
+      fireEvent.focusIn(textBox)
+      fireEvent.focusOut(textBox)
```
jordevo: https://testing-library.com/docs/guide-events

have a look at this... perhaps that's why you're feeling it's a bit flaky on your local environment? 🤔  

perhaps using imperative focus / blur is more robust:
https://testing-library.com/docs/guide-events#alternatives

components/molecule/select/test/index.test.js
alfdocimo: Good catch! will update this. Thanks! (:

components/molecule/select/test/index.test.js
alfdocimo: Interesting! Thanks for sharing! 💡 

I have given it a read and updated this with the imperative `focus()`

Seems like for some reason, calling imperatively `blur()` does not seem to work as you'd expect 🤔 
Captura de Pantalla 2022-10-14 a las 10 11 29
Captura de Pantalla 2022-10-14 a las 10 11 08

jordevo: oh that's sad... 🤔 

as an interim solution... how do you feel about doing focus first on `textBox`, as you're already doing, and then on some other element so that it _hopefully_ fires the blur event on the textbox? I know it's a bit hacky but perhaps more robust than the original `fireEvent` focus and blur.

andresin87: Better use click and tab, Its what user is going to tackle, dont care if it blurs or focus on his/her POV

alfdocimo: I agree, `click` and `tab` should be what we aim for, which was my original intention

```jsx
userEvents.click(textBox)
userEvents.tab()
```

After trying quite different things, like rendering another component and clicking that component 

```jsx

      const component = props => (
        <>
          <div>
            <button>Hello!</button>
          </div>
          <Component {...props} />
        </>
      )

      const setup = setupEnvironment(component)
```

I've realized that this behavior happens whether the browser that runs the tests is opened/closed. 🤯 

ezgif com-gif-maker (1)

only with `fireEvent.blur()` it fires the event from the node, regardless of wether the Karma browser is opened or not...

alfdocimo:
Merging this PR.
We are aware that there are some issues with focusing in and out of the Karma browser when running the tests. This is why firing an event directly seemed to fix the test. As a temporary workaround we're using `fireEvent` to trigger the blur event from the node

PR #2398, opened 2022-10-14
feat(components/molecule/selectPopover): show overlay from selected type

AgonisticKatai:
#### `❓ Ask`

**TASK**: [MTR-59903](https://jira.scmspain.com/browse/MTR-59903)

### Types of changes

- [ ] 🧾 Documentation
- [ ] 📷 Demo
- [ ] 🧪 Test
- [ ] 🧠 Refactor
- [ ] 💄 Styles

We added the functionality to optionally display an **overlay below the popover**.
Three types of overlay have been defined: **_`dark`_**, **_`light`_** & **_`none`_**.
**By default the overlay type is none**, so we guarantee backward compatibility with the current **SelectPopover** behavior.

This functionality helps us for example in cases where this component is displayed in a topbar and we want to apply an overlay only to the content that is below the SelectPopover and not on the entire screen as it would happen in a standard modal.

### DARK OVERLAY
Kapture 2022-10-14 at 12 17 39

### LIGHT OVERLAY
Kapture 2022-10-14 at 12 19 41

### NONE OVERLAY (current and default behavior)
Kapture 2022-10-14 at 12 14 13


components/molecule/selectPopover/src/index.js
```diff
@@ -258,11 +278,24 @@ function MoleculeSelectPopover({
...
-      {renderSelect()}
-      {renderContentWrapper()}
-    </div>
+    <>
+      <div className={classNames} title={title}>
+        {renderSelect()}
+        {renderContentWrapper()}
+      </div>
+      {hasOverlay && (
+        <Portal as={Fragment}>
```
andresin87, line 288:
```suggestion
        <Portal as={Fragment} isOpen={isOpen}>
```

components/molecule/selectPopover/src/styles/index.scss
```diff
@@ -101,4 +101,27 @@ $base-class: '.sui-MoleculeSelectPopover ';
       transform: rotate(-90deg) !important;
     }
   }
+
+  &-overlay {
+    @each $name, $type in $molecule-select-popover-overlay-types {
+      &--#{'' + $name} {
```
andresin87:
```suggestion
      &--#{$name} {
```

components/molecule/selectPopover/src/styles/index.scss
```diff
@@ -101,4 +101,27 @@ $base-class: '.sui-MoleculeSelectPopover ';
...
+        height: 100%;
+        left: 0;
+        opacity: 0;
+        pointer-events: none;
+        position: absolute;
+        top: 0;
+        transition: opacity 0.25s ease-in;
+        width: 100%;
+
+        &.is-open {
```
andresin87, line 120: don not need this class, assign the css rules directly to the upper selector cuz portal will remove the element from the dom

components/molecule/selectPopover/src/index.js
```diff
@@ -258,11 +278,24 @@ function MoleculeSelectPopover({
     renderContentWrapperProp && `${BASE_CLASS}--hasCustomWrapper`
   )
 
+  const overlayClassNames = cx(
+    `${BASE_CLASS}-overlay`,
+    `${BASE_CLASS}-overlay--${overlayType}`,
+    {'is-open': isOpen}
```
andresin87, line 278: not needed

andresin87: if needed, better use an accessibility proper attribute solution like `aria-hidden: false`

components/molecule/selectPopover/src/index.js
AgonisticKatai: DONE!

components/molecule/selectPopover/src/styles/index.scss
AgonisticKatai: DONE!

components/molecule/selectPopover/src/styles/index.scss
AgonisticKatai: DONE!

components/molecule/selectPopover/src/index.js
AgonisticKatai: REMOVED, DONE!

andresin87:
The overlay fadein animation is fired smoothly but fadeout do not.
Is it OK?

AgonisticKatai:
> The overlay fadein animation is fired smoothly but fadeout do not.
> Is it OK?

Hi @andresin87, no, this is an undesired behavior.
It seems that it is related to the last change uploaded in which we show the element in the dom through the **_isOpen_** property of the **_Portal_**.
This makes that when we close the modal it removes the element from the dom and we lose the animation.
Do you think it would be a better idea to show the element always visible as it was before (without passing the **_isOpen_** property and using the value true that has by default the **_usePortal_**) and show the animation through an **_is-open_** class that is added when the user interacts with the popover?

turolopezsanabria (approved):

PR #2399, opened 2022-10-14
feat(components/molecule/avatar): make optional hex background of fallback name

quinwacca:
### Types of changes

Until now, name fallback has a randomly generated background color from the name.
At motor, we need to use a defined background color for every name fallback avatar.

### DIFF
--- a/components/molecule/avatar/src/AvatarFallbackName/index.js
+++ b/components/molecule/avatar/src/AvatarFallbackName/index.js
@@ -29,7 +29,7 @@ const MoleculeAvatarFallbackName = ({
     <div
       className={className}
       aria-label={nameProp}
-      style={backgroundColor}
+      style={{backgroundColor}}
       {...others}
     >
       {name}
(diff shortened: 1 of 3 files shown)


jordevo (approved):

andresin87:
why hex? why boolean?

components/molecule/avatar/src/AvatarFallbackName/index.js
```diff
@@ -21,7 +22,8 @@ const MoleculeAvatarFallbackName = ({
       ? `${firstName.charAt(0)}${lastName.charAt(0)}`
       : firstName.charAt(0)
 
-  const backgroundColor = useConvertStringToHex(nameProp)
+  const stringToHexBackgroundColor = useConvertStringToHex(nameProp)
```
andresin87: backgroundColor = stringToHexBackgroundColor || useConvertStringToHex(nameProp)

useConvertStringToHex is a complex fn. It only needs its execution if the color is not proper declared

andresin87 (changes requested):
Also declare the prop `fallbackColor` in the MoleculeAvatar for get it documented on the doc tab and remove the rest prop-spreading from the renderContent observables and declare it imperative using `fallbackColor={fallbackColor}`

components/molecule/avatar/src/AvatarFallbackName/index.js
quinwacca: but hooks cannot be called conditionally, I'll change the hook then

andresin87 (approved):

PR #2441, opened 2022-10-25
Feat/atom slider fullwidth

javiauso:
## Atom/Slider

`🔍 Show`

**TASK**: 

### Types of changes

- [ ] 🧾 Documentation
- [ ] 📷 Demo
- [ ] 🧪 Test
- [ ] 🧠 Refactor
- [ ] 💄 Styles

In MA we need a full width design of AtomSlider, to show custom marks inside a modal.

### Screenshots - Animations

image

### DIFF
--- a/components/atom/slider/src/index.js
+++ b/components/atom/slider/src/index.js
@@ -31,7 +31,7 @@ const AtomSlider = ({
   hideTooltip = false,
   defaultValue,
   invertColors,
-  fullWidth
+  isFullWidth
 }) => {
   let initialStateValue
   const refAtomSlider = useRef()
@@ -84,7 +84,7 @@ const AtomSlider = ({
     onChange: handleChange,
     onAfterChange: handleAfterChange,
     disabled,
-    marks: hideMarks ? {} : markerFactory({step, min, max, marks, fullWidth}),
+    marks: hideMarks ? {} : markerFactory({step, min, max, marks, isFullWidth}),
     max,
     min,
     step,
@@ -95,7 +95,7 @@ const AtomSlider = ({
     BASE_CLASS,
     {[CLASS_DISABLED]: disabled},
     {[CLASS_INVERSE]: invertColors},
-    {[CLASS_FULLWIDTH]: fullWidth}
+    {[CLASS_FULLWIDTH]: isFullWidth}
   )
 
   // Determine the type of the slider according to the range prop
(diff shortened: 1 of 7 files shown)


components/atom/slider/src/index.js
```diff
@@ -168,7 +170,10 @@ AtomSlider.propTypes = {
   hideTooltip: PropTypes.bool,
 
   /* If true it will invert the colors for selected track and rail */
-  invertColors: PropTypes.bool
+  invertColors: PropTypes.bool,
+
+  /* If true it will render in a full witdh design */
```
jordevo:
```suggestion
  /* If true it will render in a full width design */
```

sorry! 🤓

components/atom/slider/src/markerFactory.js
```diff
@@ -22,11 +22,30 @@ const linearMarksFactory = ({step, min, max, ticks}) => {
...
+  const firstMarkKey = marksKeys[0]
+  const lastMarkKey = marksKeys[marksKeys.length - 1]
+
+  const nextMarks = {
+    ...marks,
+    [firstMarkKey]: {label: marks[firstMarkKey], style: {transform: 'inherit'}},
+    [lastMarkKey]: {
+      label: marks[lastMarkKey],
+      style: {transform: 'inherit', left: 'auto', right: '0'}
+    }
```
jordevo, line 36: do you think we could achieve implement this in a pure CSS way? 

just as an idea, could we style `first-child` and `last-child` markers accordingly if the component has the `fullWidth` prop set to `true`?  

IMHO it'd be cleaner... what do you think?

components/atom/slider/src/index.js
```diff
@@ -29,7 +30,8 @@ const AtomSlider = ({
   hideMarks = false,
   hideTooltip = false,
   defaultValue,
-  invertColors
+  invertColors,
+  fullWidth
```
andresin87: Boolean prop... isWhatever or hasWhatever

components/atom/slider/src/markerFactory.js
javiauso: Yes! I'm with you with the approach, but here we're using `rc-slider` and is managing this styles with inline mode, so we only do that in CSS using `!important` declaration, instead them provide this `style` prop for each markers. For me, we could do an exception using `!important`, but It looks a bit tricky.

jordevo: oh... if the only alternative we have here is using `!important`, I think your original implementation is cleaner then

andresin87 (approved):

PR #2467, opened 2022-11-16
fix(components/atom/switch): FullWidth switch should have the text with align left

SergioColladoV:
## atom/switch

`❓ Ask` 

**TASK**: 

### Types of changes

- [ ] 🧾 Documentation
- [ ] 📷 Demo
- [ ] 🧪 Test
- [ ] 🧠 Refactor
- [ ] 💄 Styles

FullWidth switch should have the text with align left

### DIFF
--- a/components/atom/switch/src/SwitchType/single.js
+++ b/components/atom/switch/src/SwitchType/single.js
@@ -65,6 +65,7 @@ export const SingleSwitchTypeRender = forwardRef(
               name={name}
               text={defaultLabelLeft ? label : labelLeft}
               optionalText={labelOptionalText}
+              inline={fullWidth ? 'left' : 'right'}
             />
           )}
           <button
--- a/components/atom/switch/src/SwitchType/single.js
+++ b/components/atom/switch/src/SwitchType/single.js
@@ -65,7 +65,6 @@ export const SingleSwitchTypeRender = forwardRef(
               name={name}
               text={defaultLabelLeft ? label : labelLeft}
               optionalText={labelOptionalText}
-              inline={fullWidth ? 'left' : 'right'}
             />
           )}
           <button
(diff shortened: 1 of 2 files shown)


turolopezsanabria (approved):

components/atom/switch/src/styles/index.scss
```diff
@@ -259,5 +259,9 @@ $base-class: '.sui-AtomSwitch';
 
   &--fullWidth {
     width: 100%;
+
+    .sui-AtomLabel {
+      text-align: left;
+    }
```
jordevo, line 265: IMHO it is not ideal to override styles like this...

not sure if `atom/label`'s `inline` prop could help? 
https://github.com/SUI-Components/sui-components/tree/master/components/atom/label#basic-usage 

anyway, I'd try to avoid overriding styles this way. It couples `atom/swith` to `atom/label`'s styles and I think this makes our styles more difficult to manage as our components grow.

andresin87 (approved):

components/atom/switch/src/styles/index.scss
SergioColladoV: Can be a good solution

components/atom/switch/src/SwitchType/single.js
```diff
@@ -65,6 +65,7 @@ export const SingleSwitchTypeRender = forwardRef(
               name={name}
               text={defaultLabelLeft ? label : labelLeft}
               optionalText={labelOptionalText}
+              inline={fullWidth ? 'left' : 'right'}
```
SergioColladoV: @jordevo I think this is your proposal but it doesn't work

SergioColladoV: @jordevo The inline prop should change the text-align but it changes the margin. It can't work :(

PR #2531, opened 2023-04-27
feat(components/atom/videoPlayer): Add a prop to configure if control…

oegea:
…s should be displayed

**Context**
Implementing video-related features in Adevinta.

**Description**
Add a prop to programmatically indicate if the videoplayer should display controls or not.

### Types of changes

### DIFF
--- a/components/atom/videoPlayer/src/components/NativePlayer.js
+++ b/components/atom/videoPlayer/src/components/NativePlayer.js
@@ -5,7 +5,7 @@ import PropTypes from 'prop-types'
 import useGetBlobAsVideoSrcEffect from '../hooks/useGetBlobAsVideoSrcEffect.js'
 import {BASE_CLASS, NATIVE_DEFAULT_TITLE} from '../settings/index.js'
 
-const NativePlayer = ({src, title = NATIVE_DEFAULT_TITLE}) => {
+const NativePlayer = ({controls, src, title = NATIVE_DEFAULT_TITLE}) => {
   const videoNode = useRef(null)
   const videoSrc = useGetBlobAsVideoSrcEffect({src, videoNode})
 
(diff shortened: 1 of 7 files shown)


components/atom/videoPlayer/src/index.js
```diff
@@ -17,7 +20,9 @@ const AtomVideoPlayer = forwardRef(({src = ''}, forwardedRef) => {
 AtomVideoPlayer.displayName = 'AtomVideoPlayer'
 
 AtomVideoPlayer.propTypes = {
+  controls: PropTypes.bool,
```
jordevo, line 23: shouldn't we stick to naming props in a boolean-ish way? ie `hasControls`? even if that's mapped down the component tree to some other prop / HTML tag...?

components/atom/videoPlayer/src/components/NativePlayer.js
```diff
@@ -15,7 +15,7 @@ const NativePlayer = ({src, title = NATIVE_DEFAULT_TITLE}) => {
         className={`${BASE_CLASS}-nativePlayerVideo`}
         ref={videoNode}
         title={title}
-        controls
+        controls={controls}
       >
         {videoSrc !== null && <source data-testid="videosrc" src={videoSrc} />}
         Your browser does not support the video tag.
```
andresin87, line 21: try to avoid hardcoded literals. Its imposible to translate later

components/atom/videoPlayer/test/index.test.js
```diff
@@ -24,7 +24,7 @@ describe('AtomVideoPlayer', () => {
 
   it('should render without crashing', () => {
     // Given
-    const props = {}
+    const props = {src: ''}
```
andresin87, line 27: if there is no value (`undefined`) given to src it crushes? In that case plz assign a defaultValue to it

PR #2532, opened 2023-04-27
feat(components/atom/videoPlayer): Add autoplay prop to atom videoPlayer

davidmartin84:
Implementing video-related features in Adevinta.

Add a prop to programmatically indicate if the videoplayer should autoplay or not.

### Types of changes

### DIFF
--- a/components/atom/videoPlayer/src/index.js
+++ b/components/atom/videoPlayer/src/index.js
@@ -5,21 +5,25 @@ import PropTypes from 'prop-types'
 import useVideoPlayer from './hooks/useVideoPlayer.js'
 import {BASE_CLASS} from './settings/index.js'
 
-const AtomVideoPlayer = forwardRef(({controls = true, src}, forwardedRef) => {
-  const [Component, props] = useVideoPlayer({
-    controls,
-    src
-  })
-  return (
-    <div ref={forwardedRef} className={BASE_CLASS}>
-      <Component {...props} />
-    </div>
-  )
-})
+const AtomVideoPlayer = forwardRef(
+  ({autoPlay = false, controls = true, src}, forwardedRef) => {
+    const [Component, props] = useVideoPlayer({
+      autoPlay,
+      controls,
+      src
+    })
+    return (
+      <div ref={forwardedRef} className={BASE_CLASS}>
+        <Component {...props} />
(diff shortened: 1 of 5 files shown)


components/atom/videoPlayer/src/index.js
```diff
@@ -5,21 +5,25 @@ import PropTypes from 'prop-types'
...
+        <Component {...props} />
+      </div>
+    )
+  }
+)
 
 AtomVideoPlayer.displayName = 'AtomVideoPlayer'
 
 AtomVideoPlayer.propTypes = {
+  autoPlay: PropTypes.bool,
```
jordevo, line 26: same as previously commented to @oegea , wouldn't it be best if we stick to boolean-ish names for boolean props? 🤔 

  ```suggestion
  hasAutoPlay: PropTypes.bool,
```

File: components/atom/videoPlayer/src/index.js [hunk 2]
```diff
@@ -5,21 +5,25 @@ import PropTypes from 'prop-types'
...
-    src
-  })
-  return (
-    <div ref={forwardedRef} className={BASE_CLASS}>
-      <Component {...props} />
-    </div>
-  )
-})
+const AtomVideoPlayer = forwardRef(
+  ({autoPlay = false, controls = true, src}, forwardedRef) => {
```
jordevo, line 9: good to see, from a user POV, it's `false` by default

components/atom/videoPlayer/src/index.js
hpintos: IMHO, make sense to leave that property name as is the same that video tag natively uses.

hpintos (approved):

quinwacca (approved):

components/atom/videoPlayer/src/index.js
oegea: Yep, we'll review all property names once we have the component in a more advanced status. 

We're not really sure which is the best approach, but currently we are trying to respect the standard html attribute naming.

Maybe if we start to add custom props that does not match with any standard html attribute (which will probably happen), then we'll have to think about changing the approach and name all props without trying to match them with html standard attributes.

So... long story short, we have this on the radar 🤓

PR #2535, opened 2023-05-02
feat(components/atom/videoPlayer): Add offset parameter to determine …

oegea:
…the start time of the video

Add offset prop to define the start time on any video played using the atom/videoPlayer component.

### Types of changes


components/atom/videoPlayer/src/index.js
```diff
@@ -25,8 +26,8 @@ AtomVideoPlayer.displayName = 'AtomVideoPlayer'
 AtomVideoPlayer.propTypes = {
```
andresin87, line 26: at the end plz you need to comment all this props. What they do?

PR #2554, opened 2023-05-10
feat(components/molecule/collapsible): add isCollapsible boolean prop

jordevo:
So that `molecule/collapsible` can be `collapsed` or not, when expanded, at will.

## Category/Component

**TASK**: 

### Types of changes

- [x] 🧾 Documentation
- [x] 📷 Demo
- [ ] 🧪 Test
- [ ] 🧠 Refactor
- [ ] 💄 Styles
- [ ] 🛠️ Tool

### Screenshots - Animations

image

image

### DIFF
--- a/components/molecule/collapsible/demo/ArticleNoCollapse.js
+++ b/components/molecule/collapsible/demo/ArticleNoCollapse.js
@@ -37,7 +37,8 @@ const ArticleNoCollapse = ({className, icon, showText, hideText, height}) => {
         </div>
       </DemoWrapper>
       <Paragraph>
-        The component is set to not collapsible, regardless of contents' height
+        The component is set to not collapse once expanded, regardless of
+        contents' height
       </Paragraph>
       <br />
       <br />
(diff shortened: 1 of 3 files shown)


components/molecule/collapsible/demo/ArticleNoCollapse.js
```diff
@@ -36,6 +36,52 @@ const ArticleNoCollapse = ({className, icon, showText, hideText, height}) => {
...
+              backgroundColor: '#fff',
+              fontSize: 14,
+              padding: 16,
+              textAlign: 'left',
+              maxWidth: 400
+            }}
+          >
+            <MoleculeCollapsible
+              icon={icon}
+              isCollapsible={false}
```
AgonisticKatai, line 59: It is little strange to me that a MoleculeCollapsible component may not be collapsible. 🧐

jordevo: good point... perhaps it makes more sense to have something like this rather than, IDK, another component say `molecule/expandable` for cases where we only want the user to be able to expand the contents, and not collapse them back.

In the end, the same could be argued about the `disabled` prop for a button and yet, I still want to have that `prop` 😉

andresin87 (approved):

PR #2565, opened 2023-05-17
feat(components/atom/button): button inner inherits text decoration

quinwacca:
The text-decoration of ALL atom buttons with link style was removed on Coches.net.
The cause was a bug fix on Chrome. The inner container SHOULD NOT inherit the text-decoration of its parent.
Believing it was a Firefox bug, a user opened a bug on Mozilla 6 years ago: https://bugzilla.mozilla.org/show_bug.cgi?id=1344545
The same engineer that replied that it was a Chrome bug, remembered this issue 3 months ago and sent the bug to the Chromium team: https://bugs.chromium.org/p/chromium/issues/detail?id=1414859

So, back to SUI Components, the fix proposed here it's to ensure the inheritance of the text-decoration on the inner element, as it was the expected behavior.

### Types of changes
- [x] 💄 Styles

### Screenshots - Animations
BEFORE: 
image

NOW (expected):
image

### DIFF
--- a/components/atom/button/src/styles/index.scss
+++ b/components/atom/button/src/styles/index.scss
@@ -72,6 +72,7 @@ $base-class: '.sui-AtomButton';
     display: inline-flex;
     height: 100%;
     pointer-events: none;
+    text-decoration: inherit;
   }
 
   &--loading {


jordevo (commented):
👏

hpintos (approved):
Man! I talk about this with UX last week because I thought something was missing! Thanks! 💪

PR #2575, opened 2023-06-08
feat(components/molecule/photoUploader): Allow extra params on photo uploaded

oriolpuig:
#### `❓ Ask`

**TASK**:  NO JIRA

With this new feature, we allow passing extra props when a new photo is uploaded from outside to internal component. 
For example, we want to perform 2 calls when user add a new photo, one to analyse the photo and the other to store it. The first call can delay a little bit the UX and that's not an option. 

So, we will create a custom id and pass it to photoUploader component, in order to identify it when is updated or deleted.

### Types of changes

### DIFF
--- a/components/molecule/photoUploader/src/PhotosPreview/index.js
+++ b/components/molecule/photoUploader/src/PhotosPreview/index.js
@@ -94,11 +94,13 @@ const PhotosPreview = ({
         list[index].blob = blob
         list[index].isModified = true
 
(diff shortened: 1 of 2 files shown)


andresin87 (approved):

andresin87:
❤️

jordevo (approved):

quinwacca (approved):

PR #2590, opened 2023-07-19
feat(components/atom/videoPlayer): Expose useDetectVideoType hook and UNKNOWN constant

oegea:

As the `atom/videoPlayer` component contains logic to identify a specific video type, and in order to avoid duplicating that logic to other places, we are modifying the component to expose the hook that currently wraps that functionality.

### Types of changes


davidmartin84 (approved):

andresin87 (commented):
✅

PR #2597, opened 2023-07-24
feat(components/atom/tag): Change the default tap highlight color for…

Alisa-InWonderland:
… Chrome mobile

#### `🔍 Show`

**TASK**: 

Add the possibility to change the default tap highlight color for Chrome mobile (only for actionable tags)

### Types of changes

- [ ] 🧾 Documentation
- [ ] 📷 Demo
- [ ] 🧪 Test
- [ ] 🧠 Refactor
- [ ] 💄 Styles
- [ ] 🛠️ Tool

### Screenshots - Animations
Chrome mobile default tap highlight color
Captura de pantalla 2023-07-24 a las 17 41 27

### DIFF
--- a/components/atom/tag/src/_settings.scss
+++ b/components/atom/tag/src/_settings.scss
@@ -1,7 +1,6 @@
 $bd-atom-tag: none !default;
 $bgc-atom-tag: color-variation($c-gray, 3) !default;
 $mw-label: 240px !default;
-$thc-atom-tag: inherit !default;
 
 // sizes
 $h-atom-tag-l: 40px !default;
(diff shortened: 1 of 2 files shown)


nucliweb:
Really we want to remove the highlighted?

_IMO, remove it is an a11y anti-pattern._

> The highlighting indicates to the user that their tap is being successfully recognized, and indicates which element they're tapping on.

Wdyt on adding the default value with a variable, to have the option to redefine in the theme?

Alisa-InWonderland:
> Really we want to remove the highlighted?
> 
> _IMO, remove it is an a11y anti-pattern._
> 
> > The highlighting indicates to the user that their tap is being successfully recognized, and indicates which element they're tapping on.
> 
> Wdyt on adding the default value with a variable, to have the option to redefine in the theme?

I wanted to have the possibility to remove the highlighting because this component already have custom styles for active state of the tag and it seems to me that the active state performs the same role as the highlighting. Now the highlighting and mentioned styles are activated at the same time. In addition, the highlighting has a rectangular shape, and the tag is rounded, which looks a bit weird. Because of all this I thought that we could use active state without highlighting and it will look better. What do you think?

Yes, with adding the default value with a variable I wanted to add the option to change the tap highlight color to transparent in the theme. This way my changes won't affect other verticals.

nucliweb (approved):
👏🏼

andresin87 (approved):

PR #2623, opened 2023-09-15
feat(components/atom/progressBar): Allow customizing the progress bar…

oegea:
… depending on its status

Due to UX requirements, we need to customize the container look of a specific progressBar status.

### Types of changes

### Screenshots - Animations

How our success bar looks: (no border, and a kind of green background on the container)

Captura de pantalla 2023-09-15 a las 13 49 43

How a standard bar looks: (White background, and 1px solid border)

Captura de pantalla 2023-09-15 a las 13 51 21

### DIFF
--- a/components/atom/progressBar/src/ProgressBarLine/settings.scss
+++ b/components/atom/progressBar/src/ProgressBarLine/settings.scss
@@ -1,4 +1,7 @@
 $bg-progress-bar: $c-gray-lightest !default;
+$bg-progress-bar--success: $bg-progress-bar !default;
+$bg-progress-bar--loading: $bg-progress-bar !default;
+$bg-progress-bar--error: $bg-progress-bar !default;
 $bc-progress-bar: $c-gray-light !default;
 $c-progress-bar-fill: $c-primary !default;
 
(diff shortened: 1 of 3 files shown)


components/atom/progressBar/src/ProgressBarLine/settings.scss
```diff
@@ -1,11 +1,17 @@
...
 $bc-progress-bar: $c-gray-light !default;
 $c-progress-bar-fill: $c-primary !default;
 
 $trsdu-progress-bar: 500ms !default;
 $trstf-progress-bar: linear !default;
 
 $bd-line-progress-container: $bdw-s solid $bc-progress-bar !default;
+$bd-line-progress-container--success: $bd-line-progress-container !default;
+$bd-line-progress-container--loading: $bd-line-progress-container !default;
+$bd-line-progress-container--error: $bd-line-progress-container !default;
```
jordevo, line 14: IMHO, according to our naming convention, variables like this:

```suggestion
$bd-line-progress-container
```

should be named 
```suggestion
$bd-progress-bar-container-line
```

the first part after the emmet bit (`bd`), should reference the component, `progress-bar` in this case.

jordevo (approved):

components/atom/progressBar/src/ProgressBarLine/settings.scss
oegea: After discussing with Jordi, we're going to leave it as it is now to avoid backward compatibility issues. We'll talk this with Andres at some moment.

PR #2634, opened 2023-09-26
Feat/xcomp 10406 phone input

1izpena:
#### `🔍 Show`

**TASK**: 

Make MoleculePhoneInput use MoleculeInputField instead of AtomInput. This way, we get more features like different types of help texts. Also, we remove the unnecessary wrapper of AtomIcon.

### Types of changes

- [ ] 🧾 Documentation
- [ ] 📷 Demo
- [ ] 🧪 Test
- [ ] 🧠 Refactor
- [ ] 💄 Styles
- [ ] 🛠️ Tool

### Screenshots - Animations
Screenshot 2023-09-26 at 12 06 13

### DIFF
--- a/components/molecule/phoneInput/src/index.js
+++ b/components/molecule/phoneInput/src/index.js
@@ -4,32 +4,34 @@ import cx from 'classnames'
 import PropTypes from 'prop-types'
 import flagIcons from 'rendered-country-flags'
 
-import AtomIcon, {
-  ATOM_ICON_COLORS,
-  ATOM_ICON_SIZES
-} from '@s-ui/react-atom-icon'
-import AtomInput from '@s-ui/react-atom-input'
 import {TYPES} from '@s-ui/react-atom-input/lib/config.js'
 import MoleculeDropdownList, {
   moleculeDropdownListDesigns
 } from '@s-ui/react-molecule-dropdown-list'
 import MoleculeDropdownOption from '@s-ui/react-molecule-dropdown-option'
+import MoleculeInputField from '@s-ui/react-molecule-input-field'
 
 import {phoneValidationType} from './settings.js'
 
 const BASE_CLASS = 'sui-MoleculePhoneInput'
 
+export {PREFIXES} from './settings.js'
 export default function MoleculePhoneInput({
+  alertText,
+  autoHideHelpText = false,
   dropdownCloseIcon,
   dropdownIcon,
-  value = '',
-  setFormattedValue,
-  prefixes = [],
+  errorText,
+  hasError,
+  helpText,
   onChange,
   placeholder,
-  hasError,
+  prefixes = [],
   initialSelectedPrefix = prefixes[0],
+  setFormattedValue,
+  successText,
   type = phoneValidationType.DEFAULT,
+  value = '',
   visiblePrefixes = true
 }) {
   const [showDropdown, setShowDropdown] = useState(false)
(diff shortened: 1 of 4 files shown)


xaviermariaconejo (approved):

components/molecule/phoneInput/src/index.scss
```diff
@@ -37,15 +37,40 @@
...
+          left: 0;
         }
 
         &.splitted {
           border-radius: $bdrs-l;
           border: 1px solid $c-gray-light;
           width: 100%;
 
+          & .sui-AtomInput-input--noBorder {
+            border-width: 1px 0;
```
Alverd04: I think there is a size for borders :3

components/molecule/phoneInput/src/index.scss
```diff
@@ -37,15 +37,40 @@
...
+            border-style: solid;
+            border-color: $c-gray-light;
+          }
+
           &.sui-MoleculePhoneInput--error {
             border-color: $c-error;
+
+            & .sui-AtomInput-input--noBorder {
+              border-width: 1px 0;
```
Alverd04: Same here

Alverd04 (approved):

Alverd04 (commented):
Did we check the coverage for the component, and add some tests to check what happens if no help text is given, or if there's an error make sure the error text is renderes if given?

components/molecule/phoneInput/src/index.js
```diff
@@ -102,28 +104,28 @@ export default function MoleculePhoneInput({
...
           )}
-          {visiblePrefixes && (
-            <AtomIcon
-              size={ATOM_ICON_SIZES.medium}
-              color={ATOM_ICON_COLORS.currentColor}
-            >
-              {showDropdown ? dropdownCloseIcon : dropdownIcon}
-            </AtomIcon>
-          )}
+          {visiblePrefixes && (showDropdown ? dropdownCloseIcon : dropdownIcon)}
```
andresin87, line 107:
```suggestion
          {visiblePrefixes && (
            <PrimitiveInjector
              size={ATOM_ICON_SIZES.medium}
              color={ATOM_ICON_COLORS.currentColor}
            >
              (showDropdown ? dropdownCloseIcon : dropdownIcon)
            </PrimitiveInjector>
            )}
```

There is a way to do not loose that configuration adding the functionality to let the user/consumer give its own Icon react component wrapper to the svg element

1izpena:
@Alverd04 Regarding your comment: _Did we check the coverage for the component, and add some tests to check what happens if no help text is given, or if there's an error, make sure the error text is rendered if given?_ I think it might not be necessary, as that responsibility belongs to `MoleculeInputField`, to which the handling of help texts is delegated, and it should be tested at the `MoleculeInputField` level as well.

components/molecule/phoneInput/src/index.scss
```diff
@@ -37,15 +37,40 @@
 
         & > input {
           background: none;
+
+          &.sui-AtomInput-input--noBorder {
+            position: relative;
+          }
```
jordevo, line 43: is it ok to be overriding styles this way? 🤔 shouldn't we allow for `atom/input` itself to behave as we want when required?

components/molecule/phoneInput/src/index.js
1izpena: I think the problem is that the icons already come with a wrapper that includes that configuration, and it was being duplicated and overwritten. 
@Alverd04 could you confirm this? 🙇‍♀️

components/molecule/phoneInput/src/index.scss
1izpena: What change would you propose? 😄 
The issue is that `MoleculeInputField` wasn't designed to work within the context of `MoleculePhoneInput`, and its behaviour varies within this context. On its own, `MoleculeInputField` works well; it's a single input and occupies the same space, so they move in harmony. However, within `MoleculePhoneInput`, there are additional components that break this behaviour. So, I understand that `MoleculeInputField` shouldn't be aware of the issues that `MoleculePhoneInput` might cause. 😄

jordevo: > So, I understand that `MoleculeInputField` shouldn't be aware of the issues that `MoleculePhoneInput` might cause. 😄

of course I don't mean any coupling or making `MoleculeInputField` aware of any wrapper / consumer. My point has more to do with potential problems down the road with this approach of overriding styles. If there's no other way around, fine, but we might find unexpected styles when theming for `MoleculeInputField`.

jordevo (approved):

ivanmlaborda (approved):

PR #2652, opened 2023-10-04
chore(Root): upgrade to node 18 and npm 9

andresin87:
### Types of changes

- [x] ⚙️ Environment


.github/workflows/node.js.yml
```diff
@@ -31,7 +31,7 @@ jobs:
       - uses: actions/checkout@v2
       - uses: actions/setup-node@v3
         with:
-          node-version: 16.15.0
+          node-version: 18.18.0
```
kikoruiz: I think you can avoid hardcoding the version by using this other way...

```suggestion
          node-version-file: '.nvmrc'
```

I will take the version from the `nvmrc` file.

kikoruiz (approved):

PR #2654, opened 2023-10-25
Feature/header label wrap prop

AgonisticKatai:
#### `🔍 Show`

**TASK**: [MTR-76371](https://jira.ets.mpi-internal.com/browse/MTR-76371) & [MTR-76164](https://jira.ets.mpi-internal.com/browse/MTR-76164)

Be able to define different header label wrap behaviors

### Types of changes

- [ ] 🧾 Documentation
- [ ] 📷 Demo
- [ ] 🧪 Test
- [ ] 🧠 Refactor
- [ ] 💄 Styles
- [ ] 🛠️ Tool

#### WITH WRAP BEHAVIOR (default)
image

#### WITH NO WRAP
image

### DIFF
--- a/components/molecule/accordion/src/settings.js
+++ b/components/molecule/accordion/src/settings.js
@@ -70,3 +70,8 @@ export const HEADER_ICON_POSITION = {
   LEFT: 'left',
   RIGHT: 'right'
 }
+
+export const HEADER_LABEL_WRAPS = {
+  NO_WRAP: 'noWrap',
+  WRAP: 'wrap'
+}
--- a/components/molecule/accordion/src/styles/index.scss
+++ b/components/molecule/accordion/src/styles/index.scss
@@ -107,8 +107,13 @@ $base-class-item-panel: '#{$base-class-item}Panel';
     font-family: $ff-molecule-accordion-header;
     overflow: hidden;
     text-align: inherit;
-    text-overflow: ellipsis;
-    white-space: nowrap;
+    &--wrap {
+      white-space: wrap;
+    }
+    &--noWrap {
+      text-overflow: ellipsis;
+      white-space: nowrap;
+    }
   }
   #{$base-class-item-header}ButtonContent,
   #{$base-class-item-header}Icon {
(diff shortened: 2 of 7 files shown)


components/molecule/accordion/src/styles/index.scss
```diff
@@ -107,8 +107,13 @@ $base-class-item-panel: '#{$base-class-item}Panel';
...
     text-align: inherit;
-    text-overflow: ellipsis;
-    white-space: nowrap;
+    &--wrap {
+      white-space: wrap;
+    }
+    &--noWrap {
+      text-overflow: ellipsis;
+      white-space: nowrap;
+    }
```
jordevo, line 116: 🍆 would it make sense to keep one of the styles as the default one? 🤔 we could skip one of the two modifiers this way

jordevo (approved):

quinwacca (approved):

components/molecule/accordion/src/settings.js
```diff
@@ -70,3 +70,8 @@ export const HEADER_ICON_POSITION = {
   LEFT: 'left',
   RIGHT: 'right'
 }
+
+export const HEADER_LABEL_WRAPS = {
+  NO_WRAP: 'noWrap',
+  WRAP: 'wrap'
+}
```
jordevo, line 77: 🍆 unless you're thinking of adding more styles of wrapping in the future... wouldn't a `boolean` prop do the job? 🤔

emiliovillu (approved):

components/molecule/accordion/src/styles/index.scss
AgonisticKatai: Personally, I believe that the management of modifiers can be done in both ways and both are correct.
The positioning of the icon is the same approach. In that case we could also use as default the right one and pass a prop isLeftIconPosition and it would do the same. In the case of the icon there is not going to be any other case than left or right and the modifier is always applied to indicate the positioning.

In case we see a big detriment to do it this way we can change it, but I personally do not see any inconvenience in doing it with modifiers that indicate that type of wrap has the label, besides that it gives us the possibility to apply more cases if in the future it is necessary.

andresin87:
https://developer.mozilla.org/en-US/docs/Web/CSS/text-wrap

What about considering adding text-wrap: balance for the wrap behavior for those users who uses a browser compatible?.

It will make it more readable IMO

Issue #2609, opened 2023-08-04
Improve checking of buttons props on MoleculeNotification ✨


jelowin:
### Is your iteration request related to a problem?

On MoleculeNotification component, we have a prop that is an array called `buttons` and we are not checking **buttons.length** to show buttons `sui-MoleculeNotification-buttonsContainer`.

_This is causing that, if we pass an empty array, the container is it painted._

![Screenshot 2023-08-04 at 11 30 55](https://github.com/SUI-Components/sui-components/assets/74491547/4630155f-aaf9-4585-8957-98ae451333e3)

**Incorrect View**
![Screenshot 2023-08-04 at 11 30 19](https://github.com/SUI-Components/sui-components/assets/74491547/2471167f-32dc-4d2f-9437-4d82b4518ea1)

**Correct View**
![Screenshot 2023-08-04 at 11 32 49](https://github.com/SUI-Components/sui-components/assets/74491547/ee506b35-cf3b-4d93-93e0-91684ca2eb54)

### Describe the solution you'd like

![Screenshot 2023-08-04 at 11 31 32](https://github.com/SUI-Components/sui-components/assets/74491547/cd097c57-aced-436a-a2fb-3dc3538c5113)

![Screenshot 2023-08-04 at 11 31 51](https://github.com/SUI-Components/sui-components/assets/74491547/c90eabd0-df89-4028-9d82-f2ca1dcca428)

PR #2655, opened 2023-10-26
❓ fix(components/molecule/notification): Fix MoleculeNotification buttons prop to avoid render container if no length

jelowin:
## molecule/notification

**https://github.com/SUI-Components/sui-components/issues/2609**: 

### Types of changes

- [ ] 🧾 Documentation
- [ ] 📷 Demo
- [ ] 🧪 Test
- [ ] 🧠 Refactor
- [ ] 💄 Styles
- [ ] 🛠️ Tool

### DIFF
--- a/components/molecule/notification/src/index.js
+++ b/components/molecule/notification/src/index.js
@@ -30,7 +30,7 @@ const MoleculeNotification = ({
   autoClose: autoCloseTiming = AUTO_CLOSE.short,
   onClose = EMPTY_METHOD,
   effect = true,
-  buttons,
+  buttons = [],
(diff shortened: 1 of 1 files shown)


components/molecule/notification/src/index.js
```diff
@@ -131,9 +131,9 @@ const MoleculeNotification = ({
             </div>
           )}
         </div>
-        {buttons && (
+        {buttons.length ? (
```
andresin87, line 134:
```suggestion
        {buttons?.length ? (
```

and remove the default value

PR #2694, opened 2024-02-06
feat(components/atom/loading): Add atom loading component

oegea:
### Types of changes


davidmartin84 (approved):

andresin87:
wait

jordevo (approved):

components/atom/loading/src/index.js
```diff
@@ -0,0 +1,11 @@
+import PropTypes from 'prop-types'
+
+const AtomLoading = ({role = 'status', ...props}) => {
+  return <div role="status" {...props} className="sui-AtomLoading" />
```
oriolpuig: We have defined a default value for `role` prop but we are not using it inside, right?

```suggestion
  return <div role={role} {...props} className="sui-AtomLoading" />
```

components/atom/loading/test/index.test.js
```diff
@@ -0,0 +1,55 @@
...
+
+    // When
+    const {container} = setup(props)
+
+    // Then
+    expect(container.innerHTML).to.be.a('string')
+    expect(container.innerHTML).to.not.have.lengthOf(0)
+  })
+
+  it.skip('should NOT extend classNames', () => {
```
oriolpuig: Can we remove or unskip this test?

oriolpuig (approved):
👌🏻 I don't know if make sense to use this `AtomLoading` inside `AtomButton` to reuse the code and components instead of have the logic duplicated 🤔

oegea:
> 👌🏻 I don't know if make sense to use this `AtomLoading` inside `AtomButton` to reuse the code and components instead of have the logic duplicated 🤔

Yes! We haven't included it yet. First we want to create the component and release it, and then create a different pr to requiring it from the atom button. We are doing this in two steps just to avoid CI problems when submitting changes.

andresin87 (commented):
✅

PR #2695, opened 2024-02-06
feat(components/molecule/stepper): improve compressed style UX

jordevo:
## molecule/stepper
🔍 Show

Making step label's text flow through the component:
image

Also fixing for vertical disposition:
image

**TASK**: 

### Types of changes

- [ ] 🧾 Documentation
- [ ] 📷 Demo
- [ ] 🧪 Test
- [ ] 🧠 Refactor
- [x] 💄 Styles
- [ ] 🛠️ Tool

### DIFF
--- a/components/molecule/stepper/src/Step/settings.js
+++ b/components/molecule/stepper/src/Step/settings.js
@@ -26,6 +26,13 @@ export const getIcon = ({
 
 export const getLabel = ({label, steps, step, design, current}) => {
   if (design === DESIGN.DEFAULT) return label
-  else if (design === DESIGN.BASIC) return label
-  else if (design === DESIGN.COMPRESSED && current) return `${step}/${steps}${label ? `: ${label}` : ''}`
+  if (design === DESIGN.BASIC) return label
+  if (design === DESIGN.COMPRESSED && current)
+    return (
+      <>
+        {`${step}/${steps}`}
+        {label && `: `}
+        {label}
+      </>
+    )
 }
--- a/components/molecule/stepper/src/Step/settings.js
+++ b/components/molecule/stepper/src/Step/settings.js
@@ -27,12 +27,5 @@ export const getIcon = ({
 export const getLabel = ({label, steps, step, design, current}) => {
   if (design === DESIGN.DEFAULT) return label
   if (design === DESIGN.BASIC) return label
-  if (design === DESIGN.COMPRESSED && current)
-    return (
-      <>
-        {`${step}/${steps}`}
-        {label && `: `}
-        {label}
-      </>
-    )
+  if (design === DESIGN.COMPRESSED && current) return `${step}/${steps}${label ? `: ${label}` : ''}`
 }
--- a/components/molecule/stepper/src/Step/DefaultStep.js
+++ b/components/molecule/stepper/src/Step/DefaultStep.js
@@ -40,6 +40,13 @@ const DefaultStep = ({
                 marginLeft: `calc(-${(step - 1) * 100}% - ${(step - 1) * 8}px)`
               }
             })}
+          {...(design === DESIGN.COMPRESSED &&
+            current &&
+            alignment === ALIGNMENT.VERTICAL && {
+              style: {
+                marginTop: `calc(-${(step - 1) * 100}% - ${(step - 1) * 8}px)`
+              }
+            })}
         >
           {getLabel({steps, step, design, label, current})}
         </div>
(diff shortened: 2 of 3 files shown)


emiliovillu (approved):

andresin87 (approved):
Does it work in vertical mode?

components/molecule/stepper/src/Step/settings.js
```diff
@@ -26,6 +26,13 @@ export const getIcon = ({
...
-  else if (design === DESIGN.BASIC) return label
-  else if (design === DESIGN.COMPRESSED && current) return `${step}/${steps}${label ? `: ${label}` : ''}`
+  if (design === DESIGN.BASIC) return label
+  if (design === DESIGN.COMPRESSED && current)
+    return (
+      <>
+        {`${step}/${steps}`}
+        {label && `: `}
+        {label}
+      </>
```
giodelabarrera: A question, is it necessary to return the jsx? I mean, before `getLabel` returned `string`, and with this iteration, it  would additionally return a jsx or I'm wrong.

components/molecule/stepper/src/Step/settings.js
```diff
@@ -26,6 +26,13 @@ export const getIcon = ({
...
 export const getLabel = ({label, steps, step, design, current}) => {
   if (design === DESIGN.DEFAULT) return label
-  else if (design === DESIGN.BASIC) return label
-  else if (design === DESIGN.COMPRESSED && current) return `${step}/${steps}${label ? `: ${label}` : ''}`
+  if (design === DESIGN.BASIC) return label
+  if (design === DESIGN.COMPRESSED && current)
+    return (
+      <>
+        {`${step}/${steps}`}
+        {label && `: `}
```
giodelabarrera: If `label` is `undefined` we will have an `"undefined"`, could it be?

components/molecule/stepper/src/Step/settings.js
```diff
@@ -26,6 +26,13 @@ export const getIcon = ({
...
+        {label}
```
andresin87: 🚨 🚨 🚨
Cannot do that because in case of using google translate addon it wraps every "" dom text with a `<font/>`deprecated tag. The problem is the react-virtual-dom  tries to reconciliate, the read don had change and the react-fiber crashes.

andresin87: Thats why the complete text is composed as a single template string.

jordevo: sadly we can't return what we're currently returning... ie `3/5: [Object object]`

components/molecule/stepper/src/Step/settings.js
jordevo: should be alright... `getLabel` is also within JSX...

```
        <div {...stuff}>
          {getLabel({steps, step, design, label, current})}
        </div>
```

components/molecule/stepper/src/Step/settings.js
andresin87: why it has Object as a label?

components/molecule/stepper/src/Step/settings.js
jordevo: nope, nothing will be rendered in that case... I had to double check TBH 😅

jordevo:
@andresin87 @giodelabarrera thanks for reviewing the PR... turns out the issue with the label when the component had the `design='COMPRESSED'` set came from the way the demo is generating the labels...

https://github.com/SUI-Components/sui-components/blob/master/components/molecule/stepper/demo/articles/ArticleDesignAlignment.js#L26-L28 

so nothing to worry about that... will remove those changes

components/molecule/stepper/src/Step/settings.js
jordevo: https://github.com/SUI-Components/sui-components/pull/2695#issuecomment-1930031816

components/molecule/stepper/src/Step/settings.js
jordevo: https://github.com/SUI-Components/sui-components/pull/2695#issuecomment-1930031816

components/molecule/stepper/src/Step/settings.js
jordevo: https://github.com/SUI-Components/sui-components/pull/2695#issuecomment-1930031816

jordevo:
> Does it work in vertical mode?

it works now! :) 

image

PR #2722, opened 2024-04-15
Feature/avoid page scrolling when drawer open prop

xmurcia:
## Category/Component

`🔍 Show`

Allow a new prop to avoid the scroll on a page when the drawer is opened. The same feature is already implemented in the molecule/modal component.

### Types of changes

- [ ] 🧾 Documentation
- [X] 📷 Demo
- [X] 🧪 Test
- [ ] 🧠 Refactor
- [ ] 💄 Styles
- [ ] 🛠️ Tool

### DIFF
--- a/components/molecule/drawer/demo/DemoPageScrollable.js
+++ b/components/molecule/drawer/demo/DemoPageScrollable.js
@@ -1,0 +1,52 @@
+import {useRef, useState} from 'react'
+
+import PropTypes from 'prop-types'
+
+import {Article, Box, H2, H4, Label, Paragraph, RadioButton} from '@s-ui/documentation-library'
+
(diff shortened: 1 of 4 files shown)


davidmartin84 (approved):

jordevo (approved):

andresin87 (approved):

andresin87:
tab navigation blocking is also implemented?

xmurcia:
> tab navigation blocking is also implemented?

This kind of behavior is not implemented, I guess you mean when the drawer is opened, to limit the scope of the focus just inside the drawer and not other DOM element, right? Maybe we can have a look in a different task in order to do it in modals and drawer because I've seen it is also missing in there.

PR #2738, opened 2024-05-13
[FEAT] migrate atom/label to TS

kikoruiz:
## Description
__Migrate the first component (`AtomLabel`) to TypeScript.__ Behaviour of the component remains the same.

## Type of changes
- [x] 🧠 Refactor
- [x] 🛠️ Tool

### Complete list of changes:

* `AtomLabel` migration to TypeScript.
* Add new CI job for typechecking.
* Define missing types and module declarations for both 3rd-party and our own packages.
* Adapt some TS configurations.


components/atom/label/src/index.tsx
```diff
@@ -0,0 +1,57 @@
+/* eslint-disable react/prop-types */
```
kikoruiz: We need to update the lint config for TS files in order to remove this rule, not needed anymore.

components/atom/label/mod.d.ts
kikoruiz, line 1: Pending to be moved to root directory.

components/atom/label/index.d.ts
kikoruiz, line 1: Pending to be moved to root directory.

components/atom/label/src/index.tsx
```diff
@@ -0,0 +1,57 @@
...
+  /**
+   * used as for attribute. Must be the same as the form element id
+   */
+  name: string
+  /**
+   * The label itself
+   */
+  text: string
+  /**
+   * Allows label to be displayed inline to de left
```
ferransimon: Typo `Allows label to be displayed inline to de left` -> `Allows label to be displayed inline to the left`

components/atom/label/src/index.tsx
```diff
@@ -0,0 +1,57 @@
...
+   */
+  inline?: string
```
ferransimon: I see in the code that this prop can be either left or right so maybe we can add some Type checking with these 2 values instead of using a string.

carlosvillu (approved):

components/atom/label/src/index.tsx
kikoruiz: Copypasted, sorry 😂

components/atom/label/src/index.tsx
kikoruiz: Yep, it makes sense, it's not used for any other reason ;)

quinwacca (approved):

components/atom/label/mod.d.ts
mariogl: Resolved in #2935

components/atom/label/index.d.ts
mariogl: Resolved in #2935

PR #2748, opened 2024-07-15
feat(components/atom/progressBar): Add hideBorder property

hpintos:
#### `🔍 Show`

**TASK**:  https://jira.ets.mpi-internal.com/browse/MTR-84991

Add property `hideBorder` to ProgressBar 

### Types of changes

### DIFF
--- a/components/atom/progressBar/src/ProgressBarLine/index.js
+++ b/components/atom/progressBar/src/ProgressBarLine/index.js
@@ -18,7 +18,7 @@ const ProgressBarLine = ({
   indicatorBottom,
   indicatorTotal,
   isAnimatedOnChange,
-  isBorderless,
+  isBorderless = false,
   percentage,
   mainBarPercentage,
   extraBarPercentage,
--- a/components/atom/progressBar/src/ProgressBarLine/index.js
+++ b/components/atom/progressBar/src/ProgressBarLine/index.js
@@ -14,11 +14,11 @@ import useIndicator from './useIndicator.js'
 import usePercentage from './usePercentage.js'
 
 const ProgressBarLine = ({
-  hideBorder,
   hideIndicator,
   indicatorBottom,
   indicatorTotal,
   isAnimatedOnChange,
+  isBorderless,
   percentage,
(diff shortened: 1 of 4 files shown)


components/atom/progressBar/src/ProgressBarLine/index.js
```diff
@@ -14,6 +14,7 @@ import useIndicator from './useIndicator.js'
 import usePercentage from './usePercentage.js'
 
 const ProgressBarLine = ({
+  hideBorder,
```
jordevo: some _positive_ boolean prop (say `borderless`) would probably make for better naming IMHO... what do you think?

hpintos: I was trying to follow the same pattern, as the property `hideIndicator` already exists. Thoughts?

jordevo: yeah saw that... the difference I see is perhaps `hideIndicator` makes more sense as an _action_ word rather than a  _boolean-ish_ named prop as the _Indicator_ seems to be a more substantial piece of the component, whereas displaying it _borderless_ or not is something that's just related to styles... if that makes sense? both are fine anyway...

jordevo (approved):

components/atom/progressBar/src/ProgressBarLine/index.js
hpintos: Make sense. Will do the change. Thanks for the feedback.

components/atom/progressBar/src/ProgressBarLine/index.js
```diff
@@ -18,6 +18,7 @@ const ProgressBarLine = ({
   indicatorBottom,
   indicatorTotal,
   isAnimatedOnChange,
+  isBorderless,
```
jordevo:
```suggestion
  isBorderless = false,
```

even if `undefined` is `falsy`, should we set a default value of `false`, perhaps?

jordevo (approved):

andresin87:
any demo?

PR #2751, opened 2024-07-29
feat(components/molecule/selectPopover): add id prop

giodelabarrera:
## Category/Component

🔍 Show

**TASK**: https://jira.ets.mpi-internal.com/browse/SCMI-124924

Add id attribute for div with states of select popover to better observability for INP (Core Web Vitals)

### Types of changes

- [ ] 🧾 Documentation
- [ ] 📷 Demo
- [ ] 🧪 Test
- [ ] 🧠 Refactor
- [ ] 💄 Styles
- [ ] 🛠️ Tool

### DIFF
--- a/components/molecule/selectPopover/src/index.js
+++ b/components/molecule/selectPopover/src/index.js
@@ -30,6 +30,7 @@ const MoleculeSelectPopover = ({
   fullWidth,
   hideActions,
   iconArrowDown: IconArrowDown,
+  id = '',
   isDisabled = false,
   isSelected = false,
   onAccept = () => {},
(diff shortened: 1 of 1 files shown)


components/molecule/selectPopover/src/index.js
```diff
@@ -30,6 +30,7 @@ const MoleculeSelectPopover = ({
   fullWidth,
   hideActions,
   iconArrowDown: IconArrowDown,
+  id = '',
```
andresin87: is it really required an empty string?
```suggestion
  id,
```

giodelabarrera: Not really, but I think it adds consistency throughout the component when any developer wants to work/iterate with `id`

giodelabarrera: do you consider it unnecessary now? or with the conditional within object is sufficient? or with the propType `string` is sufficient?

andresin87 (approved):

PR #2768, opened 2024-10-15
feat(components/molecule/photoUploader): add prop onSortPhotoStart to…

davidmartin84:
… photouploader

## Category/Component

**TASK**: 

### Types of changes

- [ ] 🧾 Documentation
- [ ] 📷 Demo
- [ ] 🧪 Test
- [ ] 🧠 Refactor
- [ ] 💄 Styles
- [ ] 🛠️ Tool

### DIFF
--- a/components/molecule/photoUploader/src/index.js
+++ b/components/molecule/photoUploader/src/index.js
@@ -76,6 +76,7 @@ const MoleculePhotoUploader = forwardRef(
       onDropFiles = noop,
       onFileDialogOpen = noop,
       onEmptyViewClick = noop,
+      onSortPhotoStart = noop,
       outputImageAspectRatio = DEFAULT_IMAGE_ASPECT_RATIO,
       outputImageAspectRatioDisabled = false,
       rejectPhotosIcon,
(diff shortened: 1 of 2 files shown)


components/molecule/photoUploader/src/index.js
```diff
@@ -516,6 +522,9 @@ MoleculePhotoUploader.propTypes = {
   /** Func to be executed when dropzone area is clicked */
   onEmptyViewClick: PropTypes.func,
 
+  /** Func to be executed when photo is being sorted */
+  onSortPhotoStart: PropTypes.func,
```
jordevo, line 526: 🍆 not sure then what does `Start` mean there...? is it because the event fires when the user "starts" sorting photos? should we aim for some more simple naming ie `onSortPhotos` or something like that?

jordevo: oh I see it's coupled to the name the ReactSortable component prop has... `onStart`

jordevo: still, I would keep it simple... `onSortPhotos` is my suggestion 😉

davidmartin84: Thanks!!!

PR #2854, opened 2025-04-30
feat(components/molecule/textareaField): add aria props to character …

quinwacca:

This PR addresses an accessibility issue reported during an audit.

The characters counter is a live region that is being updated constantly and should be reported to screen readers.
The current `<span>` will be wrapped by an element with two aria props:
- `aria-live=polite` to let screen readers announce changes, and set it `polite` to make screen readers wait until the user finishes listening to a message before another one starts.
- `aria-atomic=true` to mark every element inside as the same unit that can report changes.

### Types of changes

- [x] 🦾 a11y

### Screenshots - Animations

image

### DIFF
--- a/components/molecule/textareaField/src/index.js
+++ b/components/molecule/textareaField/src/index.js
@@ -64,7 +64,11 @@ const MoleculeTextareaField = ({
     }
   }
 
-  const helpTextComputed = computeHelpText()
+  const helpTextComputed = (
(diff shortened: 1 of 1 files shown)


oriolpuig (approved):
❤️

ivanmlaborda (approved):

andresin87 (commented):
🟢

PR #2855, opened 2025-05-02
feat(utils/sui-theme): use native functions to lighten and darken colors

quinwacca:

Use native CSS functions to lighten and darken the SUI custom button. SASS functions are not working with CSS variables, so we cannot build an app that uses them.

**Sass does not evaluate CSS custom property values, because these are run-time variables.**

### Types of changes


utils/sui-theme/src/components/_button.scss
```diff
@@ -189,11 +189,11 @@
   color: $color;
 
   &:hover {
-    background-color: lighten($background-color, 15%);
+    background-color: color-mix(in srgb, $background-color, white 15%);
```
andresin87:
```suggestion
    background-color: color-mix(in srgb, $background-color, $c-white 15%);
```

utils/sui-theme/src/components/_button.scss
```diff
@@ -189,11 +189,11 @@
...
   }
 
   &:active {
-    background-color: darken($background-color, 10%);
+    background-color: color-mix(in srgb, $background-color, black 10%);
```
andresin87:
```suggestion
    background-color: color-mix(in srgb, $background-color, $c-black 10%);
```

andresin87 (changes requested):

hpintos (approved):

PR #2858, opened 2025-05-05
feat(components/molecule/avatar): use paragraph and aria hide the nam…

quinwacca:

This PR addresses an accessibility issue reported during an audit.

We should use the paragraph semantic tag ( `<p>` ) in the name fallback (the initials of the username).
Also, we must mark it with `aria-hidden="true"` so it cannot be read by screen readers.

### Types of changes

- [x] 🦾 a11y

### DIFF
--- a/components/molecule/avatar/src/AvatarFallbackName/index.js
+++ b/components/molecule/avatar/src/AvatarFallbackName/index.js
@@ -21,9 +21,9 @@ const MoleculeAvatarFallbackName = ({
   })
 
   return (
-    <div className={className} aria-label={nameProp} style={{backgroundColor}} {...others}>
+    <p className={className} aria-hidden="true" aria-label={nameProp} style={{backgroundColor}} {...others}>
       {name}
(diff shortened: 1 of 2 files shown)


components/molecule/avatar/src/AvatarFallbackName/index.js
```diff
@@ -21,9 +21,9 @@ const MoleculeAvatarFallbackName = ({
   })
 
   return (
-    <div className={className} aria-label={nameProp} style={{backgroundColor}} {...others}>
+    <p className={className} aria-hidden="true" aria-label={nameProp} style={{backgroundColor}} {...others}>
```
AgonisticKatai: 🤪
```suggestion
    <p className={className} aria-hidden aria-label={nameProp} style={{backgroundColor}} {...others}>
```

AgonisticKatai (approved):

jordevo (approved):

PR #2865, opened 2025-05-13
Feature/a11y  aria described by help text

quinwacca:

To solve an a11y issue, we need to add a relationship between the field's help text and the input component, we need to add an id to the helpText and an aria-describedby prop to the input, with the same id.
The Molecule Field component can manage that, setting an ID from the `name` prop.

### Types of changes

- [x] 🦾 a11y

### DIFF
--- a/components/atom/helpText/src/index.js
+++ b/components/atom/helpText/src/index.js
@@ -6,11 +6,11 @@ import Injector from '@s-ui/react-primitive-injector'
 
 import {BASE_CLASS} from './settings.js'
 
-const AtomHelpText = forwardRef(({id, text}, forwardedRef) => {
+const AtomHelpText = forwardRef(({text, ...props}, forwardedRef) => {
   const isTextString = typeof text === 'string'
   const Component = isTextString ? 'span' : Injector
   return (
-    <Component className={BASE_CLASS} id={id} {...(isTextString && {ref: forwardedRef})}>
+    <Component className={BASE_CLASS} {...props} {...(isTextString && {ref: forwardedRef})}>
       {text}
     </Component>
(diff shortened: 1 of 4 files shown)


components/molecule/select/src/components/SingleSelection.js
```diff
@@ -87,7 +88,7 @@ const MoleculeSelectSingleSelection = forwardRef(
           size={selectSize}
           tabIndex={tabIndex}
         >
-          <AtomInput ref={forwardedRef} inputMode={inputTypes.NONE} noBorder />
+          <AtomInput aria-describedby={ariaDescribedBy} ref={forwardedRef} inputMode={inputTypes.NONE} noBorder />
```
hpintos: Correct me if Im wrong, but what if you use the spread props approach here? I think the`AtomInput` accepts whatever prop you throw at it and pass it down to it's `Component`. Then you wouldn't have to specifically pass down the `aria-describedby` prop
```suggestion
          <AtomInput ref={forwardedRef} inputMode={inputTypes.NONE} noBorder {...props} />
```

components/atom/helpText/src/index.js
```diff
@@ -6,11 +6,11 @@ import Injector from '@s-ui/react-primitive-injector'
 
 import {BASE_CLASS} from './settings.js'
 
-const AtomHelpText = forwardRef(({text}, forwardedRef) => {
+const AtomHelpText = forwardRef(({id, text}, forwardedRef) => {
```
hpintos: Same here. I see this pattern being used a lot.

andresin87 (approved):

PR #2871, opened 2025-05-19
Feature/a11y  aria described by validation text

quinwacca:

To solve an a11y issue, we need to add a relationship between the field's help text and the input component.

In this case, we are applying the [same changes we did](https://github.com/SUI-Components/sui-components/pull/2865) to the `helpText` component, but in the `validationText` component.

We need to add an id to the `validationText` but also a `role="alert"` if the status is `ERROR`.

The Molecule Field component can manage that, setting an ID from the `name` prop, and setting the role if it is needed.

### Types of changes

- [x] 🦾 a11y

### DIFF
--- a/components/atom/validationText/src/index.js
+++ b/components/atom/validationText/src/index.js
@@ -6,11 +6,11 @@ import Injector from '@s-ui/react-primitive-injector'
 
 import {getClassNames, TYPES} from './settings.js'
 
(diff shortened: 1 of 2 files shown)


AgonisticKatai (approved):

AgonisticKatai:
Could be interesting to add some tests to check this new behavior.

andresin87 (approved):

PR #2872, opened 2025-05-20
feat(components/molecule/tabs): making molecule tab more accessible - Gemini 2.5 Pro

jordevo:
## molecule/tabs
Trying out GH's Copilot with Gemini 2.5 Pro for a particular a11y ticket.

Given the prompt:
> _How would you make this React component more accessible?_

Given the file:
> [components/molecule/tabs/src/components/MoleculeTab.js](https://github.com/SUI-Components/sui-components/pull/2870/files#diff-9aa39d14feda7f4229414b2169e19f0d83c8d02216295d2a54a41f1bce7d97bb)

**TASK**: 

### Types of changes

- [ ] 🧾 Documentation
- [ ] 📷 Demo
- [ ] 🧪 Test
- [ ] 🧠 Refactor
- [ ] 💄 Styles
- [ ] 🛠️ Tool

### Screenshots - Animations
N/A

### AI response

Okay, I can help you make this React component more accessible. Here are some suggestions:

1.  **Keyboard Interaction**: Tabs should be operable via the keyboard. Users should be able to activate a tab using the `Enter` or `Space` key when it has focus.
2.  **Focus Management**:
    *   Each tab (`<li>`) should be focusable. You can achieve this by adding `tabIndex={0}`. For disabled tabs, `tabIndex={-1}` is more appropriate.
    *   Consider how focus is managed when a tab is activated or when arrow keys are used to navigate between tabs (though arrow key navigation is typically handled by the parent `tablist` container).
3.  **ARIA Attributes**:
    *   `aria-disabled`: Add `aria-disabled={disabled}` to the `<li>` element to communicate its state to assistive technologies.
    *   `aria-label`: If the `label` prop can be a React node (not just a string), provide a string alternative via `aria-label` for screen readers.
    *   `aria-hidden`: If the `icon` is purely decorative and doesn't convey information not already present in the `label`, add `aria-hidden="true"` to its `<span>`.

Here's how you could modify the code:
````javascript
// ...existing code...
const MoleculeTab = forwardRef(
  (
    {
      active,
      autoScrollIntoView = true,
      count,
      disabled,
      icon,
      id = 'molecule-tab-content',
      isIntersecting,
      label,
      numTab,
      onChange
    },
    forwardedRef
  ) => {
    const innerRef = useRef()

    const handleChange = ev => {
      !disabled && onChange(ev, {numTab})
    }

    const handleKeyDown = ev => {
      if (!disabled && (ev.key === 'Enter' || ev.key === ' ')) {
        ev.preventDefault() // Prevent default space scroll
        onChange(ev, {numTab})
      }
    }

    useEffect(() => {
// ...existing code...
    const className = cx(CLASS_TAB, {
      [CLASS_TAB_ACTIVE]: active,
      [CLASS_TAB_DISABLED]: disabled
    })

    const tabLabel = typeof label === 'string' ? label : `Tab ${numTab + 1}`;

    return (
      <li
        className={className}
        onClick={handleChange}
        onKeyDown={handleKeyDown}
        ref={useMergeRefs(innerRef, forwardedRef)}
        role="tab"
        tabIndex={disabled ? -1 : 0}
        aria-selected={active}
        aria-controls={`${id}-${numTab}`}
        aria-disabled={disabled}
        aria-label={typeof label !== 'string' ? tabLabel : undefined}
      >
        {icon && <span className={CLASS_TAB_ICON} aria-hidden="true">{icon}</span>}
        {!isNaN(count) && <span className={CLASS_TAB_COUNT}>{count}</span>}
        <span>{label}</span>
      </li>
    )
  }
)

MoleculeTab.propTypes = {
// ...existing code...
````
Key changes:
*   Added `handleKeyDown` to allow activation with `Enter` or `Space`.
*   Added `onKeyDown={handleKeyDown}` to the `<li>`.
*   Added `tabIndex={disabled ? -1 : 0}`.
*   Added `aria-disabled={disabled}`.
*   Added `aria-hidden="true"` to the icon's span, assuming it's decorative.
*   Added a conditional `aria-label` in case the `label` prop is a React Node. If `label` is always a string, this `aria-label` might be redundant or could simply be `label` itself if you want to ensure it's announced.

Remember that the parent component managing these tabs (the `tablist`) also has accessibility responsibilities, such as managing roving `tabIndex` or arrow key navigation between tabs.

----

## Subsequent prompting and context

Given the prompt:
> how would you make it so that a tab that has the current focus while navigating via keyboard outstands visually?

Given the component folder as context.

### AI response

Okay, I can help you with that. Here's a plan to make the focused tab stand out visually using CSS:

1.  **Define a new SCSS variable** in `_settings.scss` for the focus style (e.g., an outline).
2.  **Update `index.scss`** to apply this focus style to the tab item using the `:focus-visible` pseudo-class. This ensures the style only applies during keyboard navigation and not on mouse clicks.

Here are the suggested changes:

### [_settings.scss](file:///Users/jordi.munoz/sui-components/components/molecule/tabs/src/_settings.scss)

Add a new variable for the focus outline.

```scss
// ...existing code...
$w-sui-molecule-tabs-vertical-item: 100% !default;
$ws-sui-molecule-tabs-vertical-item: nowrap !default;
$outline-sui-molecule-tabs-focus-visible: $c-primary auto 1px !default;
```

### [index.scss](file:///Users/jordi.munoz/sui-components/components/molecule/tabs/src/styles/index.scss)

Add a new style rule for the tab item when it receives focus via keyboard.

```scss


Made changes.

// ...existing code...
  &-item {
    @include reset-link;
    align-items: center;
// ...existing code...
    white-space: nowrap;
    width: 100%;
    scroll-snap-align: $ssa-sui-molecule-tabs;

    &:focus-visible {
      outline: $outline-sui-molecule-tabs-focus-visible;
      // Or, for a more subtle approach if you prefer box-shadow:
      // box-shadow: 0 0 0 2px $c-primary;
      // border-radius: $bdrs-sui-molecule-tabs; // Optional: if you want the focus outline to follow border-radius
    }

    &:first-of-type {
      margin: $m-sui-molecule-tabs-first-item;
// ...existing code...
```

Made changes.

## Notes on AI behaviour

- after the second prompt and given the context of the whole component folder, it didn't follow the Emmet abbreviation convention for Sass variables


components/molecule/tabs/src/styles/index.scss
```diff
@@ -245,6 +245,10 @@ $item-base-class: '#{$base-class}-item';
     width: 100%;
     scroll-snap-align: $ssa-sui-molecule-tabs;
 
+    &:focus-visible {
+      outline: $ol-sui-molecule-tabs-focus-visible;
```
andresin87, line 249: https://github.com/SUI-Components/sui-components/pull/2870/files#diff-550276866a860d2a254b47d78839bbe31002a6a9d92afa70f721a7cbc2acdfc9L247

andresin87: prefiero este aproach

PR #2883, opened 2025-05-26
Feat add prop text aria label breadcrumb

emiliovillu:
#### `🔍 Show`

**TASK**: [MTR-99131](https://jira.ets.mpi-internal.com/browse/MTR-99131)

Add `labelAttr`prop to can add aria-label with translate

### Types of changes

- [ ] 🧾 Documentation
- [X] 📷 Demo
- [ ] 🧪 Test
- [ ] 🧠 Refactor
- [ ] 💄 Styles
- [ ] 🛠️ Tool

### DIFF
--- a/components/molecule/breadcrumb/src/index.js
+++ b/components/molecule/breadcrumb/src/index.js
@@ -6,7 +6,7 @@ import ChevronRight from '@s-ui/react-icons/lib/Chevronright'
 import {BASE_CLASS, breadcrumbClassName, isFunction} from './settings.js'
 
 const BreadcrumbBasic = ({
-  ariaLabelText,
+  labelAttr,
   items,
   icon,
   linkFactory: Link = ({to, href, className, children}) => (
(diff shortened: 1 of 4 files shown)


jordevo (approved):
not sure why we couldn't be using `aria-label` attribute itself, rather than having a new prop...?

components/molecule/breadcrumb/src/index.js
```diff
@@ -6,6 +6,7 @@ import ChevronRight from '@s-ui/react-icons/lib/Chevronright'
 import {BASE_CLASS, breadcrumbClassName, isFunction} from './settings.js'
 
 const BreadcrumbBasic = ({
+  ariaLabelText,
```
quinwacca: We can use the existing label as default:
```suggestion
  ariaLabelText = "breadcrumb",
```

quinwacca (approved):

components/molecule/breadcrumb/src/index.js
emiliovillu: Yesterday, talking with @andresin87, he told me not to put values for defects in the component, only in the Demo.

PR #2904, opened 2025-06-11
fix(components/molecule/photoUploader): Pass external props on rotate action

oriolpuig:
#### `❓ Ask`

**TASK**:  😂  NO IDEA 😂 

As a consumer  of this component, you can pass the `url` and other props to be stored inside the component, and this will let you know more info outside this component when the user performs some action (rotate, delete, order, ...). For example, you can pass an `id`, and this will make your life easier because then a user do an action into the component, the component will pass you all file information. 

With this PR, we aim to improve the `rotate` action, which currently only passes the `url` and not the extra props.

This approach is well-managed in other components' flow, but not in the rotate action. 

### Types of changes

### DIFF
--- a/components/molecule/photoUploader/src/PhotosPreview/index.js
+++ b/components/molecule/photoUploader/src/PhotosPreview/index.js
@@ -99,8 +99,8 @@ const PhotosPreview = ({
         list[index].blob = blob
         list[index].isModified = true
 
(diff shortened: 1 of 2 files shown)


andresin87 (approved):

candymd (approved):
🚀

PR #2955, opened 2025-11-17
Feat/molecule autosuggest dropdownlist a11y

Sabri1209:
Apply some a11y requirements

### Types of changes

- [ ] 🧾 Documentation
- [ ] 📷 Demo
- [ ] 🧪 Test
- [ ] 🧠 Refactor
- [ ] 💄 Styles
- [ ] 🛠️ Tool

### DIFF
--- a/components/molecule/autosuggest/src/index.js
+++ b/components/molecule/autosuggest/src/index.js
@@ -89,6 +89,13 @@ const MoleculeAutosuggest = ({
     }
   )
 
+  const accessibilityProps = {
+    tabIndex: '0',
+    role: 'combobox',
+    'aria-expanded': isOpenState,
+    ...(multiselection && {'aria-controls': id})
+  }
+
   const closeList = ev => {
     const {current: domMoleculeAutosuggest} = innerRefMoleculeAutosuggest
     handleToggle(ev, {isOpen: false})
@@ -209,6 +216,7 @@ const MoleculeAutosuggest = ({
     onSelect,
     onToggle: handleToggle,
     state,
+    ...(!multiselection && {...accessibilityProps}),
     ...restProps
   }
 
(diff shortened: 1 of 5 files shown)


components/molecule/autosuggest/src/index.js
```diff
@@ -89,6 +89,13 @@ const MoleculeAutosuggest = ({
     }
   )
 
+  const accessibilityProps = {
```
andresin87, line 92: set it into the object  directly (if it is used only once) or define it outside of the component fn cycle

Sabri1209: The problem is that the object that’s already created is related to the autosuggest (**autosuggestSelectionProps**), but not to the wrapper `div`. Defining it outside the component isn’t possible either, since I’m using a prop value.

components/molecule/autosuggest/test/index.test.js
```diff
@@ -70,19 +70,6 @@ describe(json.name, () => {
       expect(container.innerHTML).to.not.have.lengthOf(0)
     })
 
-    it('should NOT extend classNames', () => {
```
andresin87, line 73: Do not remove it. just change it to 'should extend classNames'

components/molecule/autosuggest/test/index.test.js
```diff
@@ -70,19 +70,6 @@ describe(json.name, () => {
...
-      // Given
-      const props = {className: 'extended-classNames'}
-      const findSentence = str => string => string.match(new RegExp(`S*${str}S*`))
-
-      // When
-      const {container} = setup(props)
-      const findClassName = findSentence(props.className)
-
-      // Then
-      expect(findClassName(container.innerHTML)).to.be.null
```
andresin87, line 83:
```js
	  const props = {className: 'extended-classNames'}
      const findSentence = str => string => string.match(new RegExp(`S*${str}S*`))

      // When
      const {container} = setup(props)
      const findClassName = findSentence(props.className)

      // Then
      expect(findClassName(container.innerHTML)).to.not.be.null
```

andresin87: `(******).to.not.be.null`

andresin87 (approved):

